PR #289, opened 2024-02-14
chore(relayer/app): buffer and worker tests

Lazar955:
- Adds tests for worker and buffer
- Removes old `startRelayer` code
- Simnet and smoke tests are using workers now

task: none


halo/cmd/init.go
```diff
@@ -194,6 +194,16 @@ func InitFiles(ctx context.Context, initCfg InitConfig) error {
 					Name:   "omni",
 					IsOmni: true,
 				},
+				{
```
corverroos, line 197: add comments where this comes from please, also TODO to make it dynamic.

relayer/app/buffer.go
```diff
@@ -32,11 +32,11 @@ func newActiveBuffer(chainName string, mempoolLimit int64, size int, sender Send
 	}
 }
 
-func (b *activeBuffer) AddInput(_ context.Context, submission xchain.Submission) error {
+func (b *activeBuffer) AddInput(ctx context.Context, submission xchain.Submission) error {
 	select {
+	case <-ctx.Done():
+		b.submitErr(errors.Wrap(ctx.Err(), "context canceled"))
```
corverroos, line 38: pls add a comment why we are not returning the error

relayer/app/buffer_internal_test.go
```diff
@@ -0,0 +1,70 @@
...
+			default:
+			}
+			err := buffer.Run(ctx)
+			require.NoError(t, err)
+		}
+	}()
+
+	// Add an item to the buffer
+	err := buffer.AddInput(ctx, xchain.Submission{})
+	require.NoError(t, err)
```
corverroos, line 68: this is going to flap since it is non-deterministic.

Please refactor to deterministic tests.

```
buf := new()

for sub in input {
 go addAsync(sub)
 }
 
 var output []Sub
 while len(output) != len(intput) {
   append(putput, <-sender.Next())
 }
 
 assertEqual(input, output)

```

relayer/app/buffer.go
```diff
@@ -32,11 +32,11 @@ func newActiveBuffer(chainName string, mempoolLimit int64, size int, sender Send
...
 	case b.buffer <- submission:
```
Lazar955, line 39: we are blocking now, instead of restarting worker when buffer is full

corverroos (approved):

test/smoke/smoke_test.go
```diff
@@ -92,6 +92,8 @@ func testSmoke(t *testing.T, ethCl engine.API) {
 	const (
 		srcChainBlockPeriod = 1 * time.Millisecond * 100
 		srcChainID          = 1
+		chainA              = 100
+		chainB              = 200
```
corverroos, line 96: same here add a comment where this comes from

PR #288, opened 2024-02-14
feat(contracts): initial omni avs

kevinhalliday:
Initial Omni AVS contract.

- AVS contract reads operator stake from eigen layer core, and sends to omni
- Updates sent to omni discriminate between operator stake & delegations
- forge integration tests added

task: https://app.asana.com/0/1206208509925075/1206566996089593

PR #292, opened 2024-02-15
feat(contracts): use transparent proxies

ttarsi:
- use upgradeable contracts for `OmniPortal` and `FeeOracleV1`
- update tests (necessary, as using proxies changes initialization / constructor setup)
- update e2e with proxies

task: none


contracts/package.json
```diff
@@ -8,6 +8,7 @@
   },
   "devDependencies": {
     "@openzeppelin/contracts": "^5.0.1",
+    "@openzeppelin/contracts-upgradeable": "^5.0.1",
```
kevinhalliday, line 11: We'll have conflicts with openzeppelin version beween this and https://github.com/omni-network/omni/pull/288

test/e2e/netman/helpers.go
```diff
@@ -30,32 +30,112 @@ func deployContract(ctx context.Context, chainID uint64, client *ethclient.Clien
...
 
-	portalAddr, tx, _, err := bindings.DeployOmniPortal(txOpts, client, owner, feeOracleAddr, valSetID, validators)
+	encodedFeeOracleV1Initializer, err := feeOracleV1ABI.Pack("initialize", owner, fee)
 	if err != nil {
-		return common.Address{}, nil, nil, errors.Wrap(err, "deploy portal contract")
+		return common.Address{}, errors.Wrap(err, "pack fee oracle v1 initializer")
+	}
+
+	feeOracleAddr, tx, _, err := bindings.DeployTransparentUpgradeableProxy(
+		txOpts, client, receipt.ContractAddress, common.HexToAddress(pubKeyHex0), encodedFeeOracleV1Initializer,
```
kevinhalliday, line 79: So proxy admin to `pubKeyHex0` for now - right? If we use these scripts for testnet and mainnet in future, which I think is the pla, we should have a proxy admin deployment

this will add additional complexity to netman deployments & outputs

we can add a ticket for this

kevinhalliday (approved):

test/e2e/netman/helpers.go
```diff
@@ -30,32 +30,112 @@ func deployContract(ctx context.Context, chainID uint64, client *ethclient.Clien
...
+	)
+	if err != nil {
+		return common.Address{}, errors.Wrap(err, "deploy fee oracle proxy")
 	}
 
 	receipt, err = bind.WaitMined(ctx, client, tx)
```
corverroos, line 85: is all the waitminded really required? This results in super slow deploys for slow chains like goerli.

corverroos (approved):

contracts/.gas-snapshot
```diff
@@ -1,30 +1,30 @@
...
+OmniPortal_xsubmit_Test:test_xsubmit_duplicateValidator_reverts() (gas: 13337260)
+OmniPortal_xsubmit_Test:test_xsubmit_invalidAttestationRoot_reverts() (gas: 13383151)
+OmniPortal_xsubmit_Test:test_xsubmit_invalidMsgs_reverts() (gas: 13370182)
+OmniPortal_xsubmit_Test:test_xsubmit_noQuorum_reverts() (gas: 13344093)
+OmniPortal_xsubmit_Test:test_xsubmit_wrongChainId_reverts() (gas: 13402409)
+OmniPortal_xsubmit_Test:test_xsubmit_wrongStreamOffset_reverts() (gas: 13396233)
+OmniPortal_xsubmit_Test:test_xsubmit_xblock1_chainB_succeeds() (gas: 13601419)
+OmniPortal_xsubmit_Test:test_xsubmit_xblock1_succeeds() (gas: 13602679)
+OmniPortal_xsubmit_Test:test_xsubmit_xblock2_chainB_succeeds() (gas: 13879461)
+OmniPortal_xsubmit_Test:test_xsubmit_xblock2_succeeds() (gas: 13882068)
```
pavelnikolov, line 30: How was this snapshot generated?

contracts/Makefile
```diff
@@ -9,7 +9,7 @@ build: version ## Build contracts.
 	forge build
 
 bindings: build ## Generate golang contract bindings.
-	./bindings/gen.sh OmniPortal FeeOracleV1
+	./bindings/gen.sh OmniPortal FeeOracleV1 TransparentUpgradeableProxy ProxyAdmin
 
 examples-bindings: build ## Generate golang contract bindings for example contracts
 	BINDINGS_DIR=./bindings/examples ./bindings/gen.sh PingPong
```
pavelnikolov, line 15: This should not be done here. Ideally, we need a go file named `gen.go` or something like that. Then inside of that file we'd use [a Go generate comment](https://go.dev/blog/generate), for example:
```go
//go:generate ../bindings/gen.sh OmniPortal FeeOracleV1 TransparentUpgradeableProxy ProxyAdmin
```
Then before we need to use the go files, we just run `go generate ./...` to execute all the generate commands through out the project.

test/e2e/netman/helpers.go
```diff
@@ -30,32 +30,112 @@ func deployContract(ctx context.Context, chainID uint64, client *ethclient.Clien
...
+		return common.Address{}, errors.Wrap(err, "wait mined fee oracle impl")
+	}
+
+	feeOracleV1ABI, err := bindings.FeeOracleV1MetaData.GetAbi()
+	if err != nil {
+		return common.Address{}, errors.Wrap(err, "get fee oracle v1 abi")
 	}
 
-	portalAddr, tx, _, err := bindings.DeployOmniPortal(txOpts, client, owner, feeOracleAddr, valSetID, validators)
+	encodedFeeOracleV1Initializer, err := feeOracleV1ABI.Pack("initialize", owner, fee)
```
pavelnikolov, line 73: > [Variable names in Go should be short rather than long](https://go.dev/wiki/CodeReviewComments#variable-names). 

Especially, when the variable is only ever used few lines below. A better alternative would be: `feeOracleInit` or even `foi`.

pavelnikolov (approved):

contracts/Makefile
ttarsi: added a task

contracts/.gas-snapshot
ttarsi: We use `forge snapshot`. It is run during pre-commit

test/e2e/netman/helpers.go
ttarsi: @kevinhalliday , what do you think?

test/e2e/netman/helpers.go
ttarsi: Yes, that's not super obvious right now. Agreed, added a ticket for it

test/e2e/netman/helpers.go
ttarsi: We might run into nonce issues – I'm down to remove it and monitor. @kevinhalliday is nonce issues why we added it originally?

kevinhalliday: I found I needed it some cases when deploying fee oracle w/ omni portal, was getting some "no code at address" errors. but we don't need it everywhere

.secrets.baseline
```diff
@@ -188,6 +184,36 @@
         "line_number": 66
       }
     ],
+    "halo2/genutil/testdata/TestMakeGenesis.golden": [
+      {
+        "type": "Base64 High Entropy String",
+        "filename": "halo2/genutil/testdata/TestMakeGenesis.golden",
```
kevinhalliday, line 190: we can probably ignore **/testdata/** folders

test/e2e/netman/helpers.go
kevinhalliday: we get errors on gas estimation

PR #298, opened 2024-02-15
fix(halo2): add engineevm module skeleton

corverroos:
Implements the basic skeleton (including all the protobufs) for the new `engineevm` module

task: none


engineevm/x/engineevm/types/tx.pb.go
ttarsi, line 1: do we have to keep this in the root dir like this?

ttarsi (approved):
🙌

engineevm/x/engineevm/types/tx.pb.go
corverroos: not sure what you mean, the `tx.pb.go` types are usually in the `/x/moduel/types` package, which in this case is `evmengine/types`

corverroos: oh sorry yeah, rouge proto file, will delete

PR #308, opened 2024-02-15
chore(contracts): remove eigen middleware

kevinhalliday:
Remove eigenlayer middleware

task: none


.github/workflows/soltest.yml
```diff
@@ -19,7 +19,9 @@ jobs:
         run: forge fmt --check
         working-directory: contracts
       - name: Check snapshot
-        run: forge snapshot --check --tolerance 10
+        # fuzz tests snapshots are different in CI, so we ignore them
+        # TODO: figure out why, and fix
+        run: forge snapshot --check --tolerance 5 --no-match-test testFuzz_syncWithOmni
```
kevinhalliday, line 24: This fuzz test has been causing too much trouble with snapshots.

kevinhalliday:
This also extends tests to test `xcall` - sorry, could have split into a separate pr

.pre-commit/run_forge_tests.sh
```diff
@@ -11,5 +11,7 @@ source .pre-commit/foundry_utils.sh
 
 for dir in $(foundryroots $@); do
   echo "Running 'forge test' in ./$dir"
-  (cd $dir && pnpm install && forge test && forge snapshot)
+  # fuzz tests snapshots are different in CI, so we ignore them
+  # TODO: figure out why, and fix
+  (cd $dir && pnpm install && forge test && forge snapshot --no-match-test testFuzz_syncWithOmni)
```
ttarsi, line 16: so we're running the test, just not outputting the snapshot data for that test?

contracts/src/protocol/OmniAVS.sol
```diff
@@ -1,106 +1,169 @@
...
- */
-contract OmniAVS is IOmniAVS, IOmniAVSAdmin, ServiceManagerBase, OperatorStateRetriever {
-    /// @dev AVS Quorum numbers, Omni only has one quorum
-    bytes public constant QUORUM_NUMBERS = hex"00";
+contract OmniAVS is IOmniAVS, IOmniAVSAdmin, IServiceManager, OwnableUpgradeable {
+    /// @notice Constant used as a divisor in calculating weights
+    uint256 public constant WEIGHTING_DIVISOR = 1e18;
+
+    /// @dev OmniPortal.xcall gas limit per each validator in syncWithOmni
+    uint256 internal constant _XCALL_GAS_LIMIT_PER_VALIDATOR = 10_000;
```
ttarsi, line 23: - why underscore here?
- should we make these state variables, so they can be updated for the restaking contract?

contracts/src/protocol/OmniAVS.sol
```diff
@@ -1,106 +1,169 @@
...
-    /// @dev AVS Quorum number
-    uint8 public constant QUORUM_NUMBER = 0;
+    // TODO: revisit gas limits when OmniEthRestaking is written, and we can measure gas usage
 
-    /// @dev Omni chain id, used to make xcalls to the Omni chain
+    /// @notice EigenLayer core DelegationManager
+    IDelegationManager public immutable _delegationManager;
+
+    /// @notice Maximum number of operators that can be registered
+    uint32 public maxOperatorCount;
```
ttarsi, line 34: Let's add a function to be able to increase this in the future

contracts/src/protocol/OmniAVS.sol
```diff
@@ -1,106 +1,169 @@
 // SPDX-License-Identifier: GPL-3.0-only
-pragma solidity 0.8.12;
+pragma solidity =0.8.12;
```
ttarsi, line 2: can we change this to `0.8.23` now? We're not importing contracts with this requirement anymore

ttarsi (approved):

.pre-commit/run_forge_tests.sh
kevinhalliday: yes - for now, it's just been a little annoying

PR #309, opened 2024-02-15
feat(e2e): add --image-tag to e2e cmd

pavelnikolov:
This PR adds `--image-tag` to the e2e runner binary so that the generated docker compose files can pin docker images by tag.

task: https://app.asana.com/0/1206208509925075/1206590491637443/f

### DIFF
--- a/test/e2e/tests/e2e_test.go
+++ b/test/e2e/tests/e2e_test.go
@@ -164,7 +164,7 @@ func loadEnv(t *testing.T) (types.Testnet, netconf.Network) {
 	}
 	require.NoError(t, err)
 
-	testnet, err := app.TestnetFromManifest(m, manifestFile, ifd, nil)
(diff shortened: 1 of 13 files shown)


test/e2e/tests/e2e_test.go
```diff
@@ -164,7 +164,7 @@ func loadEnv(t *testing.T) (types.Testnet, netconf.Network) {
 	}
 	require.NoError(t, err)
 
-	testnet, err := app.TestnetFromManifest(m, manifestFile, ifd, nil)
+	testnet, err := app.TestnetFromManifest(m, manifestFile, ifd, nil, "main")
```
ttarsi, line 167: we always want to use `main` here?

ttarsi (approved):
🎉

test/e2e/tests/e2e_test.go
pavelnikolov: This is a test, and the tag is not even used.

PR #311, opened 2024-02-16
chore(contracts): update to eigen m2 fixes

kevinhalliday:
Update to eigen dependency to m2-mainnet-fixes branch

- add AVSDirectory
- trim unnused eigenlayer test utils

task: none


contracts/src/protocol/OmniAVS.sol
```diff
@@ -214,6 +219,11 @@ contract OmniAVS is IOmniAVS, IOmniAVSAdmin, IServiceManager, OwnableUpgradeable
         return strategies;
     }
 
+    /// @inheritdoc IServiceManager
+    function avsDirectory() external view returns (address) {
```
ttarsi, line 223: we should just be able to use the public getter for this. the interface wrapper is just a solidity convenience, on chain it is just the address

ttarsi (approved):

contracts/src/protocol/OmniAVS.sol
kevinhalliday: Yes but solidity complains

kevinhalliday: Complains that we have a different return type than the `IServiceManager`

PR #314, opened 2024-02-16
fix(lib/netconf): format period as string

corverroos:
Improve Chain json

task: none

### DIFF
--- a/lib/netconf/netconf.go
+++ b/lib/netconf/netconf.go
@@ -77,14 +77,14 @@ func (n Network) Chain(id uint64) (Chain, bool) {
 // the Omni cross chain protocol. This is most supported Rollup EVM, but
 // also the Omni EVM.
 type Chain struct {
-	ID            uint64        `json:"id"`                    // Chain ID asa per https://chainlist.org
-	Name          string        `json:"name"`                  // Chain name as per https://chainlist.org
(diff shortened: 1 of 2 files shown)


Lazar955 (approved):

pavelnikolov (changes requested):
I don't think we need this type. If the only purpose is to deserialize in JSON. Let's leave the `json` tags.

lib/netconf/netconf.go
```diff
@@ -115,3 +115,61 @@ func Save(network Network, path string) error {
...
+type chainJSON struct {
+	ID            uint64 `json:"id"`
+	Name          string `json:"name"`
+	RPCURL        string `json:"rpcurl"`
+	AuthRPCURL    string `json:"auth_rpcurl,omitempty"`
+	PortalAddress string `json:"portal_address"`
+	DeployHeight  uint64 `json:"deploy_height"`
+	IsOmni        bool   `json:"is_omni,omitempty"`
+	BlockPeriod   string `json:"block_period"`
+}
```
pavelnikolov, line 128: We don't need this type here. Its only purpose seems to be to serialize/deserialize to/from JSON. Its members match exactly the names of the `Chain` struct. Let's use the `json` tags next to its fields and remove the need for another type.

PR #322, opened 2024-02-16
feat(relayer/txmgr): additional txmgr metrics

Lazar955:
adds more metrics for txmgr 

task: none

### DIFF
--- a/relayer/txmgr/metrics.go
+++ b/relayer/txmgr/metrics.go
@@ -13,4 +13,25 @@ var (
 		Name:      "resend_total",
 		Help:      "The total number of transaction resends to a destination chain",
 	}, []string{"chain"})
+
+	txL1GasFee = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "relayer",
+		Subsystem: "txmgr",
+		Name:      "tx_fee_gwei",
+		Help:      "L1 gas fee for transactions in GWEI",
+	}, []string{"chain"})
+
+	txFees = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "relayer",
+		Subsystem: "txmgr",
+		Name:      "tx_fee_gwei_total",
+		Help:      "Sum of fees spent for all transactions in GWEI",
+	}, []string{"chain"})
+
+	txConfirmationLatency = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "relayer",
+		Subsystem: "txmgr",
+		Name:      "tx_confirmation_latency",
+		Help:      "Latency between transaction submission and confirmation in seconds",
+	}, []string{"chain"})
 )
--- a/relayer/txmgr/metrics.go
+++ b/relayer/txmgr/metrics.go
@@ -14,18 +14,11 @@ var (
 		Help:      "The total number of transaction resends to a destination chain",
 	}, []string{"chain"})
 
-	txL1GasFee = promauto.NewGaugeVec(prometheus.GaugeOpts{
+	txEffectiveGasPrice = promauto.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: "relayer",
 		Subsystem: "txmgr",
 		Name:      "tx_fee_gwei",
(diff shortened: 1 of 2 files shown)


DanFlannel (approved):
LGTM

relayer/txmgr/metrics.go
```diff
@@ -13,4 +13,25 @@ var (
...
+		Name:      "tx_fee_gwei_total",
+		Help:      "Sum of fees spent for all transactions in GWEI",
+	}, []string{"chain"})
+
+	txConfirmationLatency = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "relayer",
+		Subsystem: "txmgr",
+		Name:      "tx_confirmation_latency",
+		Help:      "Latency between transaction submission and confirmation in seconds",
+	}, []string{"chain"})
```
pavelnikolov, line 36: I'm not sure about the nature of this latency, but I'm used to always use histograms for durations. This way we can visualize/calculate the 99th percentile latency, and remove outliers from the graphs (better for alerting). I'm not sure if we care about that. 
In addition to that with a histogram we can display a heatmap graph.
But I'm not sure if that's applicable here. I'm happy to show an example.

pavelnikolov (approved):

relayer/txmgr/txmgr.go
```diff
@@ -359,6 +360,13 @@ func (m *SimpleTxManager) sendTx(ctx context.Context, tx *types.Transaction) (*t
 			return nil, errors.Wrap(ctx.Err(), "context canceled")
 
 		case receipt := <-receiptChan:
+			if receipt.EffectiveGasPrice == nil {
+				return receipt, nil
+			}
+			fee := float64(receipt.EffectiveGasPrice.Uint64() * receipt.GasUsed / params.GWei)
```
corverroos, line 366: lets rather track effective gas price as a gauge.

relayer/txmgr/txmgr.go
```diff
@@ -437,6 +446,7 @@ func (m *SimpleTxManager) waitForTx(ctx context.Context, tx *types.Transaction,
 	}
 	select {
 	case receiptChan <- receipt:
+		txConfirmationLatency.WithLabelValues(m.chainName).Set(time.Since(t).Seconds())
```
corverroos, line 449: put this above the `select`.

relayer/txmgr/metrics.go
corverroos: So yes, normally latencies should be a histogram. Especially if the individual measurement are indiepedent of each other. Your HTTP request and my HTTP request is concurrent and unrelated.

But in blocks chain, we do things sequentially. One after the other. In this case, txs are sent and then confirmed one after the other. So the lag/latency is correlated. The second cannot confirm until the first is confirmed.

So a gauge is fine and simpler in this case.

relayer/txmgr/metrics.go
```diff
@@ -13,4 +13,25 @@ var (
 		Name:      "resend_total",
 		Help:      "The total number of transaction resends to a destination chain",
 	}, []string{"chain"})
+
+	txL1GasFee = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "relayer",
+		Subsystem: "txmgr",
+		Name:      "tx_fee_gwei",
+		Help:      "L1 gas fee for transactions in GWEI",
```
corverroos, line 21: I think this should actually be a histogram, since individual submissions's fee are uncorrelated. And I would rather just track `GasUsed`.

corverroos (approved):
TL;DR:

Lets rather track: 
   - histo:`gas_used` (histo because they are uncorrelated)
   - gauge:`effective_gas_price` (gauge because they are correlated)

relayer/txmgr/metrics.go
pavelnikolov: Understood and I agree about that - the current value is enough to monitor the system health as a whole. But still in alerts we can track 95th or 99th percentile latency and ignore sporadical outliers.

PR #323, opened 2024-02-16
refactor(contracts): remove eigen-core import

ttarsi:
Add an unexported function to a custom `IDelegationManager` to avoid importing the whole `DelegationManager` in our AVS contract. Ideally, Eigen adds this function to the interface

task: none


kevinhalliday (approved):
Nice

PR #327, opened 2024-02-19
test(e2e): introduce deployinfo for tests of contracts

corverroos:
Adds the `DeployInfos` type that is populated with deploy xdapps and contract info. This is then passed to the e2e tests so we can add tests for xdapps. 

task: none


Lazar955 (approved):

kevinhalliday (approved):
Lovely

PR #339, opened 2024-02-19
chore(relayer/txmgr): moving txmgr to lib

DanFlannel:
`omni/relayer/txmgr` -> `omni/lib/txmgr`

task: none


pavelnikolov:
@DanFlannel Is this the only change? If so was this accidentally marked as `draft` or are there more changes to be added?

pavelnikolov (approved):

DanFlannel:
Got impatient and realized I should wait for signals to pass before marking for review haha

PR #342, opened 2024-02-20
test(contracts): remove mock avs directory

kevinhalliday:
Remove mock directory. Have test operators submit signatures.

task: none


ttarsi (approved):
nice

PR #343, opened 2024-02-20
fix(e2e): reset anvil loader to block 0

ttarsi:
The `xprovider` watches a chain's portal from a provided start block. We set our private anvil chains to start at block 0 (via portal `DeployInfo`). When we loaded anvil state for the Eigenlayer deployments, it started from block 12. Then our `xprovider` was watching for block 0, and never processing any `XBlocks`.

The solution is to just set the anvil pre-loaded chain to start at block 0. Another solution could be to set portal `DeployInfo` once the portal is actually deployed, but this is a more difficult – as we don't actually spin up the anvil chains until after setting the `network.json`

task: none

### DIFF
--- a/test/e2e/docker/compose.yaml.tmpl
+++ b/test/e2e/docker/compose.yaml.tmpl
@@ -44,7 +44,7 @@ services:
       e2e: true
     container_name: {{ .Chain.Name }}
     image: ghcr.io/foundry-rs/foundry:latest
-    entrypoint: ['anvil','--host','0.0.0.0','--chain-id','{{ .Chain.ID }}','--block-time','1', '--silent', {{ if .LoadState }}'--load-state','/anvil/state.json'{{ end }}]
+    entrypoint: ['anvil','--host','0.0.0.0','--chain-id','{{ .Chain.ID }}','--block-time','1', {{ if .LoadState }}'--load-state','/anvil/state.json'{{ end }}]
     ports:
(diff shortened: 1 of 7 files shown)


test/e2e/docker/compose.yaml.tmpl
```diff
@@ -44,7 +44,7 @@ services:
       e2e: true
     container_name: {{ .Chain.Name }}
     image: ghcr.io/foundry-rs/foundry:latest
-    entrypoint: ['anvil','--host','0.0.0.0','--chain-id','{{ .Chain.ID }}','--block-time','1', '--silent', {{ if .LoadState }}'--load-state','/anvil/state.json'{{ end }}]
+    entrypoint: ['anvil','--host','0.0.0.0','--chain-id','{{ .Chain.ID }}','--block-time','1', {{ if .LoadState }}'--load-state','/anvil/state.json'{{ end }}]
```
ttarsi, line 47: no need to make this silent. docker logs are much more helpful

kevinhalliday (approved):

test/e2e/docker/compose.yaml.tmpl
corverroos: the problem is that is spams tons and tons of logs which then make debugging failed e2e tests basically impossible as it overloads the github UI. Best would be to make it verbose when debugging locally but make it silent when running in CI

PR #344, opened 2024-02-20
feat(contracts): add operator allowlist

kevinhalliday:
Add operator allowlist to AVS.

task: https://app.asana.com/0/1206208509925075/1206623482381792


kevinhalliday:
We also have a ticket to "eject" an opeator. Should we make the function `removeFromAllowlist` also unregister the operator from eigen core?

contracts/src/interfaces/IOmniAVSAdmin.sol
```diff
@@ -69,4 +76,25 @@ interface IOmniAVSAdmin {
...
+     * @param operator The operator to add
+     */
+    function addToAllowlist(address operator) external;
+
+    /**
+     * @notice Remove an operator from the allowlist.
+     * @dev Only the owner can call this function.
+     * @param operator The operator to remove
+     */
+    function removeFromAllowlist(address operator) external;
```
kevinhalliday, line 99: Dunno if we need all of these - maybe just add & remove? Or just set?

contracts/src/protocol/OmniAVS.sol
```diff
@@ -49,6 +46,12 @@ contract OmniAVS is IOmniAVS, IOmniAVSAdmin, IServiceManager, OwnableUpgradeable
     /// @dev OmniPortal.xcall base gas limit in syncWithOmni
     uint256 internal xcallBaseGasLimit = 75_000;
 
+    /// @notice List of currently register operators, used to sync EigenCore
+    address[] internal _operators;
```
kevinhalliday, line 50: internal saves gas, public getter is not very useful (it's by index)

contracts/src/protocol/OmniAVS.sol
```diff
@@ -197,19 +203,69 @@ contract OmniAVS is IOmniAVS, IOmniAVSAdmin, IServiceManager, OwnableUpgradeable
...
+    /// @inheritdoc IOmniAVSAdmin
+    function removeFromAllowlist(address operator) external onlyOwner {
+        for (uint256 i = 0; i < _allowlist.length; i++) {
+            if (_allowlist[i] == operator) {
+                _allowlist[i] = _allowlist[_allowlist.length - 1];
+                _allowlist.pop();
+                return;
+            }
+        }
+        revert("OmniAVS: not in allowlist");
```
kevinhalliday, line 226: We revert here, but don't revert in `removeOperator`, and check if the operator exists first in another loop. I think we should follow a single pattern - which one do you prefer?

ttarsi (changes requested):
- should be a mapping for gas costs, the _isInAllowList func is frequently used and expensive
- probably use openzeppelin access control?
- just add and remove makes sense, no need to set the full list

ttarsi:
> We also have a ticket to "eject" an opeator. Should we make the function `removeFromAllowlist` also unregister the operator from eigen core?

let's discuss sync, couple moving parts here

kevinhalliday:
Yeah I guess we could just go mapping + events to get list of allowlisted operators

contracts/src/interfaces/IOmniAVSAdmin.sol
```diff
@@ -9,13 +9,26 @@ import { IOmniAVS } from "./IOmniAVS.sol";
...
+     * @param operator The operator
+     */
+    event OperatorAddedToAllowlist(address operator);
+
+    /**
+     * @notice Emitted when an operator is removed from the allowlist.
+     * @param operator The operator
+     */
+    event OperatorRemovedFromAllowlist(address operator);
+
```
kevinhalliday, line 23: Long even names 🤔

ttarsi: Lol how about:
- `OperatorAllowed`
- `OperatorDisallowed`

ttarsi (approved):

jmozah:
> We also have a ticket to "eject" an opeator. Should we make the function `removeFromAllowlist` also unregister the operator from eigen core?

I think, given that we have only controlled release at first,  It should be enough if we just pause the operator.

kevinhalliday:
I'll merge and we can checkout access control in a separate pr

ttarsi:
@kevinhalliday regen bindings 🙏

PR #348, opened 2024-02-20
fix(e2e): avs defaults

corverroos:
Refactors e2e AVS to use sensible defaults:
- Embed AVS `anvil_state.json` and `el_deployments.json` in e2e app in `static` dir.
- Load that anvil state in all anvils always (no reason not to?)
- Use embedded el_deployments if non provided.

When running e2e, no flags are requried.
When deploying staging, we just need to provide a `eigen-file` flag.

Also:
- Add a `make build-halo-relayer` for faster local docker image builds.
- Add a `make fix-golden` to fix golden test fixtures.
- Rename all embedded files to use dashes instead of underscore

task: none


Lazar955 (approved):

ttarsi (commented):
very nice!

PR #352, opened 2024-02-20
chore(lib/metrics): fix log and git metrics

corverroos:
Fixes `lib/log` and `lib/gitinfo` metrics. Also enabled cosmosSDK metrics.

task: none

### DIFF
--- a/halo/app/start.go
+++ b/halo/app/start.go
@@ -30,6 +30,7 @@ import (
 	dbm "github.com/cosmos/cosmos-db"
 	"github.com/cosmos/cosmos-sdk/baseapp"
 	"github.com/cosmos/cosmos-sdk/server"
+	sdktelemetry "github.com/cosmos/cosmos-sdk/telemetry"
 	"github.com/cosmos/cosmos-sdk/types/mempool"
 )
 
@@ -73,6 +74,10 @@ func Start(ctx context.Context, cfg Config) (func(context.Context) error, error)
 	log.Info(ctx, "Starting halo consensus client")
 
 	gitinfo.Instrument(ctx)
+
+	if err := enableSDKTelemetry(); err != nil {
+		return nil, errors.Wrap(err, "enable cosmos-sdk telemetry")
+	}
 
 	// Load private validator key and state from disk (this hard exits on any error).
 	privVal := privval.LoadFilePV(cfg.Comet.PrivValidatorKeyFile(), cfg.Comet.PrivValidatorStateFile())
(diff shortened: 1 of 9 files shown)


halo/app/start.go
```diff
@@ -268,3 +273,19 @@ func newEngineClient(ctx context.Context, cfg Config, network netconf.Network) (
...
 }
+
+// enableSDKTelemetry enables prometheus based cosmos-sdk telemetry.
+func enableSDKTelemetry() error {
+	const farFuture = time.Hour * 24 * 365 * 10 // 10 years ~= infinity.
+
+	_, err := sdktelemetry.New(sdktelemetry.Config{
+		ServiceName:             "cosmos",
+		Enabled:                 true,
+		PrometheusRetentionTime: int64(farFuture.Seconds()), // Prometheus metrics never expire once created in-app.
```
pavelnikolov, line 284: Don't we scrape them every 30s? Then we "ship" the metrics to Grafana - there we have "infinite" retention. I don't think we need metrics retention in the app as well.

lib/gitinfo/metrics.go
```diff
@@ -1,19 +1,22 @@
...
 
-	timestampGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+	timestampGauge = promauto.NewGauge(prometheus.GaugeOpts{
 		Namespace: "lib",
 		Subsystem: "git",
-		Name:      "timestamp",
-		Help:      "Constant gauge with label 'timestamp' set to the current git commit timestamp.",
-	}, []string{"timestamp"})
+		Name:      "timestamp_unix",
+		Help:      "Build info git commit timestamp in unix seconds.",
```
pavelnikolov, line 20: We need only one gauge - `up` and it should have all the other data as labels: `git_commit`, `go_version`, and other important versions, e.g. cosmos SDK version etc.

pavelnikolov (approved):

Lazar955 (approved):

halo/app/start.go
corverroos: Some metrics never change ,like `up` and `git_commit_hash`

lib/gitinfo/metrics.go
corverroos: `up` is built in, can't change it

pavelnikolov: ok, then we use the build in, but we still need only one gauge, e.g. `build_info` and add all other labels to that one gauge. If we use the built-in it should have the go version out of the box.

PR #354, opened 2024-02-20
chore(relayer): improve logging

corverroos:
Improve relayer logging and metrics.

task: none

### DIFF
--- a/halo/attest/keeper/keeper.go
+++ b/halo/attest/keeper/keeper.go
@@ -130,7 +130,12 @@ func (k Keeper) addOne(ctx context.Context, agg *types.AggAttestation) error {
 		})
 		if errors.Is(err, ormerrors.UniqueKeyViolation) {
 			// TODO(corver): We should prevent this from happening earlier.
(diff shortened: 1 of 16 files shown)


Lazar955 (approved):
gj

halo/attest/keeper/keeper.go
```diff
@@ -130,7 +130,12 @@ func (k Keeper) addOne(ctx context.Context, agg *types.AggAttestation) error {
 		})
 		if errors.Is(err, ormerrors.UniqueKeyViolation) {
 			// TODO(corver): We should prevent this from happening earlier.
-			log.Warn(ctx, "Ignoring duplicate attestation", nil, "agg_id", aggID, "validator", sig.ValidatorAddress)
+			log.Warn(ctx, "Ignoring duplicate attestation", nil,
+				"agg_id", aggID, "",
```
pavelnikolov, line 134: Correct me if I'm wrong, but aren't we supposed to have alternating key/value pairs as the log args?
Why do we have 3 values passed here? Am I missing something?

PR #359, opened 2024-02-21
fix(readability): stop requiring hard line length limit

pavelnikolov:
This PR removes the hard line length limit for Go code.

task: none


ttarsi (approved):
lgtm just make sure our pre-commit hooks and ci-linter match, looks like a couple are failing

PR #360, opened 2024-02-21
test(contracts): support avs deployment fork test

kevinhalliday:
Refactor AVS deployment & tests to support fork tests of same script used for deployment.

Plus general cleanup.

task: https://app.asana.com/0/1206208509925075/1206644163432112


contracts/test/avs/AVSBase.sol
```diff
@@ -2,72 +2,89 @@
...
 import { IOmniAVS } from "src/interfaces/IOmniAVS.sol";
-import { OmniAVSHarness } from "./OmniAVSHarness.sol";
+import { OmniAVS } from "src/protocol/OmniAVS.sol";
 import { EigenLayerTestHelper } from "./eigen/EigenLayerTestHelper.t.sol";
+import { EigenM2GoerliDeployments } from "./eigen/EigenM2GoerliDeployments.sol";
 import { MockPortal } from "./MockPortal.sol";
 
+import { DeployGoerliAVS } from "script/avs/DeployGoerliAVS.s.sol";
+import { StrategyParams } from "script/avs/StrategyParams.sol";
+import { AVSDeploy } from "script/avs/AVSDeploy.sol";
```
ttarsi, line 18: Let's call this `AVSDeployer` so that it's clear this is different than `DeployGoerliAVS`. If it's a library (same with `StrategyParams`), shouldn't it live in `libraries`?

ttarsi: Maybe instead we just have the `AVSDeployer` include logic for deploying to Goerli vs. Local? The we could just call: `deployer.DeployGoerli(...)` or `deployer.DeployLocal(...)`. All this inheritance and importing is making my brain hurt

ttarsi (approved):

contracts/test/avs/AVSBase.sol
kevinhalliday: Yeah I think there's a way to make this clearer. Let's talk through it sync tomorrow? 


We might not even need the AVSDeploy lib. It's purpose was just to make sure same code was used for deploying everywhere. But for the most part it just forwards params.

PR #363, opened 2024-02-21
fix(halo/attest): fix bug in attestation aggregation

corverroos:
Fixes a bug in `aggregateAtts`. We assumed `BlockHashes` are unique across chains, which is probably the case in the wild but definitely not the case for anvil chains in e2e. Use `BlockHeader` for uniqueness instead.

Also use the cosmos sdk `baseapp.ValidateVoteExtensions`. This requires moving the attestation CProvider to attest keeper proper. And makes evmkeeper depend on attest keeper so had to change deps.

Also improve relayer worker reset backoff and logs.
Also improve e2e test logs.

task: none

### DIFF
--- a/halo/attest/keeper/cpayload.go
+++ b/halo/attest/keeper/cpayload.go
@@ -1,4 +1,4 @@
-package attester
+package keeper
 
 import (
 	"bytes"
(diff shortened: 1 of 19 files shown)


halo/attest/keeper/cpayload.go
```diff
@@ -57,10 +65,10 @@ func aggregatesFromLastCommit(info abci.ExtendedCommitInfo) (*types.MsgAggAttest
 
 // aggregateAtts aggregates the provided attestations by block header.
 func aggregateAtts(atts []*types.Attestation) []*types.AggAttestation {
-	aggsByHash := make(map[string]*types.AggAttestation) // map[BlockHash]AggAttestation
+	aggsByHeader := make(map[xchain.BlockHeader]*types.AggAttestation) // map[BlockHash]AggAttestation
 	for _, att := range atts {
-		hashStr := string(att.BlockHeader.Hash)
```
corverroos, line 62: this was the bug

Lazar955 (approved):

halo/attest/keeper/cpayload.go
```diff
@@ -8,22 +8,27 @@ import (
...
 
-// CPayloadProvider implements the CPayloadProvider interface for the attester module.
-// It extracts the aggregate attestations from the vote extensions of the last local commit.
-type CPayloadProvider struct {
-}
+func (k Keeper) PreparePayload(ctx context.Context, height uint64, commit abci.ExtendedCommitInfo) ([]sdk.Msg, error) {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	if err := baseapp.ValidateVoteExtensions(sdkCtx, k.skeeper, int64(height), sdkCtx.ChainID(), commit); err != nil {
+		log.Error(ctx, "Cannot include invalid vote extensions in payload", err, "height", height)
+		return nil, nil
```
pavelnikolov, line 29: Is it intentional that we log an error, however, we return `nil` for the `error` return value? I find this a little confusing.

PR #368, opened 2024-02-22
test(contracts): test beacon eth deposits

kevinhalliday:
Test our AVS handles beacon eth stake &  delegations.

task: none


kevinhalliday:
Marking as draft because I thought this would break goerli fork tests, but it doesn't. So need to figure out why

contracts/src/protocol/OmniAVS.sol
```diff
@@ -220,7 +220,9 @@ contract OmniAVS is IOmniAVS, IOmniAVSAdmin, IServiceManager, OwnableUpgradeable
 
             uint96 total = _getTotalDelegations(operator);
             uint96 staked = _getSelfDelegations(operator);
-            uint96 delegated = total - staked;
+
+            // this should never happen, but just in case
+            uint96 delegated = total > staked ? total - staked : 0;
```
kevinhalliday, line 225: @ttarsi thoughts? this happened in a tests cause I was manipulating state in eigenpod manager harness, but not correctly

contracts/test/avs/OmniAVS.t.sol
```diff
@@ -102,12 +109,16 @@ contract OmniAVS_Test is AVSBase, AVSUtils {
...
         // assert operator has initial stake
         for (uint32 i = 0; i < numOperators; i++) {
-            assertEq(validators[i].staked, initialOperatorStake);
-            assertEq(validators[i].delegated, 0);
+            assertEq(
+                validators[i].staked,
+                initialOperatorStake,
+                "_testRegisterOperatorsWithAVS: validator.staked should be initialOperatorStake"
+            );
+            assertEq(validators[i].delegated, 0, "_testRegisterOperatorsWithAVS: validator.delegated should be 0");
```
kevinhalliday, line 121: added a bunch of these assert error messages because they help with debugging in this big ole tests. we should have them in more places

contracts/test/avs/eigen/deploy/EigenLayerLocal.sol
```diff
@@ -207,7 +217,22 @@ contract EigenLayerLocal is IEigenDeployer {
...
+
+        for (uint256 i = 0; i < strats.length; i++) {
+            _strategy[i] = IStrategy(strats[i]);
+            _thirdPartyTransfersForbiddenValues[i] = false;
+        }
+
+        strategyManager.addStrategiesToDepositWhitelist(_strategy, _thirdPartyTransfersForbiddenValues);
+
+        vm.stopPrank();
+    }
```
kevinhalliday, line 237: Whitelisting all strategies in advance, so saved anvil state has our local strategies whitelisted.

Before, we were only doing this on deposit (if strat not whitelisted, whitelist it, then deposit)

contracts/src/protocol/OmniAVS.sol
ttarsi: Yes, validate all data not bounded by our own system

ttarsi (approved):
lgtm

PR #369, opened 2024-02-22
refactor(halo/attest): implement validator set hash and rename types

corverroos:
Big refactor of attest module:
- Use ValidatorSetHash instead of VaSetID since hash is readily available while ID is not and would need to a dedicated module/keeper.
- Rename types:
  - Individual validators "vote" for xblocks (instead of attest).
  - Once sufficient "votes" are collected, you have an approved "attestation".

task: none


halo/attest/voter/vote.go
```diff
@@ -9,8 +9,8 @@ import (
 	"github.com/cometbft/cometbft/crypto"
 )
 
-// CreateAttestation creates an attestation for the given block.
-func CreateAttestation(privKey crypto.PrivKey, block xchain.Block) (*types.Attestation, error) {
+// CreateVote creates an attestation for the given block.
```
Lazar955, line 12: nit comment update

Lazar955 (approved):

PR #371, opened 2024-02-22
feat(halo/evmengine): trigger optimistic build on finalise

corverroos:
Trigger evm block building on previous block finalise if we are the next proposer. This allows much more time for block building. This will keep our block times to a minimum.

Also adds the following two halo flags:
- `--evm-build-delay`
- `--evm-build-optimistic`

task: none


halo/evmengine/keeper/abci.go
```diff
@@ -47,53 +46,63 @@ func (k *Keeper) PrepareProposal(ctx sdk.Context, req *abci.RequestPreparePropos
...
+		// TODO(corver): Maybe error if timestamp is not greater than latest execution block.
+		timestamp := uint64(req.Time.Unix())
+		if timestamp <= latestEBlock.Time() {
+			timestamp = latestEBlock.Time() + 1
+		}
 
-	payloadAttrs := engine.PayloadAttributes{
-		Timestamp:             timestamp,
-		Random:                latestEBlock.Hash(),                        // TODO(corver): implement proper randao.
-		SuggestedFeeRecipient: common.BytesToAddress(req.ProposerAddress), // TODO(corver): Ensure this is correct.
```
corverroos, line 79: this was a bug, `req.ProposerAddress` is not a ethereum address.

halo/evmengine/keeper/proposal_server.go
```diff
@@ -14,13 +14,11 @@ type proposalServer struct {
 // ExecutionPayload handles a new execution payload proposed in a block.
 func (s proposalServer) ExecutionPayload(ctx context.Context, msg *types.MsgExecutionPayload,
 ) (*types.ExecutionPayloadResponse, error) {
-	_, err := newPayload(ctx, s.ethCl, msg)
+	_, err := pushPayload(ctx, s.ethCl, msg)
 	if err != nil {
 		return nil, err
 	}
 
-	// TODO(corver): If this validator is proposer for next block, we should trigger block building.
```
corverroos, line 22: we can't actually do this on `ProcessProposal` since EgineAPI only allows building on head, which is being proposed at this point.

PR #374, opened 2024-02-22
fix(halo/attest): fix approve bug and add metric

corverroos:
Fix attestation approval to not skip heights when approving. Height X+1 must not be approved until height X is approved.

Also add approved height metric. 

task: none

### DIFF
--- a/halo/attest/module/module.go
+++ b/halo/attest/module/module.go
@@ -118,6 +118,7 @@ type ModuleInputs struct {
 	EthClient    engine.API
 	SKeeper      *skeeper.Keeper
 	Voter        types.Voter
+	Namer        types.ChainNameFunc
 }
 
 type ModuleOutputs struct {
@@ -134,6 +135,7 @@ func ProvideModule(in ModuleInputs) (ModuleOutputs, error) {
 		in.EthClient,
 		in.SKeeper,
 		in.Voter,
+		in.Namer,
 	)
 	if err != nil {
 		return ModuleOutputs{}, err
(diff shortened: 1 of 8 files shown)


pavelnikolov (approved):

halo/attest/module/module.go
```diff
@@ -134,6 +135,7 @@ func ProvideModule(in ModuleInputs) (ModuleOutputs, error) {
 		in.EthClient,
 		in.SKeeper,
 		in.Voter,
+		in.Namer,
```
pavelnikolov, line 138: Do we need the Namer as output as well?

halo/attest/types/types.go
```diff
@@ -0,0 +1,4 @@
+package types
+
+// ChainNameFunc is a function that returns the name of a chain given its ID.
+type ChainNameFunc func(chainID uint64) string
```
pavelnikolov, line 4: This should not be a type since we do not do anything with this type. A good example would be `http.HandlerFunc` - it is used as `http.Handler`. However, in this case we do not do anything with this function type. This is why I'd replace it with an interface - see comment above ^

pavelnikolov (changes requested):
Please, use interface instead of function type.

corverroos:
@pavelnikolov func types are a common pattern in go as a replacement of single function interfaces, specifically in cosmosSDK, see [example](https://github.com/cosmos/cosmos-sdk/blob/main/types/abci.go).

pavelnikolov (approved):

PR #376, opened 2024-02-22
chore(e2e): moving sum function to dedicated file

DanFlannel:
We shouldn't have a random math function in `run.go`

task: none

### DIFF
--- a/lib/math/sum.go
+++ b/lib/math/sum.go
@@ -1,0 +1,10 @@
+package math
+
+func Sum(batches []int) uint64 {
+	var resp int
+	for _, b := range batches {
+		resp += b
+	}
+
+	return uint64(resp)
+}
--- a/test/e2e/app/math.go
+++ b/test/e2e/app/math.go
@@ -1,4 +1,4 @@
-package math
+package app
 
 func Sum(batches []int) uint64 {
 	var resp int
--- a/test/e2e/app/run.go
+++ b/test/e2e/app/run.go
@@ -7,7 +7,6 @@ import (
 	"github.com/omni-network/omni/lib/errors"
 	"github.com/omni-network/omni/lib/k1util"
 	"github.com/omni-network/omni/lib/log"
-	"github.com/omni-network/omni/lib/math"
 	"github.com/omni-network/omni/test/e2e/netman/pingpong"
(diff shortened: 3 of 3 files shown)


lib/math/sum.go
```diff
@@ -0,0 +1,10 @@
+package math
+
+func Sum(batches []int) uint64 {
```
pavelnikolov, line 3: This should be generic, IMHO:
```go
func Of[TS ~[]T, T Number](elements TS) T {
	return slice.Sum(elements)
}
```

You will need `Number`:
```go
import "golang.org/x/exp/constraints"

type Number interface {
    constraints.Integer | constraints.Float
}
```

pavelnikolov: But maybe an overkill for our needs : )

ttarsi (commented):
i'm not sure we need to add a new separate package for this – "a little copying is better than a little dependency" (plus we don't even have to copy since it's only used once) – but i will default to our gophers 🦔

pavelnikolov:
Agree with @ttarsi on this. Adding a package is fine, adding a dependency not so much 👍 . Let's skip the generics until we actually need to use it with multiple types for now.

DanFlannel:
I just don't like the random function in the `run.go` class 🤷 what about putting it in a file in the same dir?

test/e2e/app/math.go
```diff
@@ -0,0 +1,10 @@
+package app
+
+func Sum(batches []int) uint64 {
```
pavelnikolov, line 3: Sum doesn't need to be exported.
```suggestion
func sum(batches []int) uint64 {
```

test/e2e/app/math.go
```diff
@@ -0,0 +1,10 @@
...
+	var resp int
+	for _, b := range batches {
+		resp += b
+	}
+
+	return uint64(resp)
+}
```
pavelnikolov, line 10: Sum doesn't need to be exported. Also it doesn't need to know that it sums batches (it can be any slice of ints).
```suggestion
func sum(v []int) uint64 {
	var res int
	for _, n := range v {
		res += n
	}

	return uint64(res)
}
```

pavelnikolov (approved):

PR #377, opened 2024-02-22
chore(e2e): moving pingpong trigger to deploy config

DanFlannel:
removing `DeployWithPingPong` in favor of adding field to our deploy config definition

task: none


pavelnikolov (approved):

test/e2e/app/run.go
```diff
@@ -16,31 +16,7 @@ import (
...
-	if err != nil {
-		return nil, errors.Wrap(err, "deploy pingpong")
-	} else if err := pp.StartAllEdges(ctx, pingPongN); err != nil {
-		return nil, errors.Wrap(err, "start all edges")
-	}
-
-	pp.ExportDeployInfo(deployInfo)
-
-	return deployInfo, nil
+	PingPongN *uint64
```
corverroos, line 19: avoid pointers. use zero values instead. If PingPong==0 skip it.

test/e2e/cmd/cmd.go
```diff
@@ -69,13 +69,15 @@ func New() *cobra.Command {
 
 func newDeployCmd(def *app.Definition) *cobra.Command {
 	var cfg app.DeployConfig
+
 	pingPongN := defaultPingPongDeploy // Default to 1000 ping pongs.
+	cfg.PingPongN = &pingPongN
```
corverroos, line 74: so now you'll need to align this with our general config/flag pattern:
- default default flag values in `DefaultDeployConfig` function (in app pkg next to type def).
- Define flags of config struct in one place. So move `--ping-poing` below into `bindDeployFlags`

DanFlannel: Awesome! I'll work on this "tomorrow" wanted to ask what was the right way on how to do it was but I didn't want to create noise too late at night for everyone

DanFlannel: @corverroos done, if you could check to confirm it that would be great

corverroos (approved):

test/e2e/cmd/cmd.go
corverroos: Cool, looks good I added ` app.DefaultDeployConfig` function to align with our flags pattern. Also only bound `PromSecrets` during E2E fags.

PR #379, opened 2024-02-22
fix(contracts/test): fix goerli fork tests

kevinhalliday:
Fix goerli fork tests.

- fix goerli fork "deployment"
- allow tests on goerli strategies (like stETH and rETH)
- handle strategy shares vs. underlying
- add avs fork tests make target

task: none

### DIFF
--- a/contracts/foundry.toml
+++ b/contracts/foundry.toml
@@ -19,7 +19,6 @@ fs_permissions = [
 
 [fuzz]
 seed = "0x6eed"
-runs = 10
 
 [fmt]
 bracket_spacing = true
--- a/contracts/foundry.toml
+++ b/contracts/foundry.toml
@@ -19,6 +19,7 @@ fs_permissions = [
 
 [fuzz]
 seed = "0x6eed"
+runs = 10
 
 [fmt]
 bracket_spacing = true
--- a/contracts/script/avs/DeployGoerliAVS.s.sol
+++ b/contracts/script/avs/DeployGoerliAVS.s.sol
@@ -15,8 +15,6 @@ import { StrategyParams } from "./StrategyParams.sol";
 
 import { Script } from "forge-std/Script.sol";
 
-// solhint-disable const-name-snakecase
-
 /**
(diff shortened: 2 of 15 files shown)


contracts/foundry.toml
```diff
@@ -19,6 +19,7 @@ fs_permissions = [
 
 [fuzz]
 seed = "0x6eed"
+runs = 10
```
ttarsi, line 22: just 10?

contracts/script/avs/DeployGoerliAVS.s.sol
```diff
@@ -52,13 +50,25 @@ contract DeployGoerliAVS is Script {
...
                 allowlist,
                 StrategyParams.goerli()
             )
         );
 
         return proxy;
     }
+
+    /// @dev deploy OmniAVS, but with a prank. necessary because we cannot
+    //       vm.startPrank() outside of this contract
```
ttarsi, line 64: I thought if `vm.startPrank` is called, all txns are called with that address until stopped – regardless of where the calls happen

ttarsi (approved):

contracts/script/avs/DeployGoerliAVS.s.sol
kevinhalliday: I did too so I wrote it that way first. But it didn't work. It's just all calls from the contract in which `vm.startPrank` was called. But that actually makes sense. You don't want `vm.startPrank()` to affect all `msg.sender`s in a chain of function calls

So

```
contract A {
     function foo() {
        vm.startPrank(me);
        b.foo();
    }
}

contract B {
    function bar() {
        // msg.sender == me
        c.baz()
   }
}


contract C {
   function baz() {
       // msg.sender == address(b)
   }
}
```

contracts/foundry.toml
kevinhalliday: Did this cause fork tests were very slow let me change just to do 10 runs for fork tests

PR #389, opened 2024-02-24
chore(contracts): simplify and test xapp

kevinhalliday:
Simplify xapp contract, test it in e2e.

- remove onlyXCall
- change setXMsg to xfunc
- update pingpong to use xapp, so e2e tests it

task: none


ttarsi (approved):
Eat your own dog food 😤

Looks awesome, love how simple it makes these calls

PR #390, opened 2024-02-24
chore(docs): add algolia search index

idea404:
This PR couples the Algolia crawler with the docs site search box. It commit an `appID` for the crawler and an `apiKey` which is read-only and [safe to commit](https://docusaurus.io/docs/search#connecting-algolia). 

This PR also adds the omni banner to the docusaurus site static images as it is used in `themeConfig` and the home page. 

task: none


ttarsi (approved):
looks good – approving but disabled auto-merge because just wanted to confirm if the search bar should be working on that preview link? It doesn't appear to be

idea404:
@ttarsi that's right, the search bar currently does not work for two reasons which are both addressed by this PR:

1. The search bar was not coupled to the Algolia crawler with the `appId` and `apiKey`
2. The `url` field in the config prevented the crawler from crawling the site because it was not referring to the docs site URL, yielding the following error for the crawler on the crawl attempt: 

```
🟠 Ignored 
GET https://docs.omni.network/ 
HTTP 
200
This page had a canonical URL (https://your-docusaurus-site.example.com/), which was skipped for this reason: "domain_not_allowed"
LOGS 
Nothing was logged from recordExtractor 
RECORDS 
No records extracted 
LINKS 
Found 0 link matching your configuration 
```

PR #391, opened 2024-02-24
refactor(e2e/send): rename sender to backends

corverroos:
Refactor `send.Sender` to `backend.Backends` to align more with its use as a bindings "Backend". It is used for more things than sendings. It is a general bindings backend.

Also adds a handy struct to wait for multiple transactions to be mined in tests.

task: none


Lazar955 (approved):

idea404 (approved):
nice
*and adds a handy struct to wait for multiple transactions to be mined in tests

PR #394, opened 2024-02-25
refactor(lib/txmgr): check if tx is pending first

pavelnikolov:
According to [the Ethereum docs](https://ethereum.org/en/developers/docs/apis/json-rpc#eth_gettransactionreceipt) a receipt is not available for a pending transaction. This PR refactors the txmgr so that it first checks if a transaction is pending before trying trying to get a receipt for it. The benefit of this is that we check it in a type-safe way. Otherwise, our code depends on an error message that is not documented and subject to change. If Ethereum's  error message changes, for example, from `transaction indexing is in progress` to `pending transaction` our code would stop working silently and we'd have no way to detect that.

task: none

### DIFF
--- a/lib/txmgr/txmgr.go
+++ b/lib/txmgr/txmgr.go
@@ -56,19 +56,11 @@ type TxManager interface {
 // ETHBackend is the set of methods that the transaction manager uses to resubmit gas & determine
 // when transactions are included on L1.
 type ETHBackend interface {
-	// BlockNumber returns the most recent block number.
-	BlockNumber(ctx context.Context) (uint64, error)
-
-	// CallContract executes an eth_call against the provided contract.
-	CallContract(ctx context.Context, msg ethereum.CallMsg, blockNumber *big.Int) ([]byte, error)
-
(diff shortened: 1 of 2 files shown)


lib/txmgr/txmgr.go
```diff
@@ -495,16 +484,25 @@ func (m *SimpleTxManager) queryReceipt(ctx context.Context, txHash common.Hash,
 	sendState *SendState) (*types.Receipt, bool, error) {
 	ctx, cancel := context.WithTimeout(ctx, m.cfg.NetworkTimeout)
 	defer cancel()
+
+	// there are no receipts for pending transactions, therefore check if the
+	// transaction is pending first and only then proceed
+	_, pending, err := m.backend.TransactionByHash(ctx, txHash)
```
corverroos, line 490: See [geth code](https://github.com/ethereum/go-ethereum/blob/fe91d476ba3e29316b6dc99b6efd4a571481d888/core/blockchain_reader.go#L282), this call can also return "transaction indexing still in progress".

corverroos (changes requested):
This doesn't solve the problem

PR #396, opened 2024-02-26
refactor(lib/engine): move engine API to ethclient package

corverroos:
Moves the engine API to the ethclient package since it is part of the ethereum JSON RPC API.

task: none


halo/attest/keeper/keeper.go
```diff
@@ -51,7 +49,6 @@ type Keeper struct {
 func NewKeeper(
 	cdc codec.BinaryCodec,
 	storeSvc store.KVStoreService,
-	ethCl engine.API,
```
corverroos, line 54: not used

Lazar955 (approved):

PR #399, opened 2024-02-26
feat(metrics): add metrics exempalrs

pavelnikolov:
adds prometheus exemplars

task: none


corverroos (commented):
Examplars are cool, we should definitely use them. Not sure this example is the best use-case though.

pavelnikolov:
Yeah, this is a terrible example. This is why I need your help to fine an actual use case that would be useful.

PR #400, opened 2024-02-26
chore(relayer): monitor attested height from relayer

corverroos:
Monitor "latest attested height" from relayer. This ensures consistent/synchronised reporting of key cross-chain metrics and avoids prometheus scrap offset issues resulting in negative lag in grafana.

Added a `LatestAttestation` query endpoint to `halo's` `attest` module.

task: none

### DIFF
--- a/relayer/app/metrics.go
+++ b/relayer/app/metrics.go
@@ -75,4 +75,11 @@ var (
 		Name:      "head_height",
 		Help:      "The latest height of different types of head blocks on a specific chain",
 	}, []string{"chain", "type"})
+
+	attestedHeight = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "relayer",
+		Subsystem: "monitor",
+		Name:      "halo_attested_height",
+		Help:      "The latest halo attested height of a specific chain",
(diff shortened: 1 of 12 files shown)


pavelnikolov (approved):

relayer/app/metrics.go
```diff
@@ -75,4 +75,11 @@ var (
 		Name:      "head_height",
 		Help:      "The latest height of different types of head blocks on a specific chain",
 	}, []string{"chain", "type"})
+
+	attestedHeight = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "relayer",
+		Subsystem: "monitor",
+		Name:      "halo_attested_height",
+		Help:      "The latest halo attested height of a specific chain",
+	}, []string{"chain"})
```
pavelnikolov, line 84: Maybe this can be a good candidate for an exemplar.

Lazar955 (approved):

relayer/app/metrics.go
corverroos: mmm, yeah, let me think what we can add

PR #405, opened 2024-02-26
chore(docs): move unused docs

idea404:
Moves `archive/` docs outside of `./docs/site/docs` to `/docs/site` which will avoid them from being hosted in the deployed site.

This requires re-indexing the site after this PR is merged.

task: https://app.asana.com/0/1206684215872507/1206684219212557


ttarsi (approved):
🙌

what does re-indexing the site entail? can we automate that?

PR #409, opened 2024-02-26
feat(contracts): eject an operator

kevinhalliday:
Allow avs owner to eject an operator.

task: https://app.asana.com/0/1206208509925075/1206623482381790


contracts/src/protocol/OmniAVS.sol
```diff
@@ -341,6 +341,19 @@ contract OmniAVS is IOmniAVS, IOmniAVSAdmin, IServiceManager, OwnableUpgradeable
...
+    function ejectOperator(address operator) external onlyOwner {
+        require(_isOperator(operator), "OmniAVS: not an operator");
+
+        _removeOperator(operator);
+        _avsDirectory.deregisterOperatorFromAVS(operator);
+        _allowlist[operator] = false;
+
+        emit OperatorRemoved(operator);
+        emit OperatorDisallowed(operator);
+        emit OperatorEjected(operator);
```
kevinhalliday, line 354: Are all these events needed?

ttarsi: Hm, it seems we might just want to add this functionality to `removeFromAllowlist`. If we're removing them from the allow list, shouldn't we also deregister them from the directory, and remove them from our list?

ttarsi: I don't think we need all these events. One way to think about events is that it should reflect state changes, not "actions". Disallowed and Removed are 2 state changes, but Ejected is not

contracts/src/protocol/OmniAVS.sol
```diff
@@ -341,6 +341,19 @@ contract OmniAVS is IOmniAVS, IOmniAVSAdmin, IServiceManager, OwnableUpgradeable
         emit OperatorDisallowed(operator);
     }
 
+    /// @inheritdoc IOmniAVSAdmin
+    function ejectOperator(address operator) external onlyOwner {
+        require(_isOperator(operator), "OmniAVS: not an operator");
+
+        _removeOperator(operator);
+        _avsDirectory.deregisterOperatorFromAVS(operator);
+        _allowlist[operator] = false;
```
ttarsi, line 350: perhaps use `_removeFromAllowlist`

contracts/src/protocol/OmniAVS.sol
ttarsi: We can have our events in the internal funcs, so we don't have to add them here

kevinhalliday: having events mirror state changes I think is a rule that helps towards an aim, but is not the aim itself. the aim should be - can we look at the history of this contract and know whats' heppened

so another way to look at it - do we want to be able to look at the history and see who was ejected? 

tho since this is an admin function, not super important IMO

contracts/src/protocol/OmniAVS.sol
```diff
@@ -328,16 +333,41 @@ contract OmniAVS is IOmniAVS, IOmniAVSAdmin, IServiceManager, OwnableUpgradeable
...
+
+    /// @inheritdoc IOmniAVSAdmin
+    function ejectOperator(address operator) external onlyOwner {
+        require(_isOperator(operator), "OmniAVS: not an operator");
+
+        _removeFromAllowlist(operator);
+        _deregsiterOperatorFromAVS(operator);
+
+        emit OperatorEjected(operator);
+    }
```
kevinhalliday, line 353: It seems like `removeFromAllowList` and `ejectOperator` are basically the same - except

- `ejectOperator` can only be called when an operator is already an operator
- `removeFromAllowList` can be called before an operator is an operator

Should we just join these into one from function?

PR #412, opened 2024-02-27
fix(contracts): set feeFor as view in interaface

kevinhalliday:
Make feeForSync a view function in IOmniAVS.

The comment describing why it is not a view func is no longer true.

task: none


ttarsi (approved):
Enable auto-merge for [smol 'tings](https://x.com/inversebrah) like this

PR #413, opened 2024-02-27
feat(halo/attest): verify votes in window

corverroos:
When proposing and verifying and processing votes, ensure they are in the current vote window.

task: none


halo/attest/keeper/msg_server.go
```diff
@@ -86,3 +72,34 @@ func headersByAddress(aggregates []*types.AggVote, address common.Address) []*ty
...
+		slog.Int("votes", len(headers)),
+	}
+	for cid, hs := range heights {
+		attrs = append(attrs, slog.String(
+			strconv.FormatUint(cid, 10),
+			fmt.Sprint(hs),
+		))
+	}
+
+	log.Debug(ctx, "Marked local votes as "+typ, attrs...)
```
Lazar955, line 104: nit sprintf instead of concat strings

Lazar955 (approved):

halo/attest/keeper/msg_server.go
corverroos: will fix in next PR thanks

PR #421, opened 2024-02-27
fix(docs): update outdated code snippet

kevinhalliday:
Fix outdated code snippet.

task: none


kevinhalliday:
This is the issue with copying source code in docs 🤷‍♂️ we have to be on top of this - or just link to github

idea404:
@kevinhalliday  true a good feature for the future would be to directly reference the src

idea404 (approved):

PR #426, opened 2024-02-27
refactor(halo/voter): harden voter

corverroos:
Hardens the voter:
- Only start voting when local node is a validator (also vice versa)
- When starting, use latest approved height
- Restart if behind vote window (catch up if lagging behind)

TODO:
- Wire it up and start it with app context.
- Trim attestations from state that are behind vote window (they will never confirm).
- Add tests for all this.
- Rename a receiver to v.

task: none


halo/attest/voter/voter.go
```diff
@@ -63,38 +78,121 @@ func LoadVoter(ctx context.Context, privKey crypto.PrivKey, path string, provide
...
+	if err != nil {
+		return err
+	}
+
+	// Wait for error or context cancel.
+	select {
+	case err := <-errChan:
+		return err
+	case <-ctx.Done():
+		return ctx.Err()
```
idea404, line 190: Why always return an error?

halo/attest/voter/voter.go
```diff
@@ -63,38 +78,121 @@ func LoadVoter(ctx context.Context, privKey crypto.PrivKey, path string, provide
...
 		}
+
+		log.Warn(ctx, "Vote runner failed (will retry)", err)
+		backoff()
+	}
+}
+
+// runOnce blocks, streaming xblocks from the provided chain until an error is encountered.
+// It always returns a non-nil error.
+func (a *Voter) runOnce(ctx context.Context, app App, chainID uint64) error {
```
idea404, line 124: Why always return a non-nil error?

pavelnikolov (approved):

halo/attest/voter/voter.go
corverroos: because it returns only when:
 - a) it encounters an error (it returns that error)
 - b) the context is cancelled (it return ErrContextCancelled)

PR #429, opened 2024-02-27
feat(contracts): make avs pausable

kevinhalliday:
When paused, users cannot
  - register as an operator
  - deregister as an operator
  - syncWithOmni

Paused tests added. Initialize tests started.

task: https://app.asana.com/0/1206208509925075/1206650262528031


ttarsi (approved):
why do we need to pause in the initializer?

idea404 (approved):
also wondering why init of pause in constructor is required

kevinhalliday:
would be to let you deploy with contract initially paused. inspired by eigens initialization pattern. but discussed, and we don't need it

PR #432, opened 2024-02-27
chore(contracts): comment bindings type conflcts

kevinhalliday:
Cleanup bindings, add autocomment.

task: none


kevinhalliday:
Cleanup: 
- gen bindings for MockERC20, not ERC20
- remove EIP1271SignatureUtils bindings (it's a library)

Add autocomment:
- abigen generators go structs for each solidity struct it encounters in a file, this leads to conflicts
- on latest bindings update, we merged with manual comments
- bindings should not need manual comments, as we will run them in CI

pavelnikolov (approved):

contracts/bindings/scripts/commenttypes.go
```diff
@@ -0,0 +1,130 @@
...
+// commentTypes comments out type definitions in a Go file.
+func commentTypes(content string, types []string) string {
+	var modifiedLines []string
+	lines := strings.Split(content, "\n")
+
+	// tracks whether we're in a type definition
+	//
+	// type Name {
+	// ...
+	// }
```
pavelnikolov, line 77: Ideally, we need to do this using some form of AST.

kevinhalliday: yeah but this is pretty lowstakes, single usecase, so quick and dirty is fine

pavelnikolov: Take a look at [this library](https://github.com/unpackdev/solgo).
Alternatively [this SO answer](https://ethereum.stackexchange.com/questions/28523/how-to-go-from-solidity-ast-back-to-source-code) recommends [this library](https://github.com/chao-peng/SIF).

pavelnikolov: I didn't realize that this was Go code. Here is [an example](https://go.dev/src/go/ast/example_test.go).

PR #433, opened 2024-02-27
chore(contracts): refactor avs storage

kevinhalliday:
Refactor avs contract storage

- move it to it's own contract
- rename minimumOperatorStake to minOperatorStake

task: none


contracts/src/protocol/OmniAVS.sol
```diff
@@ -9,53 +9,25 @@ import { IStrategy } from "eigenlayer-contracts/src/contracts/interfaces/IStrate
 import { ISignatureUtils } from "eigenlayer-contracts/src/contracts/interfaces/ISignatureUtils.sol";
 import { IServiceManager } from "eigenlayer-middleware/src/interfaces/IServiceManager.sol";
 
+import { IDelegationManager } from "../interfaces/IDelegationManager.sol";
 import { OmniPredeploys } from "../libraries/OmniPredeploys.sol";
-import { IOmniPortal } from "../interfaces/IOmniPortal.sol";
+import { OmniAVSStorage } from "./OmniAVSStorage.sol";
```
ttarsi, line 14: Import this after all the interfaces, so it's clear there are definitions imported from this contract

contracts/src/protocol/OmniAVSStorage.sol
```diff
@@ -0,0 +1,51 @@
...
+
+import { IAVSDirectory } from "eigenlayer-contracts/src/contracts/interfaces/IAVSDirectory.sol";
+
+import { IDelegationManager } from "../interfaces/IDelegationManager.sol";
+import { IOmniPortal } from "../interfaces/IOmniPortal.sol";
+import { IOmniAVS } from "../interfaces/IOmniAVS.sol";
+
+abstract contract OmniAVSStorage {
+    /// @notice Constant used as a divisor in calculating weights
+    uint256 internal constant WEIGHTING_DIVISOR = 1e18;
```
ttarsi, line 12: since this isn't storage, should we put it in the inheritoor?

contracts/src/protocol/OmniAVSStorage.sol
```diff
@@ -0,0 +1,51 @@
...
+    /// @notice OmniPortal.xcall base gas limit in syncWithOmnj
+    uint64 public xcallBaseGasLimit = 75_000;
+
+    /// @notice Minimum stake required for an operator to register, not including delegations
+    uint96 public minOperatorStake;
+
+    /// @notice Omni portal contract, used to make xcalls to the Omni chain
+    IOmniPortal public omni;
+
+    constructor(IDelegationManager delegationManager_, IAVSDirectory avsDirectory_) {
```
ttarsi, line 47: this is logic so should still be in the other contract if it can be. This should just be storage variables. Then it's clear that it is an append only list of storage vars and we can add to the end of it, but can't change anything in the middle

ttarsi (approved):

contracts/src/protocol/OmniAVSStorage.sol
kevinhalliday: If we want immutables and constants in the storage contract we need this here.

So yeah guess question is the same yours above - do we want to list immutables and constants here, or in parent?

Like you said, they're not storage so yeah no real need for them to be in storage contract. Other than just organization. 

I do like having all of the "storage /  constants / immutables" in one place visually. But I won't fight for it. 

This is pattern in eigen contracts   so I think either way we're fine.

ttarsi: Cool. You make the call

kevinhalliday: I'm gonna go with immutables and constants in parent. So the storage can be only storage, and it's clear that anything there must be append only.

PR #436, opened 2024-02-28
feat(halo/config): support snapshot keep recent flag

corverroos:
Adds a `--snapshot-keep-recent` flag to halo to configure the number of snapshots to keep on disk.

Disable snapshot pruning in e2e tests to avoid [flapping](https://github.com/omni-network/omni/actions/runs/8073387873/job/22057172343) test if restoring a snapshot is slow (due to network blips).

task: none

### DIFF
--- a/halo/config/config.go
+++ b/halo/config/config.go
@@ -25,9 +25,10 @@ const (
 	networkFile     = "network.json"
 	attestStateFile = "xattestations_state.json"
 
-	DefaultHomeDir          = "./halo" // Defaults to "halo" in current directory
-	defaultSnapshotInterval = 1000     // Roughly once an hour (given 3s blocks)
-	defaultMinRetainBlocks  = 0        // Retain all blocks
+	DefaultHomeDir            = "./halo" // Defaults to "halo" in current directory
+	defaultSnapshotInterval   = 1000     // Roughly once an hour (given 3s blocks)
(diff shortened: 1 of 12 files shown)


halo/config/config.go
```diff
@@ -25,9 +25,10 @@ const (
 	networkFile     = "network.json"
 	attestStateFile = "xattestations_state.json"
 
-	DefaultHomeDir          = "./halo" // Defaults to "halo" in current directory
-	defaultSnapshotInterval = 1000     // Roughly once an hour (given 3s blocks)
-	defaultMinRetainBlocks  = 0        // Retain all blocks
+	DefaultHomeDir            = "./halo" // Defaults to "halo" in current directory
+	defaultSnapshotInterval   = 1000     // Roughly once an hour (given 3s blocks)
+	defaultSnapshotKeepRecent = 2
```
jmozah, line 30: Do you think having the same defaultSnapshotKeepRecent in the main net is safe? is there any. scenario in which having more snapshots might help. May be when there is a stall or something..

jmozah (approved):

halo/config/config.go
corverroos: this is the default of cosmos-SDK. I assume two is good enough. Basically just latest, and a backup of latest.

PR #437, opened 2024-02-28
fix(lib/txmgr): fix data race in test

Lazar955:
fixes data race by introducing rwlock. Tests ran with:
```go
go test -run TestTxMgrConfirmsMinGasPriceAfterBumping -v -race -count 100 -failfast
```

task: none


lib/txmgr/txmgr_internal_test.go
```diff
@@ -218,6 +218,9 @@ func (b *mockBackend) HeaderByNumber(ctx context.Context, number *big.Int) (*typ
 		num.Set(number)
 	}
 
+	b.g.mu.RLock()
```
corverroos, line 221: prefer to access shared resources through getters, instead of havings user wrangle locks themselves.

```
func (fooer) SetFoo(foo)
f.mu.lock
defer f.mu.unlock
f.foo = foo
}

func (fooer) GetFoo()
f.mu.lock
defer f.mu.unlock
return f.foo
}

```

corverroos (approved):

pavelnikolov (approved):

PR #438, opened 2024-02-28
feat(halo/cmd): register operator with AVS

jmozah:
Register or deregister operator with AVS

task: none


halo/cmd/flags.go
```diff
@@ -25,3 +25,10 @@ func bindInitFlags(flags *pflag.FlagSet, cfg *InitConfig) {
 	flags.BoolVar(&cfg.Force, "force", cfg.Force, "Force initialization (overwrite existing files)")
 	flags.BoolVar(&cfg.Clean, "clean", cfg.Clean, "Delete home directory before initialization")
 }
+
+func bindOperatorFlags(flags *pflag.FlagSet, cfg *OperatorConfig) {
+	libcmd.BindHomeFlag(flags, &cfg.HomeDir)
+	flags.StringVar(&cfg.L1ChainName, "l1chain", cfg.L1ChainName, "The layer 1 chain name")
+	flags.StringVar(&cfg.AVSDirectoryAddr, "avs-dir-addr", cfg.AVSDirectoryAddr, "The AVS Directory contract address")
```
kevinhalliday, line 32: We can read the AVS direcetory from the OmniAVAS contract, so operators don't need to find the AVS directory address

halo/cmd/flags.go
```diff
@@ -25,3 +25,10 @@ func bindInitFlags(flags *pflag.FlagSet, cfg *InitConfig) {
 	flags.BoolVar(&cfg.Force, "force", cfg.Force, "Force initialization (overwrite existing files)")
 	flags.BoolVar(&cfg.Clean, "clean", cfg.Clean, "Delete home directory before initialization")
 }
+
+func bindOperatorFlags(flags *pflag.FlagSet, cfg *OperatorConfig) {
+	libcmd.BindHomeFlag(flags, &cfg.HomeDir)
+	flags.StringVar(&cfg.L1ChainName, "l1chain", cfg.L1ChainName, "The layer 1 chain name")
```
kevinhalliday, line 31: Feels a little weird for "l1 chain name" to be part of this interface for operator registration. Should netconf chain names be something operators are familiar with when registering with our AVS?

Maybe we can add `IsL1` to `netconf.Chain` , like we have `IsOmni`?  Or we could have them just provide an rpc url here?

halo/cmd/register_to_avs.go
```diff
@@ -0,0 +1,166 @@
...
+
+	operatorSignatureWithSaltAndExpiry.Signature, err = crypto.Sign(digestHash[:32], operPK)
+	if err != nil {
+		return errors.Wrap(err, "error signing)")
+	}
+
+	if len(operatorSignatureWithSaltAndExpiry.Signature) != 65 {
+		return errors.New("invalid signature length")
+	}
+	operatorSignatureWithSaltAndExpiry.Signature[64] += 27
```
kevinhalliday, line 121: I think we have some utilities for validators to sign things in `lib/k1util` - maybe use them here?

kevinhalliday (approved):

halo/cmd/flags.go
jmozah: i will add giving the url

PR #443, opened 2024-02-28
chore(contracts/avs): refactor avs contract

kevinhalliday:
Refactor avs contract.

- use full natspec comments in avs source
- update comment content & format
- revise avs func grouping
- replace "Validator" term with "Operator"
- rename "StrategyParams" to "StragegyParam"

task: none


kevinhalliday:
Some further explanation.

- made some opinionated choices in comments & func grouping. Think this makes things more digestible for new comers, will read better for audits
- renamed Validator struct to Operator - I did not like how we had to things with the same meaning. This choice was left over from when we had name conflict with eigenlayer middleware `OperatorStateRetreiver`, which we no longer use

ttarsi (approved):

PR #446, opened 2024-02-29
chore(docs): add operator info

ttarsi:
- use "introduction" subsection for each section to standardize the intro page
- use xapp rather than both xapp and xdapp (not finalized, but for consistency)
- add some operator info
- use bold for token ids

task: none


idea404 (approved):
Nice, improves the structuring and readability of the site

PR #451, opened 2024-02-29
chore(relayer): improve relayer metrics

corverroos:
Improve relayer metrics:
- Query in correct order (from back to front)
- Instrument metrics atomically.
- Instrument `xfinal` head which is the `FinalisationStrategy` height.

Also add `slow_tests` flag to e2e manifest and only run slow AVS tests in `simple` manifest.

task: none

### DIFF
--- a/relayer/app/monitor.go
+++ b/relayer/app/monitor.go
@@ -48,49 +48,62 @@ func monitorHeightsForever(ctx context.Context, chain netconf.Chain,
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			// First get chain heads
-			monitorHeadsOnce(ctx, chain.Name, client)
+			// First get attested head (so it can't be lower than heads).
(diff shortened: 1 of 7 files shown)


relayer/app/monitor.go
```diff
@@ -48,49 +48,62 @@ func monitorHeightsForever(ctx context.Context, chain netconf.Chain,
...
-			err := monitorAttestedOnce(ctx, chain.Name, chain.ID, chain.DeployHeight, cprovider)
+			// then get chain heads (so it is always higher than attested).
+			heads := getHeads(ctx, client)
+
+			// Then populate gauges "at the same time" so they update "atomically".
 			if err != nil {
 				log.Error(ctx, "Monitoring attested failed (will retry)", err,
 					"chain", chain.Name)
+			} else {
+				attestedHeight.WithLabelValues(chain.Name).Set(float64(attested))
```
pavelnikolov, line 62: Should we also increase a counter every time an error occurs here?

Lazar955 (approved):

PR #453, opened 2024-02-29
refactor(lib/avs): refactor avs and backends

corverroos:
Refactor `avs` package:
- Make it a first class citizen `lib/avs`, since we need to use in production cli
- Move some code out of testing, ready for production use.
- Use `Backends`.
- Remove avs testing from e2e, since it only needs a single anvil at this point.

Refactor `backend` package:
- Make an individual `Backend` a first class citizen.
- Support multiple `accounts` per `Backend`.
- We should probably more Backends to lib as well.

task: none


relayer/app/monitor.go
```diff
@@ -23,8 +23,7 @@ func startMonitoring(ctx context.Context, network netconf.Network, xprovider xch
 	cprovider cchain.Provider, addr common.Address, rpcClients map[uint64]ethclient.Client) {
 	for _, srcChain := range network.Chains {
 		go monitorAccountForever(ctx, addr, srcChain.Name, rpcClients[srcChain.ID])
-		go monitorHeadsForever(ctx, srcChain.Name, rpcClients[srcChain.ID])
```
corverroos, line 26: can ignore this relayer/monitor changes

Lazar955 (approved):

PR #466, opened 2024-02-29
chore(contracts): remove allowlist from initializer

kevinhalliday:
Remove allowlist from initializer. 

task: none


kevinhalliday:
@ttarsi for your consideration. we don't need allowlist in initializer. and less code is good

ttarsi (approved):

kevinhalliday:
Lot of bindings updated - probably time do introduce ci flow for this

PR #470, opened 2024-03-01
feat(contracts): move avs eth stake inbox to storage

kevinhalliday:
Introduce ethStakeInbox AVS storage.

- rename OmniEthRestaking to EthStakeInbox
- move tracking the inbox address to avs storage

task: none


kevinhalliday:
Trying a out new name

contracts/src/libraries/OmniPredeploys.sol
```diff
@@ -7,5 +7,5 @@ pragma solidity ^0.8.12;
  */
 library OmniPredeploys {
     // TODO: replace with actual predeploy address
-    address public constant OMNI_ETH_RESTAKING = address(0);
+    address public constant ETH_STAKE_INBOX = address(12_345);
```
ttarsi, line 10: maybe move this to `test`, so that it's clear we're not inputting a dummy address into part of the protocol

ttarsi (approved):
why did this change so many bindings, like transparentupgradeableproxy?

kevinhalliday:
I don't know why bindings are changing so frequently - I'lll investigate.

kevinhalliday:
It's an abigen version issue. I'll look into fix. Could dockerize. Is this something @DanFlannel  is working on?

PR #472, opened 2024-03-01
fix(e2e): fix prometheus scrape configs in staging

corverroos:
Fixes `vmcompose` infra provider to deploy correct prometheus config that only tries to scrape services that are deployed on each specific host.

Previously, each prometheus tried to scrape all targets, even though they were not present on all hosts. This resulted in "relayer" down alerts.

Also add a "host" label to prometheus metrics so we can differentiate where the metric are coming from.

task: none

### DIFF
--- a/test/e2e/app/setup.go
+++ b/test/e2e/app/setup.go
@@ -20,6 +20,7 @@ import (
 	"github.com/omni-network/omni/lib/log"
 	"github.com/omni-network/omni/lib/netconf"
 	relayapp "github.com/omni-network/omni/relayer/app"
+	"github.com/omni-network/omni/test/e2e/app/agent"
 	"github.com/omni-network/omni/test/e2e/types"
 	"github.com/omni-network/omni/test/e2e/vmcompose"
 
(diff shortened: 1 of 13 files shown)


test/e2e/app/setup.go
```diff
@@ -49,17 +50,13 @@ const (
...
 // Setup sets up the testnet configuration.
-func Setup(ctx context.Context, def Definition, promSecrets PromSecrets, testCfg bool) error {
+func Setup(ctx context.Context, def Definition, agentSecrets agent.Secrets, testCfg bool) error {
 	log.Info(ctx, "Setup testnet", "dir", def.Testnet.Dir)
 
 	if err := os.MkdirAll(def.Testnet.Dir, os.ModePerm); err != nil {
 		return errors.Wrap(err, "mkdir")
 	}
 
-	if err := def.Infra.Setup(); err != nil {
```
corverroos, line 59: had to move this down below prometheus setup

Lazar955 (approved):

PR #475, opened 2024-03-01
refactor(halo/voter): refactor voter to use synchronous streamer

corverroos:
Refactor `voter` to use the new synchronous streamer. Which reduces some boilerplate.

task: none


halo/attest/voter/voter.go
```diff
@@ -144,53 +144,29 @@ func (a *Voter) runOnce(ctx context.Context, chainID uint64) error {
 		fromHeight = latest + 1
 	}
 
-	// Channel shenanigans to wait for async subscription.
```
corverroos, line 147: no more schenanigans

pavelnikolov (approved):

Lazar955 (approved):

halo/attest/voter/voter.go
ttarsi: lol

PR #477, opened 2024-03-01
feat(contracts): reduce xrecv gas usage

kevinhalliday:
Reduce XApp.xrecv gas usage by 95k.

- Trim data stored in current xmsg, to only what's needed
- And pinpong gas usage tests
- Update mock A,portal to enforce gas limits, and error when it runs out of gas
- Refactor xtype comments.

task: none


contracts/src/examples/PingPong.sol
```diff
@@ -47,6 +49,25 @@ contract PingPong is XApp {
         _xpingpong(xmsg.sourceChainId, xmsg.sender, times, n - 1);
     }
 
+    /**
+     * @notice The pingpong xmsg loop
+     * @dev Used to test differnce in gas usage between xrecv and non-xrecv functions
+     * @param times The pingpongs in the loop
+     * @param n The number of xcalls left to make
+     */
+    function pingpong_norecv(uint64 times, uint64 n) external {
```
idea404, line 58: Shouldn't it be called 'pingpong_recv` if it requires an xcall?

idea404 (commented):
What are the drivers behind calling it `MsgFull` now?

contracts/src/examples/PingPong.sol
kevinhalliday: no this is to mark this as not using the `recv` modifier

contracts/src/libraries/XTypes.sol
```diff
@@ -7,60 +7,89 @@ pragma solidity ^0.8.12;
...
+     *         merkle trees / proofs.
+     * @custom:field sourceChainId  Chain ID of the source chain
+     * @custom:field destChainId    Chain ID of the destination chain
+     * @custom:field streamOffset   Monotonically incremented offset of Msg in source -> dest Stream
+     * @custom:field sender         msg.sender of xcall on source chain
+     * @custom:field to             Target address to call on destination chain
+     * @custom:field data           Data to provide to call on destination chain
+     * @custom:field gasLimit       Gas limit to use for call execution on destination chain
+     */
+    struct MsgFull {
```
ttarsi, line 22: Since this is the "true" `XMsg`, and the one below is a trimmed version, seems we should call this `XMsg` and the trimmed one something like `MsgShort`

But we can still keep the `xmsg.sender` in `Xapp` by setting xmsg to a `MsgShort` type

ttarsi (approved):

contracts/src/libraries/XTypes.sol
kevinhalliday: Yeah I had XMsgTrim at first.

But wanted the user facing type to be XTypes.Msg 

There might be better ways to get that result tho.

PR #480, opened 2024-03-01
chore(contracts): update storage before external calls

ttarsi:
A slither recommendation. Just to be sanitary, as these are trusted calls.

task: none


kevinhalliday (approved):
Nice

PR #484, opened 2024-03-04
feat(relayer/app): stateful relayer

Lazar955:
This PR adds the ability to persist cursors on disk upon successful submission. The state is read from the disk upon startup.

task: https://app.asana.com/0/1206208509925075/1206727532156112/


relayer/app/config.go
```diff
@@ -12,6 +12,10 @@ import (
 	_ "embed"
 )
 
+const (
+	stateFile = "relayer-state.json"
```
corverroos, line 16: add this to config

relayer/app/state.go
```diff
@@ -0,0 +1,90 @@
...
+	"encoding/json"
+	"os"
+	"sync"
+
+	"github.com/omni-network/omni/lib/errors"
+
+	"github.com/cometbft/cometbft/libs/tempfile"
+)
+
+type State interface {
```
corverroos, line 13: no need to interface if you are not mocking it. Using real implementation is ok in tests. So interface not needed.

corverroos: Maybe rename to `Cursors`

relayer/app/app.go
```diff
@@ -46,6 +46,13 @@ func Run(ctx context.Context, cfg Config) error {
 	cprov := cprovider.NewABCIProvider(tmClient, network.ChainNamesByIDs())
 	xprov := xprovider.New(network, rpcClientPerChain)
 
+	state, ok, err := Load(stateFile)
```
corverroos, line 49: `LoadCursors`

relayer/app/app.go
```diff
@@ -61,7 +68,10 @@ func Run(ctx context.Context, cfg Config) error {
 			cprov,
 			xprov,
 			CreateSubmissions,
-			sendProvider)
+			sendProvider,
+			state)
```
corverroos, line 72: rather provide the cursors for this worker in the constructor. Setter should only be used if the data is not available at constructing time.

corverroos: I think you need to `CursorsForWorker(destChainID)` function

relayer/app/state.go
```diff
@@ -0,0 +1,90 @@
...
+	Persist(srcID, dstID, height uint64) error
+}
+
+type PersistentState struct {
+	mu       sync.Mutex
+	filePath string
+	cursors  map[uint64]map[uint64]uint64 // destChainID -> srcChainID -> height
+}
+
+func NewPersistentState(filePath string) *PersistentState {
```
corverroos, line 24: rename to `EmptyCursors`?

relayer/app/worker.go
```diff
@@ -72,13 +79,23 @@ func (w *Worker) runOnce(ctx context.Context) error {
 
 	buf := newActiveBuffer(w.destChain.Name, mempoolLimit, sender)
 
+	loadedHeights, ok := w.cursors[w.destChain.ID]
+
 	var logAttrs []any //nolint:prealloc // Not worth it
 	for srcChainID, fromHeight := range FromHeights(cursors, w.destChain, w.network.Chains) {
```
corverroos, line 85: best rename existing cursors to something else.

corverroos, line 85: pass in cursors into `fromHeights` since cursors affect the from heights.

relayer/app/state.go
```diff
@@ -0,0 +1,110 @@
...
+	"encoding/json"
+	"os"
+	"sync"
+
+	"github.com/omni-network/omni/lib/errors"
+
+	"github.com/cometbft/cometbft/libs/tempfile"
+)
+
+// State represents the state of the relayer. It keeps track of the last submitted height for each source chain on each destination chain.
```
corverroos, line 13: not "last submitted height" since we don't submit empty block. Rather "last successfully process block (whether submitted or skipped due to being empty).

relayer/app/state.go
```diff
@@ -0,0 +1,110 @@
...
+	if !ok {
+		return 0
+	}
+
+	height, ok := srcMap[srcID]
+	if !ok {
+		return 0
+	}
+
+	return height
```
corverroos, line 56: You are not locking srcMap while you are reading from it...

Rather simplify to:
```
mu.lock
defer mu.unlock

return s.cursors[dstID][srcID]
```

corverroos (approved):

PR #498, opened 2024-03-05
feat(monitor): add avs contract monitoring

kevinhalliday:
Introduce monitor service, add avs monitoring.

task: https://app.asana.com/0/1206208509925075/1206650262528026

### DIFF
--- a/monitor/avs/avs.go
+++ b/monitor/avs/avs.go
@@ -1,0 +1,58 @@
+package avs
+
+import (
+	"context"
+
+	"github.com/omni-network/omni/contracts/bindings"
+	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/ethclient"
+	"github.com/omni-network/omni/lib/log"
+	"github.com/omni-network/omni/lib/netconf"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// Monitor starts monitoring the AVS contract.
+func Monitor(ctx context.Context, network netconf.Network, address common.Address) error {
+	log.Info(ctx, "Starting AVS monitor")
+
+	l1Client, err := newL1Client(network)
+	if err != nil {
+		return err
+	}
+
+	avs, err := newAVS(l1Client, address)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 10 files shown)


monitor/avs/avs.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/omni-network/omni/lib/ethclient"
+	"github.com/omni-network/omni/lib/log"
+	"github.com/omni-network/omni/lib/netconf"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// Monitor starts monitoring the AVS contract.
+func Monitor(ctx context.Context, network netconf.Network, address common.Address) error {
+	log.Info(ctx, "Starting AVS monitor")
```
pavelnikolov, line 17: You already log about this in the `Run` method.

kevinhalliday: I log `Starting monitor service` there. avs monitoring would be one thing that service monitories. Though, it's the only right now

monitor/avs/monitor.go
```diff
@@ -0,0 +1,66 @@
...
+		operatorDelegations.WithLabelValues(addr).Set(float64(delegated))
+
+		total = total + delegated + staked
+	}
+
+	numOperators.Set(float64(len(operators)))
+	totalDelegations.Set(float64(total))
+
+	return nil
+}
```
kevinhalliday, line 61: Some things I'm not sure on 

- When an operator is remove - their label will no longer be present in the metrics queries - how do we represent that correctly in grafana? i.e. we want a dashboard of currently registered operators w/ their stakes & delegations
- Related to above - the `total` doesn't feel like it needs to be here, it could be sum of the operator metrics, but I'm not sure how to handle operator removals over time

pavelnikolov (approved):

monitor/avs/monitor.go
pavelnikolov: When an operator is removed the time series data will not contain a point for this operator. We'll write unit tests for that tomorrow.

Lazar955 (approved):
nice work

PR #502, opened 2024-03-05
refactor(lib/ethbackend): move backends out of test

corverroos:
Move backends from `test/e2e/backend` to `lib/ethcient/ethbackend`. 

task: none


Lazar955 (approved):
👍

PR #509, opened 2024-03-05
fix(scripts): cli install script for linux

idea404:
This PR makes the install script compatible for linux and adds the export of `$HOME/bin` to the PATH in the shell run commands file

task: none

### DIFF
--- a/scripts/install_omni_cli.sh
+++ b/scripts/install_omni_cli.sh
@@ -4,31 +4,48 @@ set -e
 
 OS=$(uname -s)
 ARCH=$(uname -m)
+
+case $ARCH in
+    arm64) ARCH="arm64" ;;
+    aarch64) ARCH="arm64" ;;
+    x86_64) ARCH="x86_64" ;;
+    *) echo "Unsupported architecture"; exit 1 ;;
+esac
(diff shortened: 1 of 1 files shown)


corverroos (approved):

scripts/install_omni_cli.sh
```diff
@@ -4,18 +4,49 @@ set -e
...
+    echo "ℹ️ Adding '\$HOME/bin' to your PATH in $SHELL_PROFILE"
+    echo 'export PATH="$PATH:$HOME/bin"' >> "$SHELL_PROFILE"
+    export PATH=\$PATH:$HOME/bin
+fi
+
+# Check if omni is correctly installed
+if command -v omni &> /dev/null; then
+    echo "✅ omni is now installed. Run first 'source ~/$FILE' and then 'omni --help' to get started."
+else
+    echo "Error: omni executable not found in PATH. You may need to add '\$HOME/bin' to your PATH manually."
```
corverroos, line 51: can use ❌ for nice visible errors

idea404: big yes

PR #518, opened 2024-03-05
test(halo/attest): refactor keeper unit tests

pavelnikolov:
This PR:
* moves the type builders in to separate test files
* adds reusable func for `new(msgBuilder).Default()`, which is commonly used
* tests the keeper from outside - tests are in a package `keeper_test` instead of `keeper` (i.e. ZK tests )
* adds getters for the attestations and signature tables

task: https://app.asana.com/0/0/1206761482432698/f


halo/attest/keeper/keeper.go
```diff
@@ -98,6 +98,16 @@ func (k *Keeper) RegisterProposalService(server grpc1.Server) {
...
+// AttestTable returns the attestations ORM table.
+func (k *Keeper) AttestTable() AttestationTable {
+	return k.attTable
+}
+
+// SignatureTable returns the attestations ORM table.
+func (k *Keeper) SignatureTable() SignatureTable {
+	return k.sigTable
+}
+
```
corverroos, line 110: These getters should be for testing only. So move then to a `keeper_internal_test.go` file.  That means the external tests can access them, but they are available in production or to other packages.

halo/attest/keeper/keeper_test.go
```diff
@@ -0,0 +1,180 @@
...
+					{Id: 2, AttId: 1, Signature: val2.Bytes(), ValidatorAddress: val2.Address},
+				},
+			},
+		},
+		{
+			name: "two_votes_diff_blocks",
+			args: args{
+				msg: defaultMsg().
+					WithAppendVotes(
+						new(AggVoteBuilder).Default().WithBlockHeader(1, 501, blockHashes[1]).WithSignatures(sigsTuples(val1, val3)...).Vote(),
```
corverroos, line 60: cant we have a `defaultAggVote()` as well? I see a lot of `new(AggVoteBuilder).Default()` duplication

halo/attest/keeper/keeper.go
pavelnikolov: > but they are available in production or to other packages.
But anyone can copy the 5 lines from `keeper.New` and inspect the attestation table. Therefore we can't really achieve better security by removing the getters, right? 
Would you mind if the tests are in `keeper_test` package? I can still remove the getter and setter and create a separate helper func which creates these two ORM tables given a KV store, then I'll reuse this func between the keeper and the tests.

corverroos (approved):

PR #522, opened 2024-03-06
feat(contracts): add minOpStake and maxOpCount

kevinhalliday:
Add minOperatorStake and maxOperatorCount.

task: none


kevinhalliday:
I think we'll want this for goerli - just in case.
To avoid our avs being full of random, no delegations, operators, and breaking eigenlayer ui.
I also think these params will be useful in mainnet - so we can turn off allowlist without upgrading the contract. We'll need some caps on operators, because we can't use infinite gas in `syncWithOmni`. And requiring they have at least some stake before registering is useful.

contracts/script/avs/DeployGoerliAVS.s.sol
```diff
@@ -18,27 +18,36 @@ import { Script } from "forge-std/Script.sol";
...
-        uint256 deployerKey = vm.envUint("DEPLOYER_KEY");
-        address factory = vm.envOr("CREATE3_FACTORY_ADDRESS", _create3Factory);
-        address proxyAdmin = vm.envOr("PROXY_ADMIN", _proxyAdmin);
-        address owner = vm.envOr("AVS_OWNER", _owner);
-        address portal = vm.envOr("PORTAL_ADDRESS", _portal);
-        address ethStakeInbox = vm.envOr("ETH_STAKE_INBOX", _ethStakeInbox);
-        bytes32 salt = vm.envOr("CREATE3_SALT", _create3Salt);
+        require(_create3Factory != address(0), "create3Factory not set");
+        require(_proxyAdmin != address(0), "proxyAdmin not set");
+        require(_owner != address(0), "owner not set");
```
kevinhalliday, line 35: Just some cleanup here - have this in another PR too but just front running merge conflicts.

lib/avs/avs.go
```diff
@@ -85,6 +85,10 @@ func verifyRegisterOperator(ctx context.Context, contracts Contracts, operator c
 		return errors.New("operator not registered with eigen layer delegation manager")
 	}
 
+	// TODO: check if avs at maxOperatorCount
+	// TODO: check if operator has minimum stake
+	// TODO: handle allowlist disabled
```
kevinhalliday, line 90: I will handle this - the `handle allowlist disabled` is a bug in current cli for goerli right now. Because our allowlist will be disabled, and operators will fail the `isInAllowlist` check below

pavelnikolov (approved):

lib/avs/avs.go
kevinhalliday: Addressed https://github.com/omni-network/omni/pull/523

ttarsi (approved):

PR #549, opened 2024-03-07
feat(halo/attest): support inconsistent attestation root

corverroos:
Support votes for the same block but with different attestation roots. Previous implementation would stall if byzantine node voted for invalid attestation root first as honest nodes wouldn't be able to insert honest votes.

Also align naming to `AttestationRoot // Attestation merkle root of cross-chain block"

task: none


halo/attest/keeper/keeper_test.go
```diff
@@ -125,7 +125,7 @@ func TestKeeper_Add(t *testing.T) {
 				msg: defaultMsg().
 					WithVotes(
 						defaultAggVote().
-							WithBlockRoot([]byte("different root")). // the block root is intentionally different to cause an error
+							WithAttestationRoot([]byte("different root")). // the block root is intentionally different to cause an error
```
pavelnikolov, line 128: Change/remove the comment here since this doesn't cause an error anymore.

halo/attest/keeper/common_test.go
```diff
@@ -22,7 +22,9 @@ type mocks struct {
 	namer   *testutil.MockChainNamer
 }
 
-func mockDefaultExpectations(_ sdk.Context, m mocks) {}
+func mockDefaultExpectations(_ sdk.Context, m mocks) {
+	m.namer.EXPECT().ChainName(gomock.Any()).Return("test_chain").AnyTimes()
```
pavelnikolov, line 26: since by default we use the default message, maybe we want to make sure that we use the correct chain, e.g:
```suggestion
	m.namer.EXPECT().ChainName(uint64(1)).Return("test_chain").AnyTimes()
```

pavelnikolov: In my upcomming PR I have this:

```go
func TestKeeper_Approve(t *testing.T) {
	t.Skip()
	t.Parallel()
	...
	defaultExpectations := func(_ sdk.Context, m mocks) {
		m.voter.EXPECT().TrimBehind(gomock.Any()).Times(1).Return(0)
	}

	namerCalled := func(times int) expectation {
		return func(_ sdk.Context, m mocks) {
			m.namer.EXPECT().ChainName(uint64(1)).Times(times).Return("test-chain")
		}
	}

```

and then in the table tests:

```go
	expectations: []expectation{
		namerCalled(1),
		defaultExpectations,
	},
```

Note that in `common_test.go` I've defined types for:
```go

type expectation func(sdk.Context, mocks)
type prerequisite func(t *testing.T, k *keeper.Keeper, ctx sdk.Context)

```

I guess we are going to use `namerCalled` often.

halo/attest/keeper/keeper_test.go
```diff
@@ -138,14 +138,17 @@ func TestKeeper_Add(t *testing.T) {
 					require.NoError(t, err)
 				},
 			},
-			wantErr: true,
+			wantErr: false,
```
pavelnikolov, line 141: nit: `false` is the default value so just remove this field from the subtest

pavelnikolov (approved):

PR #554, opened 2024-03-07
feat(monitor/avs): add more metrics

kevinhalliday:
Add more avs metrics.

- owner
- paused
- allowlistEnabled
- strategyParams

task: none

### DIFF
--- a/monitor/avs/avs.go
+++ b/monitor/avs/avs.go
@@ -39,7 +39,7 @@ func Monitor(ctx context.Context, network netconf.Network) error {
 		return err
 	}
 
-	go monitorAVSOperatorsForever(ctx, avs)
+	startMonitoring(ctx, avs)
 
 	return nil
 }
--- a/monitor/avs/monitor.go
+++ b/monitor/avs/monitor.go
@@ -24,8 +24,8 @@ func startMonitoring(ctx context.Context, avs *bindings.OmniAVS) {
 	go monitorForever(ctx, avs, "strategyParams", monitorStrategyParamsOnce)
 }
 
-func monitorOperatorsOnce(avs *bindings.OmniAVS) error {
-	operators, err := avs.Operators(&bind.CallOpts{})
+func monitorOperatorsOnce(ctx context.Context, avs *bindings.OmniAVS) error {
+	operators, err := avs.Operators(&bind.CallOpts{Context: ctx})
 	if err != nil {
 		return errors.Wrap(err, "get operators")
 	}
@@ -52,8 +52,8 @@ func monitorOperatorsOnce(avs *bindings.OmniAVS) error {
 	return nil
 }
 
-func monitorOwnerOnce(avs *bindings.OmniAVS) error {
-	owner, err := avs.Owner(&bind.CallOpts{})
+func monitorOwnerOnce(ctx context.Context, avs *bindings.OmniAVS) error {
+	owner, err := avs.Owner(&bind.CallOpts{Context: ctx})
 	if err != nil {
 		return errors.Wrap(err, "get owner")
 	}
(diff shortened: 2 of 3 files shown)


monitor/avs/avs.go
```diff
@@ -39,7 +39,7 @@ func Monitor(ctx context.Context, network netconf.Network) error {
 		return err
 	}
 
-	go monitorAVSOperatorsForever(ctx, avs)
+	startMonitoring(ctx, avs)
```
pavelnikolov, line 42: I prefer `go startMonitoring(ctx, avs)` - no goroutines inside the method. This makes unit testing much easier.

kevinhalliday: how would that work? each of the `monitorForever` needs to be a goroutine
Or we could just get rid of `startMonitoring` and move them all here

pavelnikolov: I just saw that you do the same thing but for different bindings in parallel. 👍

monitor/avs/monitor.go
```diff
@@ -13,58 +14,146 @@ import (
...
+			err := f(avs)
+			if ctx.Err() != nil {
+				return
+			} else if err != nil {
+				log.Error(ctx, fmt.Sprintf("Monitoring AVS %s failed (will retry)", name), err)
+				continue
+			}
+		}
+	}
+}
```
pavelnikolov, line 155:
```suggestion
type monitor func(ctx context.Context, avs *bindings.OmniAVS)

// monitorForever runs the given monitor function every 30 seconds until the context is canceled.
func monitorForever(ctx context.Context, avs *bindings.OmniAVS, name string, fns ...monitor) {
	ticker := time.NewTicker(time.Second * 30)
	defer ticker.Stop()

	for {
		select {
		case <-ctx.Done():
			return
		case <-ticker.C:
			for _, f := range fns {
				go f(ctx, avs)
			}
		}
	}
}
```

monitor/avs/avs.go
pavelnikolov: See suggestion below

monitor/avs/monitor.go
kevinhalliday: So move error logging to each individual monitor? Also makes it harder to use different intervals for different monitors in future, but that's okay 30 seconds is fine for everything is what we do now

kevinhalliday: what problem were you trying to solve from current approach? aside from just refactoring

kevinhalliday: And would that `defer cancel()` on line 149 just cancel everything after the for loop, so right away?

monitor/avs/monitor.go
```diff
@@ -13,58 +14,150 @@ import (
...
+				ctx, cancel := context.WithTimeout(ctx, interval)
+				defer cancel()
+
+				return f(ctx, avs)
+			}()
+
+			if err != nil {
+				log.Error(ctx, fmt.Sprintf("Monitoring AVS %s failed (will retry)", name), err)
+			}
+		}
```
kevinhalliday, line 157: @pavelnikolov didn't change `monitorForever` to take a list of `fns`, so no need for wait group.
but I believe this gets us what we want - cancel the monitor func  if it doesn't finish before `interval` ends

pavelnikolov: This way you have one infinite loop, select block and time ticker per monitoring func instead of only one for the entire app. I'd prefer the waiting group, but not a show stopper really.

pavelnikolov (approved):

monitor/avs/monitor.go
kevinhalliday: Yes it would. Is single loop & ticker that big of a deal tho? This was pattern I followed from relayer monitoring

PR #562, opened 2024-03-09
chore(e2e): migrate to sepolia

ttarsi:
Goerli stopped finalizing blocks, RIP.
- use arb sepolia and op sepolia for rollups
- we can't use Goerli for the AVS contracts anymore (we could, but wouldn't be able to pass restaking payloads to Omni), so let's just use our anvil chain for now

task: none

### DIFF
--- a/test/e2e/manifests/staging.toml
+++ b/test/e2e/manifests/staging.toml
@@ -1,6 +1,7 @@
 network = "staging"
 public_chains = ["arb_sepolia", "op_sepolia"]
 anvil_chains = ["chain_a"]
+avs_target = "chain_a"
 multi_omni_evms = true
 prometheus = true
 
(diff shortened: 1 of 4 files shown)


test/e2e/manifests/staging.toml
```diff
@@ -1,7 +1,7 @@
 network = "staging"
-public_chains = ["goerli"]
+public_chains = ["arb_sepolia", "op_sepolia"]
 anvil_chains = ["chain_a"]
-avs_target = "goerli"
+avs_target = "chain_a"
```
kevinhalliday, line 4: This will break avs monitoring - let's just hard code it

pavelnikolov (approved):

Lazar955 (approved):

PR #565, opened 2024-03-11
chore(monitor): hardcode goerli staging avs monitor

kevinhalliday:
Hardcode staging goerli avs monitor. 

task: none


kevinhalliday:
IMO we don't need to merge this to fix our staging right now. We can wait until we handle monitoring separately.
But if we want to move staging to sepolia right now, let's merge

PR #576, opened 2024-03-12
feat(e2e): add avs-deploy subcommand

pavelnikolov:
Add `avs-deploy` subcommand to the e2e runner binary.

task: https://app.asana.com/0/1206208509925075/1206799220447925/f


kevinhalliday (approved):
This is a good start. Things we need to change

- allow configuration of proxy admin (right now this deploys a proxy admin)
- allow configuration of owner (right now owner is set to deployer)
- allow post deployment configuration (setting metadata uri) 

It will also set portal address to 0 if a portal is not yet deployed to that chain. Maybe let's log a warning?

PR #579, opened 2024-03-12
feat(halo): add evm genesis with staking predeploy

kevinhalliday:
Add dev evm genesis with OmniStake contract predeployed.

task: none


contracts/bindings/scripts/gen.sh
```diff
@@ -1,21 +1,26 @@
...
 # params:
-#  $1: contract name (ex. 'OmniPortal')
+#  $1: contract name (ex. 'OmniPortal', 'src/protocol/OmniPortal.sol:OmniPortal')
 gen_binding() {
-  name=$1
+  contract=$1
+
+  # strip path prefix, if used
+  # ex src/protocol/OmniPortal.sol:OmniPortal => OmniPortal
+  name=$(echo ${contract} | cut -d ":" -f 2)
```
kevinhalliday, line 16: Ran into issues with `forge inspect MockERC20 abi` returning the wrong abit - so needed to use the full path

corverroos (approved):

PR #581, opened 2024-03-13
feat(halo/evmstaking): add emvstaking log provider

corverroos:
Adds a `evmstaking.LogProvider` that monitors the omnistaking predeploy contract log events and creates validators accordingly. 

task: none


halo/evmstaking/evmstaking.go
```diff
@@ -0,0 +1,177 @@
...
+
+	_ = l.getOrCreateAccount(ctx, accAddr)
+
+	if err := l.bKeeper.MintCoins(ctx, AccountName, amountCoins); err != nil {
+		return errors.Wrap(err, "mint coins")
+	}
+
+	if err := l.bKeeper.SendCoinsFromModuleToAccount(ctx, AccountName, accAddr, amountCoins); err != nil {
+		return errors.Wrap(err, "send coins")
+	}
```
pavelnikolov, line 139: What happens if the coins are minted but the `SendCoinsFromModuleToAccount` fails for some reason?

halo/evmstaking/evmstaking.go
```diff
@@ -0,0 +1,177 @@
...
+		stypes.NewCommissionRates(math.LegacyZeroDec(), math.LegacyZeroDec(), math.LegacyZeroDec()),
+		sdk.DefaultPowerReduction)
+	if err != nil {
+		return errors.Wrap(err, "create validator message")
+	}
+
+	_, err = skeeper.NewMsgServerImpl(l.sKeeper).CreateValidator(ctx, msg)
+	if err != nil {
+		return errors.Wrap(err, "create validator")
+	}
```
pavelnikolov, line 155: Similarly to above, what happens if the above two succeed but this one fails?

test/e2e/app/run.go
```diff
@@ -121,6 +121,8 @@ func E2ETest(ctx context.Context, def Definition, cfg E2ETestConfig, secrets age
 
 	stopReceiptMonitor := StartMonitoringReceipts(ctx, def)
 
+	stopValidatorUpdates := StartValidatorUpdates(ctx, def)
```
pavelnikolov, line 124: Should we run `stopValidatorUpdates` using defer? Otherwise, if any of the following step errors and we return an error this will not be executed.

halo/evmstaking/evmstaking.go
```diff
@@ -0,0 +1,177 @@
...
+
+	if _, err := l.sKeeper.GetValidator(ctx, valAddr); err == nil {
+		return errors.New("validator already exists")
+	} else if _, err := l.sKeeper.GetValidatorByConsAddr(ctx, consAddr); err == nil {
+		return errors.New("validator pubkey already exists")
+	}
+
+	amountCoin, amountCoins := omniToBondCoin(deposit.Amount)
+
+	_ = l.getOrCreateAccount(ctx, accAddr)
```
pavelnikolov, line 131: So if I understand correctly, if someone stakes with Omni we create an an account and a validator. What happens if they withdraw their stake? Do we have to call [RemoveAccount](https://pkg.go.dev/github.com/cosmos/cosmos-sdk/x/auth/keeper#AccountKeeper.RemoveAccount)?

pavelnikolov (approved):

Lazar955 (approved):

halo/evmstaking/evmstaking.go
corverroos: no, I think we can keep the account forever

halo/evmstaking/evmstaking.go
corverroos: At this point, nothing should fail, but if it does, all the state updates should be reverted. It is a like any blockchain transaction that is included onchain but that fails. The state updates just doesn't happen

test/e2e/app/run.go
corverroos: this is e2e test, it doesn't matter. as soon as anything fails, the test exists

halo/evmstaking/evmstaking.go
pavelnikolov: Do we only set the validator power to 0 or something like that?

kevinhalliday (approved):

halo/evmstaking/evmstaking.go
```diff
@@ -0,0 +1,177 @@
...
+// - Create a new validator with the depositor's account.
+func (l LogProvider) DeliverLog(ctx context.Context, _ common.Hash, elog *evmenginetypes.EVMLog) error {
+	deposit, err := l.contract.ParseDeposit(elog.ToEthLog())
+	if err != nil {
+		return errors.Wrap(err, "parse deposit")
+	}
+
+	log.Info(ctx, "EVM staking deposit detected, adding new validator",
+		"depositor", deposit.Depositor.Hex(),
+		"amount", deposit.Amount.String())
```
kevinhalliday, line 110: Flagging - this could be an existing validator adding stake

kevinhalliday: Also - remove `Depositor` and replaced it with `Pubkey`.  We can add it back. Tho cchain should not need it. Could be useful for offchain tools to index val deposits by events.

halo/evmstaking/evmstaking.go
corverroos: yup

halo/evmstaking/evmstaking.go
corverroos: yup, still need to add logic for "adding more stake" as opposed to "adding initial stake".

PR #585, opened 2024-03-13
feat(contracts/avs): track operator public keys

kevinhalliday:
Track operator public keys.

- require public keys in registration
- add public keys to Operator struct sent to omni

task: https://app.asana.com/0/1206208509925075/1206799220447931


contracts/src/interfaces/IOmniAVS.sol
```diff
@@ -24,11 +25,13 @@ interface IOmniAVS {
...
+     * @custom:field pubkey     The operator's 64 byte uncompressed secp256k1 public key
      * @custom:field delegated  The total amount delegated, not including operator stake
      * @custom:field staked     The total amount staked by the operator, not including delegations
      */
     struct Operator {
         address addr;
+        bytes pubkey;
         uint96 delegated;
         uint96 staked;
     }
```
kevinhalliday, line 37: We can remove `addr` in a separate PR if we decide we don't want it. It's not strictly needed.

contracts/src/protocol/OmniAVSStorage.sol
```diff
@@ -11,6 +11,9 @@ abstract contract OmniAVSStorage {
     /// @notice Ethereum addresses of currently register operators
     address[] internal _operators;
 
+    /// @notice Map operator address to secp256k1 public key
+    mapping(address => bytes) internal _operatorPubkeys;
+
```
kevinhalliday, line 16: Added this here. I don't think we should upgrade our goerli avs. We should either just deploy a new one, and have operators reregister with public keys. Or wait until eigenlayer moves to holesky.

ttarsi (approved):

pavelnikolov (approved):

PR #587, opened 2024-03-14
chore(lib/xchain): add debug to get block

DanFlannel:
We are seeing errors for get block and we aren't seeing any useful logs 

task: none

### DIFF
--- a/lib/xchain/provider/fetch.go
+++ b/lib/xchain/provider/fetch.go
@@ -8,6 +8,7 @@ import (
 	"github.com/omni-network/omni/contracts/bindings"
 	"github.com/omni-network/omni/lib/errors"
 	"github.com/omni-network/omni/lib/ethclient"
+	"github.com/omni-network/omni/lib/log"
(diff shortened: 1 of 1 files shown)


Lazar955 (commented):
Maybe a debug log?

Lazar955 (approved):

lib/xchain/provider/fetch.go
```diff
@@ -106,6 +107,7 @@ func (p *Provider) GetBlock(ctx context.Context, chainID uint64, height uint64)
 
 	// ignore if our height is greater than the finalized height
 	if height > finalisedHeader.Number.Uint64() {
+		log.Warn(ctx, "Trying to fetch block height > finalized height", nil, "chain", chainID, "attested height", height, "finalized height", finalisedHeader.Number.Uint64())
```
ttarsi, line 110: this provider doesn't know anything about attestations so we should avoid that in the log, just remove "attested"

PR #588, opened 2024-03-15
feat(lib/fireblocks): initial fireblocks client

DanFlannel:
fireblocks api client

task: https://app.asana.com/0/1206208509925075/1206815715084836


lib/fireblocks/fireblocks.go
```diff
@@ -0,0 +1,16 @@
+package fireblocks
+
+import (
+	"context"
+
+	f "github.com/omni-network/omni/lib/fireblocks/api"
```
corverroos, line 6: as mentioned in the past, avoid these type of aliased imports, this isn't our style.

lib/fireblocks/fireblocks.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+type FireBlocks interface {
+	// CreateTransaction creates a new transaction on the FireBlocks API.
+	// We use raw signing by default
+	CreateTransaction(ctx context.Context, request f.CreateTransactionRequest) (string, error)
```
corverroos, line 12: what is the returned type? maybe make that a proper type. Also please mention in the doc what the method returns as this isn't obvious.

lib/fireblocks/fireblocks.go
```diff
@@ -0,0 +1,16 @@
...
+
+	// GetTransactionById retrieves a transaction by its ID.
+	GetTransactionById(ctx context.Context, transactionID string) (*f.CreateTransactionRequest, error)
```
corverroos, line 15: `GetTransactionById` should return a "transaction", not a "create request".

lib/fireblocks/api/get_transaction_by_id.go
```diff
@@ -0,0 +1,36 @@
+package fireblocks
```
corverroos, line 1: please don't use suck long file names, that is not our pattern

lib/fireblocks/api/client.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+type Client struct {
+	apiKey       string
+	clientSecret string
+	host         string
+	http         http.Client
+}
+
+const TransactionEndpoint string = "v1/transactions"
```
corverroos, line 25: don't export things by default

lib/fireblocks/api/client.go
```diff
@@ -0,0 +1,123 @@
...
+
+func NewClient(apiKey string, clientSecret string, host string) *Client {
```
corverroos, line 27: not sure host should be provided, isn't that constant? Where does the value come from?

lib/fireblocks/api/client.go
```diff
@@ -0,0 +1,123 @@
...
+	return &Client{
+		apiKey:       apiKey,       // pragma: allowlist secret
+		clientSecret: clientSecret, // pragma: allowlist secret
+		host:         host,
+	}
+}
+
+func (c *Client) sendRequest(ctx context.Context, req *http.Request, endpoint string, request []byte, nonce string) ([]byte, error) {
```
corverroos, line 35: only use pointer receivers for mutable state, all the Client fields are immutable, so no need for pointer receivers.

lib/fireblocks/api/client.go
```diff
@@ -0,0 +1,123 @@
...
+	apiKey       string
+	clientSecret string
+	host         string
+	http         http.Client
+}
+
+const TransactionEndpoint string = "v1/transactions"
+
+func NewClient(apiKey string, clientSecret string, host string) *Client {
+	return &Client{
```
corverroos, line 28: don't use pointers, see below

lib/fireblocks/api/client.go
```diff
@@ -0,0 +1,123 @@
...
+	"crypto/sha256"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/log"
+
+	"github.com/golang-jwt/jwt"
```
corverroos, line 15: we already have `"github.com/golang-jwt/jwt/v5"` as an import, please use that.

lib/fireblocks/api/client.go
corverroos: note we spoken about this in the past

lib/fireblocks/api/client.go
```diff
@@ -0,0 +1,123 @@
...
+
+func genJWTToken(uri string, nonce string, apiKey string, secretKey string, body string) (string, error) {
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256,
+		jwt.MapClaims{
+			"uri":      uri,
+			"nonce":    nonce,
+			"iat":      time.Now().Unix(),
+			"sub":      apiKey,
+			"bodyHash": body,
+			"exp":      time.Now().Add(time.Hour * 24).Unix(),
```
corverroos, line 114: 24 hours is a bit much? how did you decide this?

lib/fireblocks/api/client.go
```diff
@@ -0,0 +1,123 @@
...
+	bodyHash := sha256.Sum256(request)
+	authToken, err := genJWTToken(endpoint, nonce, c.apiKey, string(bodyHash[:]), c.clientSecret)
+	if err != nil {
+		return nil, errors.Wrap(err, "genJWTToken")
+	}
+	req.Header = getHeader(authToken, c.apiKey)
+
+	resp, err := c.http.Do(req)
+	if err != nil {
+		return nil, errors.New("http Do")
```
corverroos, line 45: errors.Wrap

lib/fireblocks/api/client.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+	}()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, errors.Wrap(err, "read response body")
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return body, errors.Wrap(err, "http response code: %d", resp.StatusCode)
```
corverroos, line 59: also return the body so we have more details about why it failed: `"body", string(body)`.

Also please learn how our errors library works. Never do "%d" or "%v".

lib/fireblocks/api/client.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	return body, nil
+}
+
+func (c *Client) createPostRequest(ctx context.Context, endpoint string, request []byte) (*http.Request, error) {
+	endpoint = fmt.Sprintf("%s/%s", c.host, endpoint)
```
corverroos, line 66: please use `filepath.Join` instead of `fmt.Sprint`

lib/fireblocks/api/create_transaction.go
```diff
@@ -0,0 +1,128 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"strconv"
+
+	"github.com/omni-network/omni/lib/errors"
+)
+
+type CreateTransactionRequest struct {
```
corverroos, line 11: please omit the fields we will not use.

lib/fireblocks/api/create_transaction.go
```diff
@@ -0,0 +1,128 @@
...
+	CustomerRefID      string         `json:"customerRefId"`
+	ExtraParameters    RawMessageData `json:"extraParameters"`
+}
+
+type RawMessageData struct {
+	Messages  []UnsignedRawMessage `json:"messages"`
+	Algorithm string               `json:"algorithm"`
+}
+
+type UnsignedRawMessage struct {
```
corverroos, line 40: please remove bitcoin stuff

lib/fireblocks/api/create_transaction.go
```diff
@@ -0,0 +1,128 @@
...
+	err = json.Unmarshal(response, &res)
+	if err != nil {
+		return nil, errors.Wrap(err, "unmarshal response JSON")
+	}
+
+	return &res, nil
+}
+
+type TransactionRequestOptions struct {
+	Nonce    string
```
corverroos, line 95: don't let the use of this library worry about nonce generate that on the fly.

lib/fireblocks/api/create_transaction.go
```diff
@@ -0,0 +1,128 @@
...
+	Amount   int64
```
corverroos, line 96: what is the unit of amount? if wei, this will overflow int64 easily.

lib/fireblocks/api/create_transaction.go
```diff
@@ -0,0 +1,128 @@
...
+	Messages []UnsignedRawMessage
```
corverroos, line 97: lets just provide a simple "single raw message signing API", no need for array here.

lib/fireblocks/api/client.go
```diff
@@ -0,0 +1,123 @@
...
+		bytes.NewReader(request),
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "new post request")
+	}
+
+	return req, nil
+}
+
+func (c *Client) createGetRequest(ctx context.Context, endpoint string) (*http.Request, error) {
```
corverroos, line 80: the only different between this function and createPostRequest is the method. Why not have a single function that takes the method as a parameter?

lib/fireblocks/api/get_transaction_by_id.go
```diff
@@ -0,0 +1,36 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/omni-network/omni/lib/errors"
+)
+
+func (c *Client) GetTransactionByID(ctx context.Context, transactionID string) (*TransactionResponse, error) {
+	var res TransactionResponse
+	endpoint := fmt.Sprintf("%s/%s", TransactionEndpoint, transactionID)
```
corverroos, line 13: the multiple nested wrapping of endpoint isn't great, lets make it super clear what path we querying by creating it in one place.

lib/fireblocks/api/create_transaction.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "newTransactionRequest")
+	}
+
+	req, err := json.Marshal(request)
+	if err != nil {
+		return nil, errors.Wrap(err, "marshal")
+	}
+
+	httpReq, err := c.createPostRequest(ctx, TransactionEndpoint, req)
```
corverroos, line 70: you are passing req in here and in sendRequest, that smells of bad abstractions.

lib/fireblocks/api/create_transaction.go
```diff
@@ -0,0 +1,128 @@
...
+		TransactionEndpoint,
+		req,
+		opts.Nonce,
+	)
+
+	if err != nil {
+		return nil, err
+	}
+
+	err = json.Unmarshal(response, &res)
```
corverroos, line 86: why not pass in response in sendRequest and do everything there? Including marshalling and unmarshalling. See `go-ethereum/rpc/client.go#CallContext` for example how to write a "generic" http request client. Sure this is json rpc, but you can follow that example.

DanFlannel: I still ended up type checking in the results but not unmarshaling, lmk if this aligns with what you were thinking

lib/fireblocks/api/client.go
DanFlannel: different hosts for different environments
Mainnet/Testnet: https://api.fireblocks.io/v1
Developer Sandbox: https://sandbox-api.fireblocks.io/v1
Figured we would have these as an environment variable

lib/fireblocks/api/create_transaction.go
DanFlannel: will do, is there any future where we will batch messages?

lib/fireblocks/api/client.go
DanFlannel: there is a different url for sandbox, vs testnet/mainnet

lib/fireblocks/api/create_transaction.go
DanFlannel: Type conversion from a passed in `response any` object that is the type of the response was a real pain and returning the string and doing marshaling here was a lot easier.

lib/fireblocks/api/create_transaction.go
corverroos: no, I do not think so

lib/fireblocks/create_transaction.go
```diff
@@ -0,0 +1,80 @@
...
+	wrappedMessage := "\x19Ethereum Signed Message:\n" + strconv.Itoa(len(opt.Message.Content)) + opt.Message.Content
+	_, err := h.Write([]byte(wrappedMessage))
+	if err != nil {
+		return CreateTransactionRequest{}, errors.Wrap(err, "sha256")
+	}
+	hash := h.Sum(nil)
+	str := hex.EncodeToString(hash)
+
+	rawSigningMessage := UnsignedRawMessage{
+		Content: str,
```
Lazar955, line 58: nit, no need for str variable

Lazar955 (commented):
- regarding file name convention, I know that go doesn't recommend any specific one, but in our codebase we use [flatcase](https://en.wikipedia.org/wiki/Naming_convention_(programming)#Examples_of_multiple-word_identifier_formats). So `createtransaction.go` vs `create_transaction.go`
- be consistent with sprintf vs string concat
- add comments above fcns and exported types

lib/fireblocks/types.go
```diff
@@ -0,0 +1,76 @@
+package fireblocks
+
+type CreateTransactionRequest struct {
+	Operation          string           `json:"operation"`
+	Note               string           `json:"note,omitempty"`
```
corverroos, line 5: lets remove unused fields

lib/fireblocks/types.go
```diff
@@ -0,0 +1,76 @@
+package fireblocks
+
+type CreateTransactionRequest struct {
```
corverroos, line 3: unexport all these types.

corverroos (changes requested):

lib/httpclient/http_client.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+	}()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", errors.Wrap(err, "read response body", "body")
+	}
+
+	bodyString := string(body)
+	log.Info(ctx, "Http: response body", "body", bodyString)
```
corverroos, line 69: in general, don't log in "libraries"

lib/fireblocks/create_transaction.go
```diff
@@ -0,0 +1,80 @@
...
+		return nil, errors.Wrap(err, "unmarshaling response")
+	}
+
+	return &res, nil
+}
+
+func NewTransactionRequest(opt TransactionRequestOptions) (CreateTransactionRequest, error) {
+	var rawMessageData RawMessageData
+	h := sha256.New()
+	wrappedMessage := "\x19Ethereum Signed Message:\n" + strconv.Itoa(len(opt.Message.Content)) + opt.Message.Content
```
corverroos, line 49: this `\x19Ethereum Signed Message` isn't required I think

lib/fireblocks/jwt.go
```diff
@@ -0,0 +1,48 @@
...
+	"encoding/json"
+	"time"
+
+	"github.com/omni-network/omni/lib/errors"
+
+	"github.com/golang-jwt/jwt/v5"
+	"github.com/google/uuid"
+)
+
+func (c Client) GenJWTToken(uri string, bodyJSON any) (string, error) {
```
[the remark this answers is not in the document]
DanFlannel: ah forgot to change this back because I thought something was wrong with the JWT, turns out it was something else

lib/fireblocks/client.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func NewClientWithConfig(apiKey string, privateKeyPath string, host string, cfg Config) (*Client, error) {
+	privateKey, err := genPrivateKey(privateKeyPath)
+	if err != nil {
+		return nil, errors.Wrap(err, "genPrivateKey")
+	}
+	httpClient := httpclient.NewClient(host, apiKey, "")
+	client := &Client{
+		apiKey:         apiKey,         // pragma: allowlist secret
```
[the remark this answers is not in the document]
DanFlannel: compiler complains about secrets being passed in

lib/fireblocks/client_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"testing"
+
+	"github.com/omni-network/omni/lib/fireblocks"
+
+	"github.com/stretchr/testify/require"
+)
+
+const sandboxHost = "https://api.fireblocks.io"
+const apiKey = ""
+const privateKeyPath = ""
```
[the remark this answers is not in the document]
DanFlannel: had this in for being able to test quickly

lib/httpclient/http_client.go
DanFlannel: good catch, missed removing this one from when I was testing

lib/fireblocks/create_transaction.go
DanFlannel: Their example has it, so I put it in

lib/fireblocks/client.go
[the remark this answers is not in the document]
corverroos: don't think this is the compiler. Just run `make secrets-baseline`. Please remove this.

lib/fireblocks/wait_signed.go
```diff
@@ -0,0 +1,65 @@
...
+			if resp == nil {
+				return nil, errors.New("failed to fetch transaction by id")
+			}
+
+			pending, err := evaluateTransactionStatus(*resp)
+			if err != nil {
+				return resp, err
+			}
+			if attempt%c.cfg.LogFreqFactor == 0 {
+				log.Warn(ctx, "Transaction not signed yet", nil, "attempt", attempt)
```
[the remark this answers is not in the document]
corverroos: just move it down after `return`?

lib/fireblocks/create_transaction.go
corverroos: Sure, but do you want transactions that work or that don't? Pretty sure it will not work.

lib/apirequestor/client.go
```diff
@@ -0,0 +1,87 @@
+package apirequestor
```
corverroos, line 1: lets avoid hasty abstractions, please remove this package

lib/fireblocks/client.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/omni-network/omni/lib/apirequestor"
+	"github.com/omni-network/omni/lib/errors"
+
+	"github.com/golang-jwt/jwt/v5"
+)
+
+const transactionEndpoint string = "v1/transactions"
+
+type FireBlocks interface {
```
corverroos, line 16: As mentioned before, the main interface should be named `Client`. You can call the implementation `HTTPClient`?

lib/fireblocks/client.go
```diff
@@ -0,0 +1,105 @@
...
+	// CreateTransaction creates a new transaction on the FireBlocks API.
+	// We use raw signing by default
+	CreateTransaction(ctx context.Context, opt TransactionRequestOptions) (string, error)
+
+	// GetTransactionById retrieves a transaction by its ID.
+	GetTransactionById(ctx context.Context, transactionID string) (*TransactionResponse, error)
+
+	// CreateAndWait waits for a transaction to be signed.
+	CreateAndWait(ctx context.Context, opts TransactionRequestOptions) (*TransactionResponse, error)
```
corverroos, line 25: remember, what we want is a signature. Not sure `TransactionResponse` with `SystemMessages` is something we want?

lib/fireblocks/client.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+type Client struct {
+	cfg            Config
+	apiKey         string
+	privateKeyPath string
```
corverroos, line 31: no need for a path if the private key is here already, please remove

lib/fireblocks/client.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	if c.privateKeyPath == "" {
+		return errors.New("private key path required")
+	}
+
+	return nil
+}
+
+// genPrivateKey generates a private key from a file.
+func genPrivateKey(privateKeyPath string) (*rsa.PrivateKey, error) {
```
corverroos, line 92: this doesn't "generate" a private key, it "loads" it from disk.

lib/fireblocks/createtransaction.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	return &res, nil
+}
+
+// newTransactionRequest creates a new transaction request.
+func newTransactionRequest(opt TransactionRequestOptions) (createTransactionRequest, error) {
+	h := sha256.New()
+	wrappedMessage := "\x19Ethereum Signed Message:\n" + strconv.Itoa(len(opt.Message.Content)) + opt.Message.Content
+	_, err := h.Write([]byte(wrappedMessage))
```
corverroos, line 55: as mentioned before, please use `sha256.Sum256()` instead

lib/apirequestor/client.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+func (c *Client) Send(ctx context.Context, endpoint string, httpMethod string, bodyJSON any, headers map[string]string) (string, error) {
+	endpoint = fmt.Sprintf("%s/%s", c.host, endpoint)
+	req, err := http.NewRequestWithContext(
+		ctx,
+		httpMethod,
+		endpoint,
+		nil,
```
corverroos, line 36: pass in the body here

lib/apirequestor/client.go
```diff
@@ -0,0 +1,87 @@
...
+
+func NewClient(host string, apiKey string, clientSecret string) *Client {
+	return &Client{
+		host:         host,
+		apiKey:       apiKey,       // pragma: allowlist secret
+		clientSecret: clientSecret, // pragma: allowlist secret
+	}
+}
+
+func (c *Client) Send(ctx context.Context, endpoint string, httpMethod string, bodyJSON any, headers map[string]string) (string, error) {
```
corverroos, line 30: return `[]byte` at least, not `string`

lib/fireblocks/createandwait.go
```diff
@@ -0,0 +1,61 @@
...
+			isCompleted, err := evaluateTransactionStatus(*resp)
+			if err != nil {
+				return resp, err
+			}
+			if attempt%c.cfg.LogFreqFactor == 0 {
+				log.Warn(ctx, "Transaction not signed yet", nil, "attempt", attempt)
+			}
+			if !isCompleted {
+				return resp, nil
+			}
```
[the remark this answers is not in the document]
corverroos: it should be "is completed, return response"

corverroos (approved):
Cool, I think this is actually good enough as a start. 🚀  Looking good

lib/fireblocks/client.go
DanFlannel: missed this one good catch

lib/apirequestor/client.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	bodyBytes, err := json.Marshal(bodyJSON)
+	if err != nil {
+		return "", errors.Wrap(err, "marshaling JSON")
+	}
+
+	if bodyJSON != nil {
+		req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
+	}
```
pavelnikolov, line 49:
```suggestion
	var body bytes.Buffer
	if err := json.NewEncoder(&body).Encode(reqData); err != nil {
		return err
	}
	
	req, err := http.NewRequestWithContext(ctx, method, endpoint, body)
	if err != nil {
		return "", errors.Wrap(err, "new http request")
	}
```

lib/fireblocks/client.go
pavelnikolov: If this is 1:1 match of the Fireblocks API, probably, we can leave it `TransactionResponse` but then we can wrap this client in a struct called, for example, `Signer` which would expose only what we need (e.g. `Signature()` or `Sign()`) and it could behind the scenes use the Fireblocks API. Then our code would depend on a `Signer` interface and we can have the Fireblocks implementation and another one with private keys for devnet/staging.

pavelnikolov:
We need metrics for this library:
* requests count with `status_code` and `endpoint` labels
* request duration histogram

PR #590, opened 2024-03-16
feat(monitor): add-monitor-service

pavelnikolov:
* Move AVS monitoring away from the Relayer
* Add `testnet` manifest
* Allow `only_monitor` manifest file which runs only the AVS monitor

task: https://app.asana.com/0/1206208509925075/1206799220447926/f


monitor/avs/avs.go
```diff
@@ -12,29 +12,26 @@ import (
...
+		return errors.New("no avs chain found")
+	}
+
 	log.Info(ctx, "Starting AVS monitor")
 
-	client, err := ethclient.Dial("goerli", goerliRPC)
+	client, err := ethclient.Dial(ch.Name, ch.RPCURL)
 	if err != nil {
-		return errors.Wrap(err, "dialing goerli")
+		return errors.Wrap(err, "dialing "+ch.Name)
```
corverroos, line 31: please learn how our errors library works, this is not correct

corverroos: So coding guidelines for details

test/e2e/app/definition.go
```diff
@@ -151,9 +155,51 @@ func LoadManifest(path string) (types.Manifest, error) {
 	return manifest, nil
 }
 
+func MonitorOnlyManifest(manifest types.Manifest, infd types.InfrastructureData, cfg DefinitionConfig) (types.Testnet, error) {
```
corverroos, line 158: raname to `MonitorOnlyTestnet` since it returns a testnet

test/e2e/app/definition.go
```diff
@@ -151,9 +155,51 @@ func LoadManifest(path string) (types.Manifest, error) {
...
+
+	return types.Testnet{
+		Network:      manifest.Network,
+		Testnet:      cmtTestnet,
+		PublicChains: publics,
+		OnlyMonitor:  true,
+	}, nil
+}
+
+func monitorCometTestnet(manifest e2e.Manifest, file string, ifd e2e.InfrastructureData) (*e2e.Testnet, error) {
```
corverroos, line 177: add comments please, explain what is different here please, same for above

test/e2e/app/definition.go
```diff
@@ -250,50 +311,45 @@ func TestnetFromManifest(manifest types.Manifest, infd types.InfrastructureData,
...
-		PortalAddress:     deployInfo[omniEVM.Chain].PortalAddress.Hex(),
-		DeployHeight:      deployInfo[omniEVM.Chain].DeployHeight,
-		BlockPeriod:       omniEVM.Chain.BlockPeriod,
-		FinalizationStrat: omniEVM.Chain.FinalizationStrat,
-		IsOmni:            true,
-	})
-
-	// Add all anvil chains
-	for _, anvil := range testnet.AnvilChains {
+	if !testnet.OnlyMonitor {
```
corverroos, line 321: add comments what this switch does, this isn't obvious why we exclude some thing when this is true/false.

test/e2e/app/run.go
```diff
@@ -42,55 +42,67 @@ func Deploy(ctx context.Context, def Definition, cfg DeployConfig) (types.Deploy
 	}
 
 	genesisValSetID := uint64(1) // validator set IDs start at 1
-	genesisVals, err := toPortalValidators(def.Testnet.Validators)
-	if err != nil {
-		return nil, nil, err
-	}
+	var genesisVals []bindings.Validator
```
corverroos, line 45: it almost feels like a `DeployMonitorOnly` is wroth the effort, that would result in less switching.

test/e2e/docker/compose.yaml.tmpl
```diff
@@ -140,6 +140,22 @@ services:
         {{ if $.Network }}ipv4_address: 10.186.73.200{{ end }}
 {{end}}
 
+{{- if .Monitor }}
+  monitor:
+    labels:
+      e2e: true
+    container_name: monitor
+    image: omniops/monitor:{{or .MonitorTag "main"}}
```
corverroos, line 148: would `RelayerTag` and `MonitorTag` ever be different? Lets combine it into a single `.OmniTag`?

test/e2e/docker/docker.go
```diff
@@ -122,6 +126,10 @@ func (p *Provider) StartNodes(ctx context.Context, nodes ...*e2e.Node) error {
 		return err
 	}
 
+	if len(nodes) == 0 {
```
corverroos, line 129: add a comment that explains why we handle this edgecase

corverroos (approved):

Lazar955 (approved):

test/e2e/docker/compose.yaml.tmpl
pavelnikolov: I added this for debug purposes potentially, however, if we are debugging we can hard-code the tag. I'll use the same tag for now.

test/e2e/app/run.go
pavelnikolov: Do you mean a separate deploy command?

monitor/avs/avs.go
pavelnikolov: I just didn't want the chain name to be hard-coded. Did you mean that you want this be an argument?

ttarsi: You can add attributes – [here's an example](https://github.com/omni-network/omni/blob/940ba4ea7a12b3a9a92f5715e18f8ed6fcd5ae6f/halo/app/start.go#L187)

PR #593, opened 2024-03-18
fix(halo/attest): only approve sequentially

corverroos:
Approve attestations sequentially. 

I.e., so only approve 3, if the previously approved is 2. Do not allow approving 4 if previously approved is 2.

This fixes the following critical bug:
- Given a vote window of 10.
- If the last approved was 2 (so vote window is [0,12]
- If we skip 3 and approved 12 next, and then we approve 14, the vote window is [4,24]
- Which means 3 will never be approved.
- This will cause the cross chain protocol to stall forever.

task: none

### DIFF
--- a/halo/attest/keeper/keeper.go
+++ b/halo/attest/keeper/keeper.go
@@ -209,15 +209,30 @@ func (k *Keeper) Approve(ctx context.Context, valset *cmttypes.ValidatorSet) err
 		valsByPower[addr] = val.VotingPower
 	}
 
-	skip := make(map[uint64]bool)           // Skip processing chains as soon as a pending attestation is found.
-	edgesByChain := make(map[uint64]uint64) // Track new minimum edges for updated vote windows.
+	approvedByChain := make(map[uint64]uint64) // Cache the latest approved attestation by chain.
 	for iter.Next() {
 		att, err := iter.Value()
 		if err != nil {
 			return errors.Wrap(err, "value")
 		}
-		if skip[att.GetChainId()] {
-			continue
+
+		// Ensure we approve sequentially, not skipping any heights.
+		{
+			// Populate the cache if not already.
+			if _, ok := approvedByChain[att.GetChainId()]; ok {
+				latest, found, err := k.latestAttestation(ctx, att.GetChainId())
+				if err != nil {
+					return errors.Wrap(err, "latest approved")
+				} else if found {
+					approvedByChain[att.GetChainId()] = latest.BlockHeader.Height
+				}
+			}
+			head, ok := approvedByChain[att.GetChainId()]
+			if ok && head+1 != att.GetHeight() {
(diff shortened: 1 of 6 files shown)


Lazar955 (approved):

halo/attest/keeper/keeper.go
```diff
@@ -209,15 +209,30 @@ func (k *Keeper) Approve(ctx context.Context, valset *cmttypes.ValidatorSet) err
...
 		if err != nil {
 			return errors.Wrap(err, "value")
 		}
-		if skip[att.GetChainId()] {
-			continue
+
+		// Ensure we approve sequentially, not skipping any heights.
+		{
+			// Populate the cache if not already.
+			if _, ok := approvedByChain[att.GetChainId()]; ok {
```
pavelnikolov, line 222: Is there a reason why our linters require us to use the proto methods instead of call fields directly? For example here can't we just use `att.ChainId` instead of having to call `att.GetChainId()`? The latter is particularly annoying locally when debugging there is one more func to step into.

corverroos: The linters don't require this. It is personal choice. Not sure we should optimise for debuggers.

pavelnikolov: Well, when we have a function call that we want to step into during debugging and the arguments we pass come from proto structs we step into the proto getters first and this extremely annoying.
Also, during debugging sessions, if we use the struct fields we can directly see the value of the struct field with a mouse hover over, whereas, with a getter method this is not possible and we have to actually call it first in order to see the value.

I just tried changing a random line from:

```go
edgesByChain[att.GetChainId()] = uintSub(att.GetHeight(), k.voteWindow)
```

to

```go
edgesByChain[att.ChainId] = uintSub(att.Height, k.voteWindow)
```
and then ran `golangci-lint run` and it returned it back to the original code. So, indeed, it is enforced by the linter(s) : (

halo/attest/keeper/keeper_test.go
```diff
@@ -302,49 +302,87 @@ func TestKeeper_Approve(t *testing.T) {
...
 			args: args{
-				valset: valset1_2_3,
+				valset: valset1_2,
 			},
 			want: want{
 				atts: []*keeper.Attestation{
-					{Id: 1, AttestationRoot: attRoot, ChainId: 1, Hash: blockHashes[0].Bytes(), Height: 500, Status: int32(keeper.Status_Pending)},
-					{Id: 2, AttestationRoot: attRoot, ChainId: 1, Hash: blockHashes[0].Bytes(), Height: 501, Status: int32(keeper.Status_Pending)},
+					expectApprovedAtt(1, 500, valset1_2),
+					expectPendingAtt(2, 502),
```
pavelnikolov, line 379: Under what circumstances 502 arrives before 501?

PR #596, opened 2024-03-18
feat(cli): scaffold forge project

idea404:
Adds command to scaffold a new omni project using the forge template.

task: https://app.asana.com/0/1206684215872507/1206684219212562/f

### DIFF
--- a/cli/cmd/cmd.go
+++ b/cli/cmd/cmd.go
@@ -14,7 +14,7 @@ func New() *cobra.Command {
 		"omni",
 		"CLI providing tools for interacting with omni",
 		newOperatorCmds(),
-		newDevelopCmds(),
+		newDeveloperCmds(),
 		newDevnetCmds(),
 		buildinfo.NewVersionCmd(),
 	)
@@ -22,7 +22,7 @@ func New() *cobra.Command {
 
 func newOperatorCmds() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "operate",
+		Use:   "operator",
 		Short: "Operator commands",
 		Args:  cobra.NoArgs,
 	}
--- a/cli/cmd/cmd.go
+++ b/cli/cmd/cmd.go
@@ -14,6 +14,7 @@ func New() *cobra.Command {
 		"omni",
 		"CLI providing tools for interacting with omni",
 		newOperatorCmds(),
+		newDevelopCmds(),
 		newDevnetCmds(),
 		buildinfo.NewVersionCmd(),
 	)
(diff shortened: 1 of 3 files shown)


cli/cmd/develop.go
```diff
@@ -0,0 +1,87 @@
...
+	if !checkForgeInstalled() {
+		// Forge is not installed, return an error with a suggestion.
+		return &cliError{
+			Msg:     "forge is not installed.",
+			Suggest: "You can install foundry by visiting https://github.com/foundry-rs/foundry",
+		}
+	}
+
+	// Execute the `forge init` command.
+	cmd := exec.Command("forge", "init", "--template", "https://github.com/omni-network/omni-forge-template.git")
```
pavelnikolov, line 60: Can this just be part of `go generate ./...`?

pavelnikolov: Is this going to be used by our operators?

idea404: Could you explain further?

pavelnikolov: What I meant is:If this is about using forge tools next to our Go codebase, then we can set this up as `go generate` command.

idea404: I see! It's like 2) - this command is intended for developers that will write programs that will be run on the Omni EVM or be deployed to any of the supported rollups too

cli/cmd/develop.go
```diff
@@ -0,0 +1,87 @@
...
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return errors.Wrap(err, "failed to run forge init")
+	}
+
+	// If a new directory was created, print its path to inform the user.
+	if projectName != "" {
```
ttarsi, line 68: We're not using `projectName` in the init command – which is fine, `forge init` initializes the template in the current directory (it doesn't create a new directory). So if you pass in a projectName that's different than the current directory, the printed directory won't match the one where the repo was actually started. 

I tried running this locally, and it forces me to use a project name (even though the project name is ignored).

Since we're not actually using project name, I think we can remove it.

ttarsi (approved):

cli/cmd/cmd.go
```diff
@@ -14,14 +14,15 @@ func New() *cobra.Command {
...
+		newDevelopCmds(),
 		newDevnetCmds(),
 		buildinfo.NewVersionCmd(),
 	)
 }
 
 func newOperatorCmds() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "operator",
+		Use:   "operate",
```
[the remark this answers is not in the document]
idea404: agreed, I'll also change `develop` to `developer` for consistency

PR #599, opened 2024-03-18
chore(contracts): refactor contracts

kevinhalliday:
Refactor contracts.

- bump to solc v0.8.24 where possible
- use less @inheritdoc, prefer actual comments
- reformat comments
- move portal storage to its own contract

task: none


contracts/src/protocol/OmniPortalStorage.sol
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * @title OmniPortalStorage
+ * @notice Storage layout for OmniPortal
+ */
+abstract contract OmniPortalStorage is IOmniPortal, IOmniPortalAdmin {
+    /**
+     * @notice Chain ID of the chain to which this portal is deployed
+     */
+    uint64 public immutable chainId;
```
ttarsi, line 16: since this isn't storage, should we keep it in the `OmniPortal`? Matching how we do it with `OmniAVS`

ttarsi (approved):

contracts/src/protocol/OmniPortalStorage.sol
kevinhalliday: Oh yes good point  (side note, I am thinking of removing this)

PR #600, opened 2024-03-18
feat(contracts): enable portal validator set updates

kevinhalliday:
Support dynamic validator via broadcast cchain xmsgs. 

task: https://app.asana.com/0/1206208509925075/1206648863973030


contracts/src/protocol/OmniPortal.sol
```diff
@@ -102,6 +106,8 @@ contract OmniPortal is IOmniPortal, IOmniPortalAdmin, OwnableUpgradeable, OmniPo
         require(gasLimit <= XMSG_MAX_GAS_LIMIT, "OmniPortal: gasLimit too high");
         require(gasLimit >= XMSG_MIN_GAS_LIMIT, "OmniPortal: gasLimit too low");
         require(destChainId != chainId, "OmniPortal: no same-chain xcall");
+        require(destChainId != _BROADCAST_CHAIN_ID, "OmniPortal: no broadcast xcall");
+        require(to != _VIRUTAL_PORTAL_ADDRESS, "OmniPortal: no portal xcall");
```
kevinhalliday, line 110: If we have reason to enable portal -> portal broadcasts xcalls in future, we can move these checks to external function.

contracts/src/protocol/OmniPortalConstants.sol
```diff
@@ -15,12 +21,18 @@ contract OmniPortalConstants {
...
+    uint64 internal constant _BROADCAST_CHAIN_ID = 0;
+
+    // @dev xmsg.sourceChainId of xmsgs from Omni's consensus chain
+    uint64 internal constant _CCHAIN_ID = 0;
+
+    // @dev xmg.sender for xmsgs from Omni's consensus chain
+    address internal constant _CCHAIN_SENDER = address(0);
+
+    // @dev xmsg.to for xcalls to be executed on the portal itselfj
+    address internal constant _VIRUTAL_PORTAL_ADDRESS = address(0);
```
kevinhalliday, line 37: We might want an address other than 0 here. Could end up being used for something else. For example, depending on how our "cross chain deployment factory" ends up working, we may want to allow CREATE txs to the 0 address. (tho I'd imagine that will probably be through some factory contract)

contracts/src/protocol/OmniPortalConstants.sol
```diff
@@ -15,12 +21,18 @@ contract OmniPortalConstants {
...
+    // @dev xmsg.destChainId for "broadcast" xcalls, intended for all portals
+    uint64 internal constant _BROADCAST_CHAIN_ID = 0;
+
+    // @dev xmsg.sourceChainId of xmsgs from Omni's consensus chain
+    uint64 internal constant _CCHAIN_ID = 0;
+
+    // @dev xmg.sender for xmsgs from Omni's consensus chain
+    address internal constant _CCHAIN_SENDER = address(0);
+
+    // @dev xmsg.to for xcalls to be executed on the portal itselfj
```
ttarsi, line 36:
```suggestion
    // @dev xmsg.to for xcalls to be executed on the portal itself
```

contracts/src/protocol/OmniPortal.sol
```diff
@@ -158,40 +163,60 @@ contract OmniPortal is IOmniPortal, IOmniPortalAdmin, OwnableUpgradeable, OmniPo
...
 
         // increment offset before executing xcall, to avoid reentrancy loop
         inXStreamOffset[xmsg_.sourceChainId] += 1;
 
-        // we enforce a maximum on xcall, but we trim to max here just in case
-        uint256 gasLimit = xmsg_.gasLimit > XMSG_MAX_GAS_LIMIT ? XMSG_MAX_GAS_LIMIT : xmsg_.gasLimit;
+        uint256 gasUsed;
+        bool success;
+
+        if (xmsg_.to == _VIRUTAL_PORTAL_ADDRESS) {
```
ttarsi, line 195: seeing some duplicated logic here, can we avoid the large if/else blocks? e.g. with ternary operators

contracts/src/protocol/OmniPortal.sol
```diff
@@ -233,33 +258,43 @@ contract OmniPortal is IOmniPortal, IOmniPortalAdmin, OwnableUpgradeable, OmniPo
         emit FeeOracleChanged(oldFeeOracle, feeOracle);
     }
 
+    /**
+     * @notice Add a new validator set.
+     * @dev Only callable via xcall from Omni's consensus chain
+     * @param valSetId      Validator set id
+     * @param validators    Validator set
+     */
+    function addValidatorSet(uint64 valSetId, XTypes.Validator[] memory validators) external {
```
ttarsi, line 267: can this be an internal function? Or does using address.call prevent that

contracts/src/protocol/OmniPortal.sol
```diff
@@ -233,33 +258,43 @@ contract OmniPortal is IOmniPortal, IOmniPortalAdmin, OwnableUpgradeable, OmniPo
...
-        require(valSetId == latestValidatorSetId + 1, "OmniPortal: invalid valSetId");
+    function _addValidatorSet(uint64 valSetId, XTypes.Validator[] memory validators) private {
         require(validators.length > 0, "OmniPortal: no validators");
 
-        // TODO: check for duplicates, consider requiring sorted input
-
         uint64 totalPower;
         XTypes.Validator memory val;
-        mapping(address => uint64) storage set = validatorSet[valSetId];
+        mapping(address => uint64) storage valset = validatorSet[valSetId];
```
ttarsi, line 284: camelcase?

contracts/src/protocol/OmniPortal.sol
ttarsi: Is this check required? Seems to me this require statement will filter out calls that are a subset of lines 260-70

ttarsi (approved):

contracts/src/protocol/OmniPortal.sol
kevinhalliday: I did that initially, but honestly thought this was clearer. A little duplication doesn't hurt.

Plus gaslimit gets awkward when you try to share too much logic

contracts/src/protocol/OmniPortal.sol
kevinhalliday: it prevents from using address(this).call
also it's part of the abi now, which helps with packing off chain (not that this would be a reason to do it)

contracts/src/protocol/OmniPortal.sol
kevinhalliday: This is saying that you can't make those calls from some evm (only the consensus chain can make those calls)

PR #603, opened 2024-03-19
refactor(halo/valsync): persist updates for attesting

corverroos:
Refactor `valsync` module to avoid the requirement of "immediate-no-delay-deterministic validator update proofs":
- Wrap staking keeper `EndBlocker`
- Intercept resulting staking keeper `validator updates` (don't provide to comet yet)
- Store them in valsync DB with the whole new validator set as new `unattested`
- When this val set is attested to, return the validator updates to `cometBFT`

task: none


Lazar955 (approved):

lib/cchain/provider/xblock.go
```diff
@@ -59,7 +59,7 @@ func toPortalVals(vals []cchain.Validator) []bindings.Validator {
 	for i, val := range vals {
 		portalVals[i] = bindings.Validator{
 			Addr:  val.Address,
-			Power: val.Power,
+			Power: uint64(val.Power),
```
kevinhalliday, line 62: Need to be careful with negatives here right? So we don't give validators slashed to negative power a ton of voting power on chain.

- could `val.Power` be negative? if not, maybe an error here if it is
- if it could be, let's either trim to 0, not include it in the set, or use `int64` on chain
     - if we trim to 0, we'll need to allow 0 power on chain (right now [we don't](https://github.com/omni-network/omni/blob/main/contracts/src/protocol/OmniPortal.sol#L281))

PR #605, opened 2024-03-19
feat(contracts): prevent old validator submissions

kevinhalliday:
Require an xsubmission uses the last seen validator set for a given source chain, or a new one. 
But never an old one.

task: https://app.asana.com/0/1206208509925075/1206874599884091


contracts/src/protocol/OmniPortal.sol
```diff
@@ -145,8 +155,17 @@ contract OmniPortal is IOmniPortal, IOmniPortalAdmin, OwnableUpgradeable, OmniPo
...
-        // update in stream block height
-        inXStreamBlockHeight[xsub.blockHeader.sourceChainId] = xsub.blockHeader.blockHeight;
+        // source chain block height of this submission
+        uint64 blockHeight = xsub.blockHeader.blockHeight;
+
+        // last seen block height for this source chain
+        uint64 lastBlockHeight = inXStreamBlockHeight[xsub.blockHeader.sourceChainId];
+
+        // update in stream block height, if it's new
+        if (blockHeight > lastBlockHeight) inXStreamBlockHeight[xsub.blockHeader.sourceChainId] = blockHeight;
```
ttarsi, line 165: is this if block necessary?

ttarsi (approved):

contracts/src/protocol/OmniPortal.sol
kevinhalliday: Just for partial xblock submissions to reduce gas

PR #606, opened 2024-03-19
feat(test/e2e): add create3-deploy command

kevinhalliday:
Add Create3 deploy command.

task: https://app.asana.com/0/1206208509925075/1206874791954734

### DIFF
--- a/test/e2e/cmd/cmd.go
+++ b/test/e2e/cmd/cmd.go
@@ -54,6 +54,7 @@ func New() *cobra.Command {
 
 	// Add subcommands
 	cmd.AddCommand(
+		newCreate3DeployCmd(&def),
 		newAVSDeployCmd(&def),
 		newDeployCmd(&def),
 		newLogsCmd(&def),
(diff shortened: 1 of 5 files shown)


pavelnikolov (approved):
LGTM 👏

test/e2e/cmd/cmd.go
```diff
@@ -137,3 +138,19 @@ func newAVSDeployCmd(def *app.Definition) *cobra.Command {
 
 	return cmd
 }
+
+func newCreate3DeployCmd(def *app.Definition) *cobra.Command {
+	cfg := app.DefaultCreate3DeployConfig()
```
ttarsi, line 143: maybe we change our naming from: `avs-deploy` and `create3-deploy` => `deploy-avs` and `deploy-create3`

Since `create3-deploy` sounds like we're deploying a contract using create3, rather than actually deploying create3 itself

ttarsi (approved):

test/e2e/cmd/cmd.go
kevinhalliday: Yeah had this discussion with pavel today. I'm more online for starting with `deploy`. Had thought of deploy subcommands `deploy create3`, `deploy avs` - the reason for `avs-deploy` and `create3-deploy` was to reduce risk of user error (i.e someone just deploying everything by typing deploy first and hitting enter)

kevinhalliday: I actually think we might subcommand on contract type. That way we could add other things

```
avs deploy
avs updgrade
avs list-operators
create3 deploy
create3 get-address // just gets address without deployting
```

PR #607, opened 2024-03-20
refactor(e2e): move e2e to root

ttarsi:
We're no longer using e2e just for testing, it is a first class citizen. This diverges from `cometbft`'s repo structure, but cometbft doesn't deploy networks into production.

Since `e2e` was a subdir of `test`, `golangci-lint` excluded it from various rules. I temporarily added `e2e` to that set, but we'll want to go back and fix those linting errors.

task: none


pavelnikolov:
I was thinking about refactoring some of this code. We actually use this for e2e testing as well. Maybe we should just move the "setup" logic to a separate lib and then reuse that between the e2e tests and some CLI tool that we use for administration purposes (deploy avs contracts, generating docker compose files, network.json, config tomls etc).

pavelnikolov (approved):
LGTM

PR #621, opened 2024-03-20
feat(contracts): add token contract

ttarsi:
add omni.sol

task: none


kevinhalliday (approved):
maybe place in `src/token/Omni.sol`?

PR #623, opened 2024-03-21
chore(contracts): refactor contract deployments

kevinhalliday:
Refactor contkract deployments.

Deployments for each contract should only be allowed under strict conditions, and their configuration should be static, and in version control.

- move deployment logic to lib/contracts
- introduce create3 and proxyadmin deployments
- remove object oriented AVSDeployer, in favor of functional deployments
- remove unnecessary production functions and configuration from avs lib

task: none


contracts/script/eigen/DeployEigenLayer.s.sol
```diff
@@ -35,6 +32,11 @@ contract DeployLocalEigenLayer is Script, EigenLayerLocal {
         string memory strategiesJson;
         for (uint256 i = 0; i < deps.strategies.length; i++) {
             IStrategy strat = IStrategy(deps.strategies[i]);
+
+            if (address(strat) == beaconChainETHStrategy) {
+                continue;
+            }
+
```
kevinhalliday, line 39: `make deploy-eigen-save-anvil-state` was broken

contracts/test/avs/common/eigen/EigenLayerFixtures.sol
```diff
@@ -81,4 +84,20 @@ contract EigenLayerFixtures is Test {
...
+
+        for (uint256 i = 0; i < strats.length; i++) {
+            _strategy[i] = IStrategy(strats[i]);
+            _thirdPartyTransfersForbiddenValues[i] = false;
+        }
+
+        strategyManager.addStrategiesToDepositWhitelist(_strategy, _thirdPartyTransfersForbiddenValues);
+
+        vm.stopPrank();
+    }
```
kevinhalliday, line 102: make `deploy-eigen-save-anvil-state` was broken

contracts/script/eigen/DeployEigenLayer.s.sol
pavelnikolov: Should we add a TODO to revert this later?

kevinhalliday: No this fixed it

lib/contracts/avs/deploy.go
```diff
@@ -0,0 +1,255 @@
...
+		_, err = backend.WaitMined(ctx, tx)
+		if err != nil {
+			return common.Address{}, nil, errors.Wrap(err, "wait mined disable allowlist")
+		}
+	}
+
+	tx, err = avs.SetMetadataURI(ownerTxOpts, cfg.MetadataURI)
+	if err != nil {
+		return common.Address{}, nil, errors.Wrap(err, "set metadata uri")
+	}
```
kevinhalliday, line 210: Might be worth adding metadata uri and and allowlist enabled / disabled to initializer, so we don't need admin txs here. 

And we might have admin controls runnable separately, so we can use this flow (with fireblocks keys) to call admin functions

lib/contracts/avs/eigen.go
```diff
@@ -0,0 +1,46 @@
...
+		return errors.New("avs directory is zero")
+	}
+	if (e.DelegationManager == common.Address{}) {
+		return errors.New("delegation manager is zero")
+	}
+
+	return nil
+}
+
+func holeskeyEigenDeployments() EigenDeployments {
```
ttarsi, line 31:
```suggestion
func holeskyEigenDeployments() EigenDeployments {
```

ttarsi (approved):
🚀

lib/chainids/chainids.go
```diff
@@ -0,0 +1,48 @@
+package chainids
```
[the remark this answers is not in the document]
kevinhalliday: Moved all chain ids here, remove arb goerli

PR #634, opened 2024-03-21
feat(lib/contracts): use create3 for proxy admin deployments

kevinhalliday:
Deploy ProxyAdmin via Create3 factory

task: none

### DIFF
--- a/lib/contracts/create3/utils.go
+++ b/lib/contracts/create3/utils.go
@@ -1,0 +1,13 @@
+package create3
+
+import (
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+// HashSalt returns the [32]byte hash of the salt string.
+func HashSalt(s string) [32]byte {
+	var h [32]byte
+	copy(h[:], crypto.Keccak256([]byte(s)))
+
+	return h
+}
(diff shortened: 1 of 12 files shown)


lib/contracts/create3/utils.go
```diff
@@ -0,0 +1,13 @@
+package create3
+
+import (
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+// HashSalt returns the [32]byte hash of the salt string.
+func HashSalt(s string) [32]byte {
```
ttarsi, line 8: maybe a test for this? might be overkill

ttarsi (approved):

lib/contracts/create3/utils.go
kevinhalliday: Do you know any good keccak hashes I could use

pavelnikolov: I don't think we need to test the Go `crypto` package. However, we might have a simple test to verify if the hashing algorithm(s) have changed. E.g. use one well known input and compare against its well known output as a sanity check.

pavelnikolov (approved):

lib/contracts/create3/utils.go
kevinhalliday: found one

kevinhalliday: yeah did that

PR #636, opened 2024-03-21
feat(lib/contracts): add portal create3 deployments

kevinhalliday:
Add Create3 portal deployments

task: https://app.asana.com/0/1206208509925075/1206874791954734

### DIFF
--- a/lib/contracts/portal/deploy.go
+++ b/lib/contracts/portal/deploy.go
@@ -1,0 +1,208 @@
+package portal
+
+import (
+	"context"
+
+	"github.com/omni-network/omni/contracts/bindings"
+	"github.com/omni-network/omni/lib/chainids"
+	"github.com/omni-network/omni/lib/contracts"
+	"github.com/omni-network/omni/lib/contracts/create3"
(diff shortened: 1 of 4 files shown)


lib/contracts/portal/deploy.go
```diff
@@ -0,0 +1,208 @@
...
+		return testnetDeployCfg(network), nil
+	}
+
+	return DeploymentConfig{}, errors.New("unsupported chain id")
+}
+
+func mainnetDeployCfg(network string) DeploymentConfig {
+	return DeploymentConfig{
+		Create3Factory: contracts.MainnetCreate3Factory,
+		Create3Salt:    network + "-portal",
```
ttarsi, line 74: Is network (a) `mainnet` `testnet` etc or (b) `optimism` `arbitrum` etc

if (a) - could we just have a `deployCfg(network string)` that returns based on the string passed in? rather than `mainnetDeployCfg("mainnet")`

if (b) - I don't think that works, since it will create a different portal address on each chain

ttarsi (approved):
how will we include staging in this model?

PR #649, opened 2024-03-25
ci(monitor/loadgen): add auto validator self-delegation

corverroos:
Add a `loadgen` package to `monitor` service that periodically self-delegates validator stake to test validator updates in staging/devnet.

Update monitor config to load those keys and e2e app to wire it.

task: none


halo/genutil/evm/evm.go
```diff
@@ -55,7 +55,7 @@ func precompilesAlloc() types.GenesisAlloc {
 // devPrefundAlloc returns allocs for pre-funded geth dev accounts.
 func devPrefundAlloc() types.GenesisAlloc {
 	eth1k := new(big.Int).Mul(
-		big.NewInt(1000),
```
corverroos, line 58: 1000 isn't enough

kevinhalliday (approved):
👌

PR #658, opened 2024-03-25
docs(home): add dynamic banner

idea404:
adds dynamic themed banner to home

task: none


ttarsi (commented):
Don't think these two look great together, let's pick one
Screenshot 2024-03-25 at 3 53 39 PM

ttarsi (approved):

PR #665, opened 2024-03-26
fix(lib/fireblocks): add signing note

corverroos:
Adds a `omni e2e <command> <network>` signing not fireblock signing requests. 

Also fix cache population race.

task: none

### DIFF
--- a/e2e/app/definition.go
+++ b/e2e/app/definition.go
@@ -66,7 +66,7 @@ type Definition struct {
 	Backends ethbackend.Backends
 }
 
-func MakeDefinition(ctx context.Context, cfg DefinitionConfig) (Definition, error) {
+func MakeDefinition(ctx context.Context, cfg DefinitionConfig, commandName string) (Definition, error) {
 	manifest, err := LoadManifest(cfg.ManifestFile)
 	if err != nil {
 		return Definition{}, errors.Wrap(err, "loading manifest")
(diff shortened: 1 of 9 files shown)


lib/fireblocks/sign.go
```diff
@@ -66,18 +67,21 @@ func (c Client) Sign(ctx context.Context, digest common.Hash, signer common.Addr
 	defer queryTicker.Stop()
 
 	var attempt int
+	prevStatus := status
 	for {
 		select {
 		case <-ctx.Done():
-			return [65]byte{}, errors.Wrap(ctx.Err(), "context canceled")
+			return [65]byte{}, errors.Wrap(ctx.Err(), "timeout waiting", "prev_status", prevStatus)
```
corverroos, line 74: return prev status on timeout

lib/fireblocks/account.go
```diff
@@ -17,10 +17,8 @@ import (
 
 // Accounts returns all the vault accounts from the account cache, populating it if empty.
 func (c Client) Accounts(ctx context.Context) (map[common.Address]uint64, error) {
-	if !c.cache.Populated() {
-		if err := c.populateAccountCache(ctx); err != nil {
```
corverroos, line 21: this was called twice in unit tests

e2e/app/definition.go
```diff
@@ -135,7 +135,9 @@ func newBackends(ctx context.Context, cfg DefinitionConfig, testnet types.Testne
 		return ethbackend.Backends{}, err
 	}
 
-	fireCl, err := fireblocks.New(testnet.Network, cfg.FireAPIKey, key)
+	fireCl, err := fireblocks.New(testnet.Network, cfg.FireAPIKey, key,
+		fireblocks.WithSignNote(fmt.Sprintf("omni e2e %s %s", commandName, testnet.Network)),
```
Lazar955, line 139: 👍

Lazar955 (approved):

PR #668, opened 2024-03-26
refactor(lib/netconf): introduce proper netconf ID type

corverroos:
Introduce a proper type  `netconf.ID` for `netconf.Staging` etc.

task: none

### DIFF
--- a/lib/netconf/network.go
+++ b/lib/netconf/network.go
@@ -1,32 +1,50 @@
 package netconf
 
+// ID is a network identifier.
+type ID string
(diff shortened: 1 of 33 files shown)


lib/netconf/network.go
```diff
@@ -1,32 +1,50 @@
 package netconf
 
+// ID is a network identifier.
+type ID string
```
corverroos, line 4: The type

Lazar955 (approved):

PR #676, opened 2024-03-26
refactor(lib/txmgr): increase default network timeout

corverroos:
Increase timeout for fireblocks signing since they are slow 🐢 

task: none

### DIFF
--- a/lib/txmgr/config.go
+++ b/lib/txmgr/config.go
@@ -49,7 +49,7 @@ var (
 		FeeLimitMultiplier:        uint64(5),
 		FeeLimitThresholdGwei:     100.0,
 		ResubmissionTimeout:       48 * time.Second,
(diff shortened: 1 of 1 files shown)


kevinhalliday (approved):

lib/txmgr/config.go
```diff
@@ -49,7 +49,7 @@ var (
 		FeeLimitMultiplier:        uint64(5),
 		FeeLimitThresholdGwei:     100.0,
 		ResubmissionTimeout:       48 * time.Second,
-		NetworkTimeout:            10 * time.Second,
+		NetworkTimeout:            30 * time.Second,
 		TxSendTimeout:             0 * time.Second,
```
pavelnikolov, line 53: Out of curiosity, why is `TxSendTimeout` set to `0`?

PR #685, opened 2024-03-27
fix(monitor): fix-avs-contract-addr

pavelnikolov:
Set AVS contract address.

task: none

### DIFF
--- a/e2e/types/chain.go
+++ b/e2e/types/chain.go
@@ -4,10 +4,9 @@ import (
 	"time"
 
 	"github.com/omni-network/omni/lib/chainids"
+	"github.com/omni-network/omni/lib/contracts"
 	"github.com/omni-network/omni/lib/errors"
 	"github.com/omni-network/omni/lib/netconf"
-
-	"github.com/ethereum/go-ethereum/common"
 )
 
 //nolint:gochecknoglobals // Static mappings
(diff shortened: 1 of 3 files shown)


e2e/types/chain.go
```diff
@@ -24,7 +25,7 @@ var (
 		IsPublic:           true,
 		BlockPeriod:        15 * time.Second,
 		FinalizationStrat:  netconf.StratLatest,
-		AVSContractAddress: contracts.TestnetAVS(),
+		AVSContractAddress: common.HexToAddress("0xa7b2e7830C51728832D33421670DbBE30299fD92"),
```
corverroos, line 28: lets remove this from `types.EVMChain` and from `netconf.Chain` and move it to `netconf.ID.Static()`.

So add `AVSContractAddress` as a field to `netconf.Static`.

kevinhalliday:
I'm confused why this fixes it - because these are the same address. I've checked that. Any thoughts on why?

pavelnikolov:
I'd assume that in the `monitor_only = true` manifest we don't go through some steps and that needs to be changed.

lib/netconf/static.go
```diff
@@ -40,6 +43,10 @@ var statics = map[ID]Static{
 	Staging: {
 		OmniExecutionChainID: 16561,
 	},
+	Testnet: {
+		AVSContractAddress:   common.HexToAddress("0xa7b2e7830C51728832D33421670DbBE30299fD92"),
```
ttarsi, line 47: does the create3-computed address not work here?

ttarsi (approved):

e2e/types/chain.go
pavelnikolov: I added just the `AVSContractAddress ` field in the testnet. I don't want to refactor the rest of the contracts now with the input EVMChain - this is the bare minimum that the Monitor service needs.

lib/netconf/static.go
pavelnikolov: It didn't work for the Monitor. But in the monitor config we don't go through all of the steps. I merge this as a workaround. This should be revisited/refactored later.

PR #695, opened 2024-03-27
fix(e2e): fix anvil fund type

Lazar955:
use proper type

task: none

### DIFF
--- a/e2e/app/fund.go
+++ b/e2e/app/fund.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"context"
+	"math/big"
 
 	"github.com/omni-network/omni/lib/anvil"
 	"github.com/omni-network/omni/lib/contracts"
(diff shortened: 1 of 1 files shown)


e2e/app/fund.go
```diff
@@ -53,9 +52,9 @@ func accountsToFund(network netconf.ID) []common.Address {
 // fundAccounts funds the EOAs that need funding (just on anvil chains, for now).
 func fundAccounts(ctx context.Context, def Definition) error {
 	accounts := accountsToFund(def.Testnet.Network)
-	eth100 := math.NewInt(100).MulRaw(params.Ether) // 100 ETH
+	eth100 := new(big.Int).Mul(big.NewInt(params.Ether), big.NewInt(100))
```
idea404, line 55: I see we also use `.NewInt()` then `.BigInt()` in `valupdates.go` and `evm.go`. 

Would changing it there be better too?

idea404 (approved):

PR #698, opened 2024-03-27
chore(contracts): handle unknown git commit in addr salts

kevinhalliday:
When a git repo is dirty, the git commit sha will be "unknown". This 
results in the same salt each time a contract is deployed in a dirty git repo.

This introduces a unique run id, to add to the salt if git commit is unknown.

task: none


kevinhalliday:
Dunno if this is the best way to solve this let me know if you have better ideas

PR #700, opened 2024-03-27
feat(graphql): xmsgrange query

DanFlannel:
Adding xmsg range query to graphql service

task: none


explorer/graphql/app/schema.graphql
```diff
@@ -118,6 +124,7 @@ type Query {
   xblock(sourceChainID: BigInt!, height: BigInt!): XBlock
   xblockrange(amount: BigInt!, offset: BigInt!): [XBlock]!
   xblockcount: BigInt
+  xmsgrange(amount: BigInt!, offset: BigInt!): [XMsg]!
```
pavelnikolov, line 127: How many messages do we potentially return here? Can this be a huge number that absolutely needs pagination?

explorer/graphql/data/block_provider.go
```diff
@@ -37,13 +37,13 @@ func (p Provider) XBlockRange(ctx context.Context, amount uint64, offset uint64)
...
-		log.Error(ctx, "Graphql provider err", err)
+		log.Error(ctx, "Ent query", err)
 		return nil, false, err
 	}
 
-	res := []*resolvers.XBlock{}
+	var res []*resolvers.XBlock
 	for _, b := range query {
-		graphQL, err := EntBlockToGraphQLBlock(b)
+		graphQL, err := EntBlockToGraphQLBlock(ctx, b)
```
pavelnikolov, line 46: nit `graphQL` -> `xBlock`

explorer/graphql/data/message_provider.go
```diff
@@ -23,3 +26,26 @@ func (p Provider) XMsgCount(ctx context.Context) (*hexutil.Big, bool, error) {
 
 	return &hex, true, nil
 }
+
+func (p Provider) XMsgRange(ctx context.Context, amount uint64, offset uint64) ([]*resolvers.XMsg, bool, error) {
```
pavelnikolov, line 30: We should implement this similar to the [Ethereums schema](https://github.com/ethereum/go-ethereum/blob/master/graphql/schema.go#L371) - we need to  have `from` and `to` block height. With the `amount` and `offset` you kind of reveal implementation details. We should change it.
Same applies to `Blocks(from, to)`.

pavelnikolov (changes requested):
Please, use `from` and `to` instead of `amount` and `offset`. To be consistent with Ethereum GraphQL and not to reveal implementation details.

DanFlannel:
@pavelnikolov I agree we need pagination and we will have it before the explorer goes live for testnet, just need everything else to be ready first 👍

explorer/graphql/app/schema.graphql
```diff
@@ -122,9 +122,9 @@ type XReceipt {
 
 type Query {
   xblock(sourceChainID: BigInt!, height: BigInt!): XBlock
-  xblockrange(amount: BigInt!, offset: BigInt!): [XBlock]!
+  xblockrange(from: BigInt!, to: BigInt!): [XBlock]!
```
pavelnikolov, line 125: I'm not sure if `to` should be required. Maybe you can fetch a block since a know block height.

DanFlannel: I'm going to hold off on this until we get pagination, if we don't have `to` be required right now, we would return from -> genesis

PR #703, opened 2024-03-27
feat(indexer): linking messages and receipts

DanFlannel:
setting edges for messages and receipts in our db and hook to ensure on create we search for associated messages or receipts

task: https://app.asana.com/0/1206208509925075/1206907488928637


explorer/ui/app/graphql/schema.graphql
```diff
@@ -0,0 +1,123 @@
+"""
```
DanFlannel, line 1: hmmmm why did this get added back

explorer/db/ent/schema/receipt.go
```diff
@@ -1,6 +1,11 @@
 package schema
 
 import (
+	"context"
+	gen "github.com/omni-network/omni/explorer/db/ent"
+	"github.com/omni-network/omni/explorer/db/ent/hook"
+	"github.com/omni-network/omni/explorer/db/ent/msg"
+	"github.com/omni-network/omni/lib/errors"
 	"time"
```
pavelnikolov, line 9: nit; shouldn't the imports be ordered automatically?

explorer/db/hooks_test.go
```diff
@@ -0,0 +1,153 @@
...
+		{
+			name: "create_block_then_receipt_then_msg",
+			prerequisites: func(t *testing.T, ctx context.Context) results {
+				t.Helper()
+				client := db.CreateTestEntClient(t)
+				t.Cleanup(func() {
+					if err := client.Close(); err != nil {
+						t.Error(err)
+					}
+				})
```
pavelnikolov, line 82: I believe that this entire block:
```go
client := db.CreateTestEntClient(t)
t.Cleanup(func() {
	if err := client.Close(); err != nil {
		t.Error(err)
	}
})
```
 has be moved inside a helper function `client := setupDB(t)` because you repeat that in every test. Then the test prerequisite would just receive a `client` argument.

explorer/db/hooks_test.go
```diff
@@ -0,0 +1,153 @@
...
+				}
+			},
+			want: want{
+				BlockCount:        2,
+				MsgCount:          1,
+				ReceiptCount:      1,
+				StrayReceiptCount: 0,
+				StrayMessageCount: 0,
+			},
+			wantErr: false,
```
pavelnikolov, line 107: `false` is the default you might as well delete this line.

explorer/db/hooks_test.go
DanFlannel: I ran into table doesn't exist errors when running the tests all at once in parellel so I moved it into the functions, however I just tested with calling a `setupDB(t)` method in each of the prerequisites, and that did work

explorer/db/hooks_test.go
```diff
@@ -0,0 +1,153 @@
...
+		MsgCount          int
+		ReceiptCount      int
+		StrayReceiptCount int
+		StrayMessageCount int
+	}
+
+	tests := []struct {
+		name          string
+		prerequisites prerequisite // These functions create entries on our db before the evaluation
+		wantErr       bool
```
pavelnikolov, line 36: You don't use `wantErr` in your tests - you might as well delete it.

explorer/db/hooks_test.go
DanFlannel: good catch!

explorer/db/hooks_test.go
```diff
@@ -0,0 +1,153 @@
...
+			results := tt.prerequisites(t, ctx)
+			actual := want{
+				len(results.blocks),
+				len(results.messages),
+				len(results.receipts),
+				calcStrayReceipts(ctx, results.receipts),
+				calcStrayMessages(ctx, results.messages),
+			}
+
+			if !cmp.Equal(tt.want, actual) {
```
pavelnikolov, line 122: Maybe an overkill (feel free to ignore), but I would compare the entire `results` struct against a known value. This way you don't only make sure that there are no stray msgs/receipts but also you make sure that the each receipt is connected to the correct message as well.

pavelnikolov (approved):

explorer/db/ent/schema/receipt.go
DanFlannel: weird, let me check my editor settings

explorer/db/hooks_test.go
DanFlannel: hmmmmm let me think about how to do this. I have to write another hook and subsequent test so I'll try to wrap those changes into that PR.

PR #727, opened 2024-04-02
fix(lib/contracts): update avs bindings and deploy cfg

kevinhalliday:
Update avs bindings, and require non zero addresses for portal and eth stake inbox on avs deployment.

task: none


kevinhalliday:
Zero addresses were disallowed in https://github.com/omni-network/omni/pull/718, but bindings were not updated.

+1 for some ci checks for binding updates

pavelnikolov (approved):

PR #734, opened 2024-04-03
ci(e2e): improve upgrade command

corverroos:
Add missing config files to upgrade command.

Also check balance before self-delegation.

task: none

### DIFF
--- a/e2e/vmcompose/provider.go
+++ b/e2e/vmcompose/provider.go
@@ -123,26 +123,40 @@ func (p *Provider) Setup() error {
 	return nil
 }
 
+// Upgrade copies some of the local e2e generated artifacts to the VMs and starts the docker-compose services.
 func (p *Provider) Upgrade(ctx context.Context) error {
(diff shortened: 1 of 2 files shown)


e2e/vmcompose/provider.go
```diff
@@ -123,26 +123,40 @@ func (p *Provider) Setup() error {
...
+	// TODO(corver): Also upgrade long-lived keys if changed
+	// - validator: ensure privval_state.json and voter_state.json doesn't complain
+
+	// Include halo config
 	for _, node := range p.Testnet.Nodes {
-		nodeDir := filepath.Join(p.Testnet.Dir, node.Name)
+		addFile(node.Name, "config/halo.toml")
+		addFile(node.Name, "config/config.toml")
+		addFile(node.Name, "config/network.json")
+		addFile(node.Name, "config/jwtsecret")
```
pavelnikolov, line 143: Can the secret change?

corverroos: it is randomly generated each time, so yes. Might as well rotate it.

PR #735, opened 2024-04-03
feat(relayer): reset worker cursors on error

corverroos:
Reset worker cursors on error. This mitigates the current bug where on-disk cursors are updated even if submissions fail.

task: none


corverroos:
See the issue here. 4 xmsgs are submitted, 2 fails, but we reset to 6 which then also fails.
image

PR #759, opened 2024-04-04
fix(contracts): attach value for mock portal calls

idea404:
Enables attaching value for mock calls

task: none


kevinhalliday (changes requested):
This is not something the real portal will be able to do, so we should not do it in the mock portal. 

what problem are you trying to solve?

idea404:
closing as portals do not relay values is a feature

PR #766, opened 2024-04-05
fix(explorer/graphql): adding mapping to xreceipts and xmsgs

DanFlannel:
fixing the missing mapping for fetch xblock in the graphql service

task: none


DanFlannel:
Unsure why these are failing tests....

PR #771, opened 2024-04-05
fix(halo/attest): only propose valid votes

corverroos:
Only proposed votes in vote window. This solves an issue identified on staging where out-of-window votes were proposed which causes the chain to stall.

Block N:
- vote Window is: [6,10] (latest approved att=8)
- so a validator votes: v1:[6] which is in the vote window
- but when block N is processed, att 9 and 10 is approved.

Block N+1:
- vote Window is: [8,12] (latest approved att=10)
- but the block proposal now includes v1:[6] (vote from previous block)
- which is outside the vote window
- so the proposal gets rejected.

Solution:
- proposer should filter valid votes.
- Previously, it simply includes all votes it has, even though some might now be invalid.

task: none

### DIFF
--- a/halo/attest/keeper/keeper.go
+++ b/halo/attest/keeper/keeper.go
@@ -568,11 +568,15 @@ func (k *Keeper) verifyAggVotes(ctx context.Context, valset ValSet, aggs []*type
 			return errors.Wrap(err, "verify aggregate vote")
 		}
 
+		errAttrs := []any{"chain_id", agg.BlockHeader.ChainId, "height", agg.BlockHeader.Height, log.Hex7("val0", agg.Signatures[0].ValidatorAddress)}
+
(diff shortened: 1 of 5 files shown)


halo/attest/keeper/keeper.go
```diff
@@ -568,19 +568,25 @@ func (k *Keeper) verifyAggVotes(ctx context.Context, valset ValSet, aggs []*type
 			return errors.Wrap(err, "verify aggregate vote")
 		}
 
+		errAttrs := []any{"chain_id", agg.BlockHeader.ChainId, "height", agg.BlockHeader.Height, log.Hex7("val0", agg.Signatures[0].ValidatorAddress)}
```
corverroos, line 571: improve logging

Lazar955 (approved):

PR #784, opened 2024-04-05
chore(e2e): diversify holesky public rpcs

kevinhalliday:
Hack to diversify holesky rpcs used in e2e.

Our e2e flow does not really support different network files for different validators, 
so this hack is quickest way to test out rpc diversity. 

task: none

### DIFF
--- a/lib/netconf/netconf.go
+++ b/lib/netconf/netconf.go
@@ -175,8 +175,36 @@ func Load(path string) (Network, error) {
 	return net, nil
 }
 
+// Below is used for a temprorary hack to shuffle the holesky rpcs
+
+var (
+	holeksyRPCs = []string{
+		"https://ethereum-holesky-rpc.publicnode.com",
+		"https://rpc.holesky.ethpandaops.io",
+		"https://1rpc.io/holesky",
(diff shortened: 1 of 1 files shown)


lib/netconf/netconf.go
```diff
@@ -175,8 +175,36 @@ func Load(path string) (Network, error) {
...
+		"https://rpc.holesky.ethpandaops.io",
+		"https://1rpc.io/holesky",
+		"https://holesky.drpc.org",
+	}
+)
+
+// nextHoleskyRPC cycles through the public Holesky RPCs.
+func nextHoleskyRPC() string {
+	rpcs := holeksyRPCs
+	holeksyRPCs = append(holeksyRPCs[1:], holeksyRPCs[0])
```
pavelnikolov, line 192: I haven't tested this but it doesn't look thread safe. Would this still work when called concurrently?

pavelnikolov (approved):

corverroos (changes requested):
Lets rather do this in e2e, since this is a devops thing

corverroos:
See [here](https://github.com/omni-network/omni/pull/787/files)

PR #785, opened 2024-04-05
fix(docker): add windows check for -r flag

DanFlannel:
Need to add `-r` flag to our stop container script, this is a bug when running our e2e runner on a window machine.

`-r, --no-run-if-empty` If the standard input does not contain any non blanks, do not run the command.  Normally, the command is run once even if there is no input.  This option is a GNU  extension.

task: https://app.asana.com/0/1206208509925075/1207003835964540


pavelnikolov (approved):

corverroos (approved):

corverroos:
Closing this as it isn't required anymore

PR #791, opened 2024-04-08
chore(monitor/account): add type to account monitor

Lazar955:
adds the type of address to the monitor

task: none

### DIFF
--- a/monitor/account/account.go
+++ b/monitor/account/account.go
@@ -11,6 +11,19 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 )
 
+type addressType string
+
+const (
+	deployer        addressType = "deployer"
+	create3Deployer addressType = "create3-deployer"
+	devFireblocks   addressType = "dev-devFireblocks"
+)
+
+type accountType struct {
+	addr        common.Address
+	addressType addressType
+}
+
 // Monitor starts monitoring account balances.
 func Monitor(ctx context.Context, network netconf.Network) error {
 	rpcClientPerChain := make(map[uint64]ethclient.Client)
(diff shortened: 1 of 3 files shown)


pavelnikolov (approved):

monitor/account/account.go
```diff
@@ -25,15 +38,15 @@ func Monitor(ctx context.Context, network netconf.Network) error {
...
+			{contracts.TestnetCreate3Deployer(), create3Deployer},
+			{contracts.TestnetDeployer(), deployer},
 		},
 		netconf.Staging: {
-			contracts.StagingCreate3Deployer(),
-			contracts.StagingDeployer(),
-			common.HexToAddress("0x7a6cF389082dc698285474976d7C75CAdE08ab7e"), // fb: dev
+			{contracts.StagingCreate3Deployer(), create3Deployer},
+			{contracts.StagingDeployer(), deployer},
+			{common.HexToAddress("0x7a6cF389082dc698285474976d7C75CAdE08ab7e"), devFireblocks}, // fb: dev
```
pavelnikolov, line 49: Do we need to monitor Relayer address as well?

monitor/account/account.go
```diff
@@ -11,6 +11,19 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 )
 
+type addressType string
```
corverroos, line 14: `accountType`

monitor/account/account.go
Lazar955: currently done by the relayer itself, we would need to supply the address to the monitor. I'll do it in next pr

PR #796, opened 2024-04-08
feat(explorer/db): toggle docker db based on if network is ephemeral

DanFlannel:
Setting the `explorerMockDB` to true if network is ephemeral

task: https://app.asana.com/0/1206208509925075/1207010240288816


e2e/vmcompose/provider.go
```diff
@@ -89,7 +89,7 @@ func (p *Provider) Setup() error {
 			Prometheus:     p.Testnet.Prometheus,
 			OmniTag:        p.omniTag,
 			Explorer:       p.Testnet.Explorer && (services["explorer_ui"] || services["explorer_graphql"] || services["explorer_indexer"]),
-			ExplorerMockDB: p.Testnet.ExplorerMockDB && services["explorer_mock_db"],
+			ExplorerMockDB: p.Testnet.Network.IsEphemeral() && services["explorer_mock_db"],
```
pavelnikolov, line 92: I wonder if we even need this for `vmcompose`? We are supposed to always use real DB in this provider 🤔

pavelnikolov (approved):

PR #805, opened 2024-04-09
ci(e2e): fix upgrade copied files

corverroos:
Copy files to correct path on remove VM. Only copy applicable services. Wire agent secrets.

task: none

### DIFF
--- a/lib/netconf/static.go
+++ b/lib/netconf/static.go
@@ -121,8 +121,3 @@ func ConsensusChainIDStr2Uint64(id string) (uint64, error) {
 
 	return resp, nil
 }
-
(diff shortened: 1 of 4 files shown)


lib/netconf/static.go
```diff
@@ -121,8 +121,3 @@ func ConsensusChainIDStr2Uint64(id string) (uint64, error) {
 
 	return resp, nil
 }
-
-// Version returns the version for the given network.
-func Version(network ID) string {
```
corverroos, line 126: not used

pavelnikolov (approved):

PR #827, opened 2024-04-11
ci(lib/tracer): add tracing to xprovider and cprovider

corverroos:
Add tracing to xprovider and cprovider in halo. Support grafana otlp collector. Wire flags in e2e.

task: none


pavelnikolov (approved):

lib/tracer/rooted.go
```diff
@@ -0,0 +1,29 @@
...
+	"hash/fnv"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// StartChainHeight returns a context and span rooted to the chain+height.
+// This creates a new trace root and should generally only by xprovider or cprovider.
+func StartChainHeight(ctx context.Context, chain string, height uint64, spanName string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
+	// Deterministic TraceID for chain+height. So all traces across all instances are correlated.
```
pavelnikolov, line 15: I thought that TraceID is a primary key. If we have multiple traces with the same ID, would the insert work 🤔

corverroos: the idea is to correlate traces for the same "job"

lib/tracer/trace.go
```diff
@@ -0,0 +1,142 @@
...
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.24.0"
+	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/noop"
+)
+
+var (
+	// tracer is the global app level tracer, it defaults to a noop tracer.
+	tracer   = noop.NewTracerProvider().Tracer("")
```
pavelnikolov, line 25: There is `noop.Tracer()` that you can use directly.

PR #828, opened 2024-04-11
fix(e2e): allow all incoming cors

idea404:
Adds an all-allow to `MakeGethConfig()` for the Cross-Origin Resource Sharing header to send to requesting browser clients. 

Allows requests to be served to any browser client without CORS errors. 

task: none


Lazar955 (commented):
Is this a recommended setting for CORS on public rpc nodes?

pavelnikolov (approved):

idea404:
> Is this a recommended setting for CORS on public rpc nodes?

It depends what browsers we want to enable to make requests to our node. If we want the node RPC to serve any browser's requests then we allow all like so. If we say only wanted a site like `https://remix.omni.org` or similar to be served by the node, then we would pass that name in.

PR #830, opened 2024-04-11
feat(explorer/graphql): xmsg query

DanFlannel:
xmsg(sourceChainID, destChainID, streamOffset)

task: https://app.asana.com/0/1206208509925075/1206970544933854

### DIFF
--- a/explorer/graphql/app/schema.graphql
+++ b/explorer/graphql/app/schema.graphql
@@ -60,6 +60,12 @@ type XMsg {
 
   "Hash of the source chain block"
   BlockHash: Bytes32!
+
+  "XBlock message was emitted in"
+  Block: XBlock!
+
+  "Receipts of the message"
+  Receipts: [XReceipt!]!
 }
 
 """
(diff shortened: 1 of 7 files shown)


explorer/graphql/app/schema.graphql
```diff
@@ -127,4 +127,5 @@ type Query {
   xmsgrange(from: BigInt!, to: BigInt!): [XMsg]!
   xmsgcount: BigInt
   xreceiptcount: BigInt
+  xmsg(sourceChainID: BigInt!, destChainID: BigInt! streamOffset: BigInt!): XMsg
```
pavelnikolov, line 130: If `streamOffset` is missing, should we return latest message?

pavelnikolov: Similarly, IMO,  the `xblock` query above should accept a nullable `height` in which case it should return the latest block - this is what Ethereum does.

pavelnikolov (approved):

explorer/graphql/app/schema.graphql
DanFlannel: I'll update this

PR #832, opened 2024-04-11
chore(lib/tracing): improve tracing

corverroos:
Improve tracing:
 - Correlate traces for each protected network+chain+height
 - Separate traces for ephemeral networks
 - Add tracing to ethclient.
 - Add semconv fields.

task: none

### DIFF
--- a/lib/tracer/rooted.go
+++ b/lib/tracer/rooted.go
@@ -5,15 +5,23 @@ import (
 	"encoding/binary"
 	"hash/fnv"
 
+	"github.com/omni-network/omni/lib/netconf"
+
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 )
 
-// StartChainHeight returns a context and span rooted to the chain+height.
+// StartChainHeight returns a context and span rooted to the network+network.Version+chain+height.
 // This creates a new trace root and should generally only by xprovider or cprovider.
-func StartChainHeight(ctx context.Context, chain string, height uint64, spanName string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
-	// Deterministic TraceID for chain+height. So all traces across all instances are correlated.
+func StartChainHeight(ctx context.Context, network netconf.ID, chain string, height uint64, spanName string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
+	// Deterministic TraceID for network+network.Version+chain+height.
+	// So all traces of the same block across all apps/instances of the same network are correlated.
+	// Note this only works for protected networks with consistent versions.
+	// Ephemeral network traces will not be correlated.
+
 	h := fnv.New128a()
+	_, _ = h.Write([]byte(network.String()))
+	_, _ = h.Write([]byte(network.Static().Version))
 	_, _ = h.Write([]byte(chain))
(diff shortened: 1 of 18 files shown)


lib/tracer/rooted.go
```diff
@@ -5,15 +5,23 @@ import (
...
-// StartChainHeight returns a context and span rooted to the chain+height.
+// StartChainHeight returns a context and span rooted to the network+network.Version+chain+height.
 // This creates a new trace root and should generally only by xprovider or cprovider.
-func StartChainHeight(ctx context.Context, chain string, height uint64, spanName string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
-	// Deterministic TraceID for chain+height. So all traces across all instances are correlated.
+func StartChainHeight(ctx context.Context, network netconf.ID, chain string, height uint64, spanName string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
+	// Deterministic TraceID for network+network.Version+chain+height.
+	// So all traces of the same block across all apps/instances of the same network are correlated.
+	// Note this only works for protected networks with consistent versions.
+	// Ephemeral network traces will not be correlated.
```
pavelnikolov, line 20: Why do we want the predictable traceID? Can we correlate the traces by tags? I want to learn more about this use case.

lib/tracer/trace.go
```diff
@@ -20,6 +20,10 @@ import (
 	"go.opentelemetry.io/otel/trace/noop"
 )
 
+const (
+	sampleRatio = 0.5 // Sample 50% of trace IDs
```
pavelnikolov, line 24: With Grafana Tempo we can easily trace 100% since they use S3 compatible storage (significantly cheaper than solutions which use a DB).

pavelnikolov (approved):

lib/tracer/trace.go
```diff
@@ -119,17 +132,20 @@ func WithOTLP(endpoint string, headers map[string]string) func(*options) {
...
+			semconv.ServiceInstanceID(ids.Instance),
+			semconv.DeploymentEnvironment(ids.Network.String()),
+		))
 	if err != nil {
 		return nil, errors.Wrap(err, "merge resource")
 	}
 
 	return sdktrace.NewTracerProvider(
-		sdktrace.WithSampler(sdktrace.AlwaysSample()), // TODO(corver): Reconsider 100% sampling.
+		sdktrace.WithSampler(sdktrace.TraceIDRatioBased(sampleRatio)),
```
pavelnikolov, line 148: As per above, no issues with 100%, IMO.

lib/tracer/rooted.go
corverroos: I basically want to view the spans of a "chain+block" across all our binaries in a single view. If grafana can do this using attributes, then we dont need this.

pavelnikolov: * If you mean all spans to be part of the same trace graph - I don't think it is possible - at least I'm not aware.
* If you mean searching by all the spans related to "chain+block" (different trace graphs), then it is possible.

I don't know if we can extend CosmosSDN to generate a trace graph of a consenus round. 
What I mean is that we can start a root span in the ProposeBlock method and then its trace ID + span ID could be propagated throughout all other validators and their methods. Not sure if that makes sense.

PR #838, opened 2024-04-12
fix(halo/voter): fix trimming issues

corverroos:
Fixes two issues in vote:
 - TrimBehind didn't trim proposed votes. It should since we shouldn't vote for them again.
 - TrimBehind interfered with the "ensure sequential votes" logic resulting in false positives. Rather keep explicit "latest vote" in state.

task: none

### DIFF
--- a/halo/comet/comet.go
+++ b/halo/comet/comet.go
@@ -5,11 +5,15 @@ import (
 
 	"github.com/omni-network/omni/lib/errors"
 	"github.com/omni-network/omni/lib/k1util"
+	"github.com/omni-network/omni/lib/tracer"
 
 	rpcclient "github.com/cometbft/cometbft/rpc/client"
 	cmttypes "github.com/cometbft/cometbft/types"
 
 	"github.com/ethereum/go-ethereum/common"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 )
 
 const perPageConst = 100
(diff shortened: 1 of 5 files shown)


halo/comet/comet.go
```diff
@@ -37,6 +38,9 @@ type adapter struct {
 // IsValidator returns true if the given address is a validator at the latest height.
 // It is best-effort, so returns false on any error.
 func (a adapter) IsValidator(ctx context.Context, valAddress common.Address) bool {
+	ctx, span := tracer.Start(ctx, "comet/is_validator")
```
corverroos, line 41: added some more tracing just for fun

halo/comet/comet.go
```diff
@@ -64,6 +68,9 @@ func (a adapter) IsValidator(ctx context.Context, valAddress common.Address) boo
 // Validators returns the cometBFT validators at the given height or false if not
 // available (probably due to snapshot sync after height).
 func (a adapter) Validators(ctx context.Context, height int64) (*cmttypes.ValidatorSet, bool, error) {
+	ctx, span := tracer.Start(ctx, "comet/validators")
```
pavelnikolov, line 71: you can add attributes to the span, for example, the height parameter in this case.

pavelnikolov (approved):

Lazar955 (approved):

halo/comet/comet.go
corverroos: yeah sure we should add this

PR #839, opened 2024-04-12
feat(explorer/ui): prometheus metrics for ui

DanFlannel:
Prometheus metrics for node server

task: https://app.asana.com/0/1206208509925075/1207009887030627


explorer/ui/app/entry.server.tsx
```diff
@@ -11,9 +11,21 @@ import { createReadableStreamFromReadable } from '@remix-run/node'
 import { RemixServer } from '@remix-run/react'
 import { isbot } from 'isbot'
 import { renderToPipeableStream } from 'react-dom/server'
+import { Counter, collectDefaultMetrics } from 'prom-client'
 
 const ABORT_DELAY = 5_000
 
+let reqsCount: any = null
```
Lazar955, line 18: maybe avoid `any`

Lazar955 (approved):

explorer/ui/app/entry.server.tsx
kevinhalliday: Yeah avoid any. Also avoid let if we can - can thro stuff below into a function that returns `Counter`.

Also - not sure we need to allow for this to be `null`? We should be able to have metrics code always runs. What changes per env is - what `env` the metric is tagged with, and whether or not those metrics are scraped .

Would rather not have to do null checks every time we want to set a metric.

PR #841, opened 2024-04-12
feat(explorer/ui): theming setup

morneluckybeard:
added theming support for explorer/ui and xblock page as well as timestamp formatting

task: https://app.asana.com/0/1206208509925075/1207021947077692


DanFlannel (approved):
LGTM

explorer/ui/app/routes/xblock.$id/route.tsx
```diff
@@ -0,0 +1,28 @@
+import { Footer } from '~/components/shared/footer'
+import Navbar from '~/components/shared/navbar'
+import { GetXBlock } from '../../components/queries/block'
+import { useParams, useSearchParams } from '@remix-run/react'
+
+export default function Index() {
+  const params = useParams()
+  const [searchParams] = useSearchParams()
+
+  const data = GetXBlock(params.id || '', searchParams.get('height') || '')
```
kevinhalliday, line 10: This is a wraps `useQuery` hook - so itself is a hook. should follow `use` naming convention yeah? `useXBlock`?

Also - we should can do this in another PR. But would like to make sure all big page queries are rendered on server before rendering the page. Right now the ui doesn't do that. 

Would be nice if we could use in-component queries and that would just work - but it seems it doesn't?

Let's use remix [loaders](https://remix.run/docs/en/main/route/loader)- and [useLoaderData](https://remix.run/docs/en/main/hooks/use-loader-data)

explorer/ui/app/components/home/blockDataTable.tsx
```diff
@@ -5,16 +5,17 @@ import { XBlock } from '~/graphql/graphql'
...
 }
 
 export default function XBlockDataTable() {
-  const d = useLoaderData<typeof loader>()
+  useLoaderData<typeof loader>()
 
-  let rows = GetBlocksInRange(0, 1000)
-  let count = GetBlockCount()
+  const rows = GetBlocksInRange(0, 1000)
+  const count = GetBlockCount()
```
kevinhalliday, line 18: same comment on `use` naming here - also,seems we have setup to `useLoaderData` - let's move queries there

kevinhalliday:
Also side note - @DanFlannel  would be great if we could get preview links for UI changes. Might be a bit involved, but worth it

DanFlannel:
@kevinhalliday definitely worth setting up vercel but only need it for when we make changes to the explorer folder....

kevinhalliday:
> @kevinhalliday definitely worth setting up vercel but only need it for when we make changes to the explorer folder....

yep

DanFlannel:
closing due to unverified commits

PR #847, opened 2024-04-12
chore(e2e): add mainnet monitor manifest

kevinhalliday:
Add monitor only mainnet manifest.

task: none

### DIFF
--- a/lib/netconf/netconf.go
+++ b/lib/netconf/netconf.go
@@ -3,11 +3,13 @@
 package netconf
 
 import (
+	"context"
 	"encoding/json"
 	"os"
 	"time"
 
 	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/log"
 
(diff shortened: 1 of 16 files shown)


ttarsi (approved):

pavelnikolov (approved):

lib/netconf/netconf.go
```diff
@@ -177,13 +180,13 @@ func Load(path string) (Network, error) {
...
 
 // Save saves the network configuration to the given path.
-func Save(network Network, path string) error {
+func Save(ctx context.Context, network Network, path string) error {
 	for _, chain := range network.Chains {
 		if chain.IsOmniConsensus {
 			continue
 		}
 		if chain.PortalAddress == (common.Address{}) {
-			return errors.New("empty portal address", "chain", chain.Name)
```
corverroos, line 186: can't error since monitor-only doesn't have portal addresses, so just WARN for what it is worth

PR #850, opened 2024-04-15
chore(lib/xchain): optimise fetching xblocks

corverroos:
Optimise xprovider `FetchBlock`:
- Previously, we did 4 sequential queries (in most cases):
  - HeaderByType
  - HeaderByNumber (only avoided for slow single finalising chains)
  - FilterXmsgLogs
  - FilerXReceiptLogs
- Now we do 3 parallel queries:
   - Used cached HeaderByType
   - Query HeaderByNumber, FilterXmsgLogs, FilerXReceiptLogs in parallel.
  
This should reduce fetch block latencies by around 3x-4x, especially for L2s that finalise in batches and for relayer which often lags a bit.

task: none

### DIFF
--- a/lib/xchain/provider/fetch.go
+++ b/lib/xchain/provider/fetch.go
@@ -12,6 +12,9 @@ import (
 	"github.com/omni-network/omni/lib/xchain"
 
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/core/types"
+
+	"golang.org/x/sync/errgroup"
 )
 
 // GetEmittedCursor returns the emitted cursor for the destination chain on the source chain,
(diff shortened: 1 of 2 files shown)


lib/xchain/provider/fetch.go
```diff
@@ -124,41 +127,64 @@ func (p *Provider) GetBlock(ctx context.Context, chainID uint64, height uint64)
...
+	})
+	eg.Go(func() error {
+		var err error
+		receipts, err = p.getXReceiptLogs(ctx, chainID, height)
+
+		return err
+	})
+
+	if err := eg.Wait(); err != nil {
+		return xchain.Block{}, false, errors.Wrap(err, "wait")
```
Lazar955, line 187: 👍

Lazar955 (approved):

PR #855, opened 2024-04-15
fix(operator): make operator mainnet docs

idea404:
Adds mainnet section to operator docs for registration

task: none


idea404:
requires re-indexing the site

ttarsi (changes requested):
Since we're just duplicating the pages, I think we can combine them into a single page, and just note that it's for both mainnet and testnet. That way, if we need to update any of these pages in the future, we don't have to update them in 2 places

ttarsi (approved):

PR #859, opened 2024-04-15
feat(explorer/ui): graphql queries

DanFlannel:
added graphql queries to ui component

task: none


explorer/ui/app/components/queries/block.tsx
```diff
@@ -15,42 +15,48 @@ export function GetBlocksInRange(from: number, to: number): XBlock[] {
...
-    })
-
-    block.Messages = msgs
     rows.push(block)
   })
 
   return rows
 }
 
+export function GetBlock(sourceChainID: number, height: number): XBlock {
```
kevinhalliday, line 35: These are hooks - so let's use `use` prefix

explorer/ui/app/components/queries/block.tsx
```diff
@@ -15,42 +15,48 @@ export function GetBlocksInRange(from: number, to: number): XBlock[] {
...
+      Messages: [],
+      Receipts: [],
+      BlockHash: '',
+      SourceChainID: '',
+      Timestamp: '',
+      UUID: '',
+    }
+  }
+
+  return data?.xblock as XBlock
```
kevinhalliday, line 57: if we are using these in components, let's return the full `useQuery` result. It has a bunch of useful stuff. Also - we can generate these use query hooks w/ proper types with our codegen tool

Second - let's try not to do these queries in components, and instead do them in [loaders](https://remix.run/docs/en/main/route/loader). It seems that just using them in components does not get us the SSR we want

DanFlannel: Wrote these helpers for Morne, just trying to make it easier and generate the code neccessary for him to call in the loaders

kevinhalliday: You can't use react hooks in the loader func,  so these won't work for that. You'll need to just query the client directly.

kevinhalliday: just do a `client.query` instead, or something like that

DanFlannel: yeah talked with him this morning, going to pull these out helpers out rn

DanFlannel: @kevinhalliday should be all set now

explorer/ui/app/components/queries/block.tsx
```diff
@@ -15,36 +15,17 @@ export function GetBlocksInRange(from: number, to: number): XBlock[] {
   var rows: XBlock[] = []
 
   data?.xblockrange.map((xblock: any) => {
```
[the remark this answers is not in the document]
DanFlannel: going to be deleted as we are moving to SSR calls

explorer/ui/app/components/queries/block.tsx
```diff
@@ -15,36 +15,17 @@ export function GetBlocksInRange(from: number, to: number): XBlock[] {
...
-        BlockHeight: '',
-        BlockHash: '',
-        Receipts: [],
-        Block: block,
-      }
-      msgs.push(xmsg)
-    })
-
-    block.Messages = msgs
     rows.push(block)
```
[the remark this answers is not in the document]
kevinhalliday: Cool 👌

kevinhalliday (approved):

PR #861, opened 2024-04-16
feat(lib/txmgr): support explicit nonces

corverroos:
Support explicit nones for `txmgr.Candidates`. Also simplify some things.

task: none


lib/txmgr/txmgr.go
```diff
@@ -61,12 +58,8 @@ type simple struct {
 
 	backend ethclient.Client
 
-	nonce     *uint64
-	nonceLock sync.RWMutex
-
-	pending atomic.Int64
-
-	closed atomic.Bool
```
corverroos, line 69: pending and closed isn't used, so I removed it

Lazar955 (approved):

PR #883, opened 2024-04-18
feat(contracts): add XRegistry and PortalRegistry

kevinhalliday:
Add XRegistry and PortalRegistry contracts.

- XRegistry is an xdapp that syncs registered contracts across all registered replicas
- the XRegistry is predeployed on the Omni EVM, and controlls XRegistryReplica contracts deployed alongside each portal
- the PortalRegistry is a predeploy on Omni's EVM that
      - tracks portal deployment addresses, heights, and finalization strategies, so that it may act as the source of truth for a "network" configuration
      - it also registers portals with the XRegistry, so that each portal can restrict xcalls to only supported chains

task: https://app.asana.com/0/home/1204669338630876/1206907486649329


contracts/Makefile
```diff
@@ -20,14 +20,14 @@ build: version ## Build contracts.
...
-			DelegationManager StrategyManager StrategyBase AVSDirectory \
+			DelegationManager StrategyManager StrategyBase AVSDirectory XRegistry XRegistryReplica PortalRegistry \
 			src/deploy/ProxyAdmin.sol:ProxyAdmin \
 			test/avs/common/MockERC20.sol:MockERC20
 
 .PHONY: bindings
 bindings: check-abigen-version build ## Generate golang contract bindings.
 	./bindings/scripts/gen.sh $(CONTRACTS)
-	./bindings/scripts/genmore.sh OmniStake Create3 TransparentUpgradeableProxy src/deploy/ProxyAdmin.sol:ProxyAdmin
+	./bindings/scripts/genmore.sh OmniStake Create3 TransparentUpgradeableProxy src/deploy/ProxyAdmin.sol:ProxyAdmin XRegistry PortalRegistry
```
pavelnikolov, line 30: Do we need the `XRegistryReplica` here as well?

kevinhalliday: No because it's not predeployed. We generate regular bindings for it

contracts/src/protocol/OmniPortalStorage.sol
```diff
@@ -45,6 +45,11 @@ abstract contract OmniPortalStorage is IOmniPortal, IOmniPortalAdmin {
      */
     address public feeOracle;
 
+    /**
+     * @notice The address of the XRegistry replica contract
```
ttarsi, line 49: just be specific with comments - add "on this chain"

ttarsi: and "on omni" where relevant

contracts/src/protocol/PortalRegistry.sol
```diff
@@ -0,0 +1,100 @@
...
+        uint64 chainId;
+        address addr;
+        uint64 deployHeight;
+        string finalizationStrat;
+    }
+
+    /**
+     * @notice Get the OmniPortal deployment for a chain.
+     */
+    function get(uint64 chainId) external view returns (Deployment memory) {
```
ttarsi, line 47: why can't we just use the mapping directly?

ttarsi: the struct?

contracts/src/protocol/PortalRegistry.sol
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * @notice Register a new OmniPortal deployment.
+     * @dev Zero height deployments are allowed for now, as we use them for "private" chains.
+     *      TODO: require non-zero height when e2e flow is updated to reflect real deploy heights.
+     */
+    function register(Deployment calldata deployment) external payable onlyOwner {
+        require(!isRegistered(deployment.chainId), "PortalRegistry: already set");
+        require(deployment.addr != address(0), "PortalRegistry: zero address");
+        require(bytes(deployment.finalizationStrat).length != 0, "PortalRegistry: empty strat");
```
ttarsi, line 73: we should probably have an enum type for the finalizationStrat

contracts/src/protocol/XRegistry.sol
```diff
@@ -0,0 +1,191 @@
...
+     *          follows.
+     *              - Add (Chain 1 -> 0x1111) and (Chain 2 -> 0x2222) to Chain 3
+     *              - Add (Chain 3 -> 0x3333) to Chain 1 and Chain 2
+     */
+    function _sync(uint64 chainId, string calldata name, address registrant, address addr) internal {
+        for (uint256 i = 0; i < _chainIds.length; i++) {
+            uint64 otherChainId = _chainIds[i];
+
+            // don't sync with self
+            if (otherChainId == chainId) continue;
```
ttarsi, line 81: don't we need to sync with self if there's a replica on this chain?

ttarsi (approved):

contracts/src/protocol/PortalRegistry.sol
kevinhalliday: We can - I just liked naming of `get` and `list`

contracts/src/protocol/PortalRegistry.sol
kevinhalliday: Talked with corver about this and we decided `string` would be better - in case we need to introduce special finalization strategy that wasn't "latest", "finalized", "safe"

kevinhalliday: What are your thoughts?

contracts/src/protocol/XRegistry.sol
kevinhalliday: Sorry think I mis understood - there is not "replica" on omni EVM. the `XRegistry` itself acts as the replica to the omni evm portal.

kevinhalliday:
@ttarsi going to merge and we can add future changes in smaller prs

PR #888, opened 2024-04-18
chore(*): tiny logging improvements

corverroos:
Remove super noisy cometbft logs. Improve flag redact logic.

task: none


lib/cmd/flags.go
```diff
@@ -44,6 +44,22 @@ func LogFlags(ctx context.Context, flags *pflag.FlagSet) error {
...
+		return u.Redacted()
+	}
+
+	// Don't redact --.*path or --.*file flags.
+	if strings.Contains(flag, "file") ||
+		strings.Contains(flag, "path") {
+		return val
+	}
+
 	if strings.Contains(flag, "token") ||
```
DanFlannel, line 63: should we delete this check here for token, password, secret etc... and make this a exclude list type of function instead of an include list?

DanFlannel (approved):
LGTM

PR #897, opened 2024-04-19
ci(e2e): debug validator insufficient balance

corverroos:
Adds validator balance to logs when funding validators. 

Also extract common `EtherBalanceAt` to reduce duplication.

task: none

### DIFF
--- a/cli/cmd/devnet.go
+++ b/cli/cmd/devnet.go
@@ -278,15 +278,12 @@ func devnetFund(ctx context.Context, cfg devnetFundConfig) error {
 		return errors.Wrap(err, "wait mined")
 	}
 
-	b, err := backend.BalanceAt(ctx, addr, nil)
+	balance, err := backend.EtherBalanceAt(ctx, addr)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


cli/cmd/devnet.go
```diff
@@ -278,15 +278,12 @@ func devnetFund(ctx context.Context, cfg devnetFundConfig) error {
...
+	balance, err := backend.EtherBalanceAt(ctx, addr)
 	if err != nil {
 		return errors.Wrap(err, "get balance")
 	}
 
-	bf, _ := b.Float64()
-	bf /= params.Ether
-
-	log.Info(ctx, "Account funded", "address", cfg.Address, "balance", fmt.Sprintf("%.2f ETH", bf))
+	log.Info(ctx, "Account funded", "address", cfg.Address, "balance", fmt.Sprintf("%.2f ETH", balance))
```
pavelnikolov, line 286: Should this be a metric instead?

pavelnikolov: Ignore, I saw that there is a metric below.

PR #901, opened 2024-04-19
ci(e2e): extend e2e timeouts

corverroos:
ci.toml now runs 4 omni_evms instead of 1. This causes general slowdown on github runners causing timeouts so extend it.

Also:
 - Small optimisation wrt eoa role verify
 - Tune statesync to be more robust

task: none


DanFlannel (approved):
LGTM

PR #904, opened 2024-04-19
fix(explorer/indexer): unique chain ids

DanFlannel:
`ChainID` in the xprovidercursor table should be unique and we should only ever have one entry for them

task: none


DanFlannel:
If we were to deploy this it would break staging based on the slack thread: https://omni-chain.slack.com/archives/C02HCJN6KNE/p1713562270565029?thread_ts=1713558445.445459&cid=C02HCJN6KNE

We do need to merge this so things break accordingly but let's fix what we know isn't working correctly first

corverroos (approved):

PR #914, opened 2024-04-22
feat(explorer/graphql): msg and receipt ids for ui

DanFlannel:
creating combined ids for ui in `graphql`

task: https://app.asana.com/0/1206208509925075/1207138914623383


kevinhalliday (commented):
What is the plan to use this id on backend outside of just having it in the gql schame? Will be be used as in index in db?

I ask because if it's just a convenice representation - maybe best to leave on the backend?

Also - why not just use this as the `ID` field? we have control over that

DanFlannel:
@kevinhalliday we use `ID` as a cursor in the pagination so the MsgID and ReceiptID are just for front end. I coulnd't find a way to come up with a calculated field in the DB so these fields aren't on the DB at all

pavelnikolov:
@kevinhalliday we need globally unique IDs because of the way some GraphQL front-end libs work. The convention set by Relay in the early days (and still followed by many libs is something like `base64("<entity>:<id>")`. Then the backend will still use the _id_ in the table corresponding to the _entitiy_.

explorer/graphql/app/schema.graphql
```diff
@@ -112,6 +115,9 @@ type XBlock {
 XReceipt represents a cross-chain receipt.
 """
 type XReceipt {
+  "Computed field for receipt id from source chain, dest chain, stream offset"
+  ReceiptID: String!
```
pavelnikolov, line 119: Shouldn't we use `ID` for everything?

explorer/graphql/resolvers/types.go
```diff
@@ -11,7 +11,7 @@ import (
 
 type XMsg struct {
 	ID                  graphql.ID
-	UUID                graphql.ID
+	MsgID               string
```
pavelnikolov, line 14: I believe we need to use `graphql.ID` for all IDs.

pavelnikolov (approved):
Please use `graphql.ID`.

kevinhalliday:
> we need globally unique IDs because of the way some GraphQL front-end libs work. The convention set by Relay in the early days (and still followed by many libs is something like base64("<entity>:<id>"). Then the backend will still use the id in the table corresponding to the entitiy.
 
 Yeah - my q was why we don't just use this new id as the only id. they are globally unique.
> we use ID as a cursor in the pagination so the MsgID and ReceiptID are just for front end. I coulnd't find a way to come up with a calculated field in the DB so these fields aren't on the DB at all, just generated by the graphql

If this is just in graphql schema, and is not the actual `ID`,  what's its purpose? 

And how do we use `ID` as a cursor? Like -  we take the `ID` of a response, and use it as `Cursor` in next pagination request? 

I don't like that. Couple reasons

- we're using `ID` in such away that assumes it's purpose as more than just an "identifier". The `ID` should just be an identifier.

For example, right now, the `ID` is the  `ID` in a certaion - right? Say we'd prefer to not show our database IDs publicly (which may be a wise decision). Or say we want to move away from this, because our database IDs are not unique across tables (they're just incrementing numbers). Then this change breaks pagination.

- the cursor should be context dependent. We don't just need a cursor onto a table. We need a cursor onto a quer


(Also, side note, I noticed only the Message type has an ID). Everything should have an ID

Maybe worth a call to discuess

DanFlannel:
@kevinhalliday what I can do is change the previous `ID` to be named `Cursor` if that makes things better. I created this new ID just as a FE request, I like my front ends dumb. We also don't store this new `ID` (one I created) in the DB as we have the fields to query based off of. So this is really for the FE to just concatenate a few fields. We need the cursor to be deterministic and not some hash, so the auto incrementing field of ID is extremely useful in this regard, given the field itself being auto incremented it makes our BE query of the DB extremely fast.

What we could do is hide the current ID (auto incrementing one) as a field from the FE by removing it from the response here and you have to pass one in or call the Msgs query itself to get some cursor you can work based off of.

What we could do is encode the cursor to hide it from the FE, I didn't implement this given time crunch but it sounds like I should

DanFlannel:
WIP, need to merge https://github.com/omni-network/omni/pull/939 first

pavelnikolov:
@kevinhalliday 
from [the docs](https://graphql.org/learn/schema/#:~:text=defining%20it%20as%20an%20ID%20signifies%20that%20it%20is%20not%20intended%20to%20be%20human%E2%80%90readable):
> The ID scalar type represents a unique identifier, often used to refetch an object or as the key for a cache. The ID type is serialized in the same way as a String; however, defining it as an ID signifies that it is not intended to be human‐readable

We could easily have a `graphql.ID` to be a hash of txn, msg, block etc. Having it as ID just means that it is not supposed to be human readable. Many systems which use auto-increment primary keys use `base64("<entity_name>:<id>")` as their globally unique ID. In our case we can just use the hash of our entity since it's going to be unique anyway.

Following GraphQL best practices, most schema types need to [implement the Node interface](https://graphql.org/learn/global-object-identification/):
```
interface Node {
    id: ID!
}
```
This was introduced initially by Facebook with the Relay client, however, has become the de facto standard. We might opt not to follow that and have a different implementation. Same goes for [best practices re pagination](https://graphql.org/learn/best-practices/#pagination) - many teams choose to [use cursors](https://graphql.org/learn/pagination/) because Relay (Facebook) introduced it initially and set the standard.

DanFlannel:
Going to close this and create a new PR

PR #915, opened 2024-04-22
chore(explorer/graphql): cap the amount of msgs returned for pagination

DanFlannel:
We should limit the amount of msgs one person can query, we also default to `25` now not `1`

task: none


DanFlannel:
@pavelnikolov want to know your thoughts here, we should probably document this as well

explorer/graphql/resolvers/messages.go
```diff
@@ -24,6 +24,8 @@ type XMsgsArgs struct {
 	Cursor *hexutil.Big
 }
 
+const MsgLimit = 50
```
pavelnikolov, line 27: We can make it configurable and accept a flag with a default of 50.

pavelnikolov (approved):

pavelnikolov:
> @pavelnikolov want to know your thoughts here, we should probably document this as well

GraphQL is a self-documenting API.

PR #917, opened 2024-04-23
feat(contracts): make portal xmsg params configurable

kevinhalliday:
Make portal xmsg params admin controlled, rather than constants. 

Additional:

- update documentation, including docs that were out of data

task: https://app.asana.com/0/1206208509925075/1207144685868128


ttarsi (approved):

contracts/src/protocol/OmniPortal.sol
```diff
@@ -55,27 +55,32 @@ contract OmniPortal is IOmniPortal, IOmniPortalAdmin, OwnableUpgradeable, OmniPo
         // cchain stream offset & block heights are equal to valSetId
         inXStreamOffset[omniCChainID_] = valSetId;
         inXStreamBlockHeight[omniCChainID_] = valSetId;
+
+        // init default values
+        _setXMsgDefaultGasLimit(200_000);
```
ttarsi, line 60: we should make these args to `initialize`

PR #923, opened 2024-04-23
test(contracts): add Quorum library tests

kevinhalliday:
Add quorum library tests. 

task: https://app.asana.com/0/1206208509925075/1207144685868132


kevinhalliday:
It's unclear why bindings are changing so much for the reqgistries. The `astId`s keep chaning in storage layout output. And it seems bytecode is changing

contracts/src/libraries/Quorum.sol
```diff
@@ -34,11 +34,14 @@ library Quorum {
 
             if (i > 0) {
                 XTypes.SigTuple memory prev = sigs[i - 1];
-                require(sig.validatorAddr != prev.validatorAddr, "OmniPortal: duplicate validator");
-                require(sig.validatorAddr > prev.validatorAddr, "OmniPortal: sigs not sorted");
+                require(sig.validatorAddr != prev.validatorAddr, "Quorum: duplicate validator");
+                require(sig.validatorAddr > prev.validatorAddr, "Quorum: sigs not sorted");
             }
 
-            if (_isValidSig(sig, digest)) votedPower += validators[sig.validatorAddr];
```
kevinhalliday, line 41: Decided to error here instead of just passing. I think that's more appropriate. Only valid signatures can be submitted. If any invalid signatures are, we revert.

PR #925, opened 2024-04-24
fix(monitor): capping min random value at 0.5

DanFlannel:
capping value between 0.5 -> 1

task: https://app.asana.com/0/1206208509925075/1207145102554110

### DIFF
--- a/monitor/loadgen/stake.go
+++ b/monitor/loadgen/stake.go
@@ -19,6 +19,8 @@ import (
 )
 
 const selfDelegateJitter = 0.2 // 20% jitter
+const minRand = 0.5
+const maxRand = 1.0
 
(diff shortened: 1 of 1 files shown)


monitor/loadgen/stake.go
```diff
@@ -19,14 +19,18 @@ import (
 )
 
 const selfDelegateJitter = 0.2 // 20% jitter
+const minRand = 0.5
+const maxRand = 1.0
```
pavelnikolov, line 23: nit: combine constants in a single `contst { }` block

pavelnikolov (approved):

corverroos (changes requested):
Tests should be deterministic. Tuning jitter doesn't make it deterministic. Nor does this actually address the problem of the validator not having balance.

See my [PR](https://github.com/omni-network/omni/pull/928/files)

PR #931, opened 2024-04-24
chore(e2e/app): cleanup later

Lazar955:
move cleaning up later in the deploy process

task: none


e2e/app/run.go
```diff
@@ -66,6 +66,10 @@ func Deploy(ctx context.Context, def Definition, cfg DeployConfig) (*pingpong.XD
 		return nil, err
 	}
 
+	if err := Cleanup(ctx, def); err != nil {
+		return nil, err
+	}
+
 	if err := Setup(ctx, def, cfg); err != nil {
```
corverroos, line 73: move Setup to up, to before public chain deployments, since it can fail easily.

e2e/app/run.go
```diff
@@ -66,6 +66,10 @@ func Deploy(ctx context.Context, def Definition, cfg DeployConfig) (*pingpong.XD
 		return nil, err
 	}
 
+	if err := Cleanup(ctx, def); err != nil {
```
corverroos, line 69: why not after start?

corverroos:
Replaced by https://github.com/omni-network/omni/pull/952

PR #938, opened 2024-04-24
test(contracts): add adversarial xcall tests

kevinhalliday:
Add tests for xcalls that call the portal itself. 

Additional - add explicit check to prevent a xmsg execution from calling `OmniPortal.xcall` directly, rather than relying on fee mechanism to prevent. 

task: https://app.asana.com/0/1206208509925075/1207144685868135


idea404 (approved):
I'd vouch for the error to be more telling, ie. `"OmniPortal:  self-call"` for self-call errors instead of an insufficient fee error

kevinhalliday:
I feel similarily. Tho devils advocate is - that that is a check that we do not need, to provide a "more telling" error message where a helpful error message is not needed.

Though I do generally think we should add explicit checks for explicit scenarios, rather than letting on check catch another.

ttarsi (approved):

PR #939, opened 2024-04-24
fix(explorer/graphql): fixing cursors

DanFlannel:
adding previous cursor and fixing cursor indices 

task: https://app.asana.com/0/1206208509925075/1207157766307988

### DIFF
--- a/explorer/db/testutil.go
+++ b/explorer/db/testutil.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"strconv"
 	"testing"
+	"time"
 
 	"github.com/omni-network/omni/explorer/db/ent"
 	"github.com/omni-network/omni/explorer/db/ent/enttest"
@@ -36,6 +37,7 @@ func CreateTestBlock(t *testing.T, ctx context.Context, client *ent.Client, heig
 		SetSourceChainID(sourceChainID).
 		SetBlockHeight(uint64(height)).
 		SetBlockHash(blockHashValue.Bytes()).
+		SetCreatedAt(time.Now()).
 		SaveX(ctx)
 
 	return *b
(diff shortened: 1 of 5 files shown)


explorer/db/testutil.go
```diff
@@ -48,6 +50,7 @@ func CreateTestBlocks(t *testing.T, ctx context.Context, client *ent.Client, cou
 	var msg *ent.Msg
 	var blocks []ent.Block
 	for i := 0; i < count; i++ {
+		time.Sleep(10 * time.Millisecond)
```
pavelnikolov, line 53: Why is this magic wait time needed? And how do you know 10ms is enough? Is there a better way to do this?

explorer/graphql/app/schema.graphql
```diff
@@ -216,12 +216,16 @@ type XMsgEdge {
...
 
-  "Next Page Cursor"
+  "Next Cursor"
+  NextCursor: BigInt!
+
+  "Has Next Page"
   HasNextPage: Boolean!
 
+  "Previous Cursor"
+  PrevCursor: BigInt!
```
pavelnikolov, line 227: ❤️

pavelnikolov: Why is the cursor of type `BigInt!`? Is it supposed to be human-readable? I thought that cursors are usually of type `graphql.ID` 🤔. The official documentation [recommends base64 encoding](https://graphql.org/learn/pagination/#:~:text=As%20a%20reminder%20that%20the%20cursors%20are%20opaque%20and%20that%20their%20format%20should%20not%20be%20relied%20upon%2C%20we%20suggest%20base64%20encoding%20them.) them and making them unique:
> As a reminder that the cursors are opaque and that their format should not be relied upon, we suggest base64 encoding them.

explorer/db/testutil.go
DanFlannel: Just to make sure the time stamp on the block entries in the test environment are different enough to sort by, I'll add some comments good catch

pavelnikolov (approved):

DanFlannel:
@pavelnikolov this was a quick fix for the FE to continue, appreciate our chatting about it, looking forward to rewriting this tomorrow

PR #943, opened 2024-04-25
refactor(lib/netconf): remove bools from netconf

corverroos:
Next step in netconf refactor: remove `IsX` fields.

task: none

### DIFF
--- a/e2e/test/e2e_test.go
+++ b/e2e/test/e2e_test.go
@@ -127,15 +127,17 @@ func test(t *testing.T, testFunc testFunc) {
 
 	if testFunc.TestOmniEVM != nil {
 		for _, chain := range network.Chains {
-			if chain.IsOmniEVM {
-				client, err := ethclient.Dial(chain.Name, chain.RPCURL)
-				require.NoError(t, err)
-
-				t.Run(chain.Name, func(t *testing.T) {
-					t.Parallel()
-					testFunc.TestOmniEVM(t, client)
-				})
+			if chain.ID != network.ID.Static().OmniExecutionChainID {
+				continue
 			}
+
+			client, err := ethclient.Dial(chain.Name, chain.RPCURL)
+			require.NoError(t, err)
+
+			t.Run(chain.Name, func(t *testing.T) {
+				t.Parallel()
+				testFunc.TestOmniEVM(t, client)
+			})
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


e2e/test/e2e_test.go
```diff
@@ -127,15 +127,17 @@ func test(t *testing.T, testFunc testFunc) {
...
 		for _, chain := range network.Chains {
-			if chain.IsOmniEVM {
-				client, err := ethclient.Dial(chain.Name, chain.RPCURL)
-				require.NoError(t, err)
-
-				t.Run(chain.Name, func(t *testing.T) {
-					t.Parallel()
-					testFunc.TestOmniEVM(t, client)
-				})
+			if chain.ID != network.ID.Static().OmniExecutionChainID {
```
idea404, line 130: Why not use the helper here?

e.g. 

`if !(netconf.IsOmniConsensus(network.ID, chain.ID)) {...}`

lib/netconf/netconf.go
```diff
@@ -72,7 +72,7 @@ func (n Network) ChainNamesByIDs() map[uint64]string {
 // OmniEVMChain returns the Omni execution chain config or false if it does not exist.
 func (n Network) OmniEVMChain() (Chain, bool) {
 	for _, chain := range n.Chains {
-		if chain.IsOmniEVM {
+		if chain.ID == n.ID.Static().OmniExecutionChainID {
```
idea404, line 75: idem, helper here?

lib/netconf/static.go
```diff
@@ -135,3 +135,8 @@ func ConsensusChainIDStr2Uint64(id string) (uint64, error) {
 
 	return resp, nil
 }
+
+// IsOmniConsensus returns true if provided chainID is the omni consensus chain for the network.
+func IsOmniConsensus(network ID, chainID uint64) bool {
```
idea404, line 140: Matter of preference but I would prefer to define this as:

`func IsOmniConsensus(network network, chainID chain) bool { ... }`

That way it is always called the same way throughout the codebase

idea404 (approved):
LGTM. Some comments

e2e/test/e2e_test.go
corverroos: I only added the helper when used more than 3 times

lib/netconf/static.go
corverroos: one doesn't always have a network, sometimes you only have ID

PR #954, opened 2024-04-25
feat(contracts): dynamic v1 fee model

kevinhalliday:
Improve v1 fee model.

- make it handle changes in asset and gas prices
- charge base gas limit
- allow for protocol fee


kevinhalliday:
Couple things:

- did not update bindings, because changing e2e will be a bit of work. will leave for another PR
- I just changed FeeOracleV1, rather than introducing a FeeOracleV2. Because v1 is not live anywhere, I see no need to keep it around

contracts/src/protocol/FeeOracleV1.sol
```diff
@@ -11,36 +11,133 @@ import { IFeeOracleV1 } from "../interfaces/IFeeOracleV1.sol";
...
+            _setToNativeRate(p.chainId, p.toNativeRate);
+        }
+    }
+
+    /**
+     * @notice Set the gas price for a destination chain.
+     */
+    function _setGasPrice(uint64 chainId, uint256 gasPrice) internal {
+        require(gasPrice > 0, "FeeOracleV1: no zero gas price");
+        require(chainId != 0, "FeeOracleV1: no zero chain id");
```
pavelnikolov, line 111: Can `chaindId` be negative?

pavelnikolov (approved):
LGTM

ttarsi (approved):

contracts/src/protocol/FeeOracleV1.sol
ttarsi: nope, it's a uint

PR #959, opened 2024-04-27
refactor(lib/netconf): replace on-disk with on-chain

corverroos:
Replaces on-disk `network.json` with on-chain `PortalRegistry`.

task: none


e2e/test/e2e_test.go
```diff
@@ -302,3 +305,23 @@ func fetchBlockChain(ctx context.Context, t *testing.T) []*cmttypes.Block {
 
 	return blocks
 }
+
+func makePortalRegistry(network netconf.ID, endpoints xchain.RPCEndpoints) (*bindings.PortalRegistry, error) {
```
corverroos, line 309: note this helper function is duplicated in a few places. I'm not sure where to put a common one, since it depends on a bunch of different packages. Maybe a common `wire` package. But not sure it is worth it. So just duplicated it for now.

e2e/app/run.go
```diff
@@ -264,7 +264,7 @@ func toPortalValidators(validators map[*e2e.Node]int64) ([]bindings.Validator, e
 func logRPCs(ctx context.Context, def Definition) {
 	network, endpoints := externalNetwork(def)
 	for _, chain := range network.EVMChains() {
-		rpc, _ := endpoints.GetByNameOrID(chain.Name, chain.ID)
+		rpc, _ := endpoints.ByNameOrID(chain.Name, chain.ID)
```
pavelnikolov, line 267: nit: don't ignore the error, might fail in future.

e2e/test/e2e_test.go
pavelnikolov: Maybe we need a `portal` or `portalregistry` package?

pavelnikolov (approved):

e2e/app/run.go
corverroos: it will just log an empty value which is fine in this case

e2e/test/e2e_test.go
corverroos: I'm not so sure.  A package with just a single function? Doesn't feel worth it.

PR #963, opened 2024-04-29
fix(halo): fixes few post-refactor bugs

corverroos:
Fixes a few issues introduced by the netconf refactor:
 - omni_consensus chain validator sets start at 1, not 0, so set `DeployHeight=1`
   - Add e2e tests for validator updates (had to make devnet validator keys deterministic)
 - Replace `mock_l1` on staging with `slow_l1`
   - Also remove custom workarounds for slow staging chains since we now have explicit slow chains.

Also fix two bugs in attest keeper:
 - Boolean flip while ensuring strict-sequential approvals.
 - Don't delete anything after latest approve attestation since this could cause gaps once catching up.

task: none

### DIFF
--- a/e2e/app/definition.go
+++ b/e2e/app/definition.go
@@ -392,9 +392,6 @@ func TestnetFromManifest(ctx context.Context, manifest types.Manifest, infd type
 		if infd.Provider == docker.ProviderName {
 			internalIP = chain.Name // For docker, we use container names
 		}
-		if infd.Provider == vmcompose.ProviderName {
-			chain.BlockPeriod = time.Second * 12 // Slow block times for anvils on long-lived VMs to reduce disk usage.
-		}
 
 		anvils = append(anvils, types.AnvilChain{
 			Chain:       chain,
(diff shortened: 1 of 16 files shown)


idea404 (approved):

e2e/app/definition.go
```diff
@@ -392,9 +392,6 @@ func TestnetFromManifest(ctx context.Context, manifest types.Manifest, infd type
 		if infd.Provider == docker.ProviderName {
 			internalIP = chain.Name // For docker, we use container names
 		}
-		if infd.Provider == vmcompose.ProviderName {
```
corverroos, line 395: not required anymore

e2e/app/definition.go
```diff
@@ -424,8 +421,13 @@ func TestnetFromManifest(ctx context.Context, manifest types.Manifest, infd type
 // getOrGenKey gets (based on manifest) or creates a private key for the given node and type.
 func getOrGenKey(ctx context.Context, manifest types.Manifest, nodeName string, typ key.Type) (key.Key, error) {
 	addr, ok := manifest.Keys[nodeName][typ]
-	if !ok {
-		// No key in manifest, generate a new one.
+	if !ok { // No key in manifest
+		// Generate an insecure deterministic key for devnet
+		if manifest.Network == netconf.Devnet {
+			return key.GenerateInsecureDeterministic(manifest.Network, typ, nodeName), nil
```
corverroos, line 427: deterministic keys so e2e tests can verify and check validator updates which requires the same key

PR #969, opened 2024-04-29
feat(e2e): deploy new fee oracle v1

kevinhalliday:
Update FeeOracleV1  bindings and deployment.

Add initial FeeManager, now just used for FeeOracleV1 deployment.
It will be used in long lived fee manager process, that keeps on chain fee params 
in sync with the state of the world. 


contracts/src/protocol/OmniPortal.sol
```diff
@@ -137,13 +137,13 @@ contract OmniPortal is
...
-        require(msg.value >= feeFor(destChainId, data, gasLimit), "OmniPortal: insufficient fee");
-        require(gasLimit <= xmsgMaxGasLimit, "OmniPortal: gasLimit too high");
-        require(gasLimit >= xmsgMinGasLimit, "OmniPortal: gasLimit too low");
         require(destChainId != chainId(), "OmniPortal: no same-chain xcall");
         require(destChainId != _BROADCAST_CHAIN_ID, "OmniPortal: no broadcast xcall");
         require(to != _VIRTUAL_PORTAL_ADDRESS, "OmniPortal: no portal xcall");
+        require(gasLimit <= xmsgMaxGasLimit, "OmniPortal: gasLimit too high");
+        require(gasLimit >= xmsgMinGasLimit, "OmniPortal: gasLimit too low");
+        require(msg.sender != address(this), "OmniPortal: portal cannot xcall");
+        require(msg.value >= feeFor(destChainId, data, gasLimit), "OmniPortal: insufficient fee");
```
kevinhalliday, line 146: Just changed some require ordering, so we do feeFor last. So we avoid "no fee parameters" errors for cases that should be handled by other statments (like "no same-chain xcall")

e2e/netman/pingpong/pingpong.go
```diff
@@ -123,7 +123,7 @@ func (d *XDapp) fund(ctx context.Context) error {
 			return err
 		}
 
-		fund := new(big.Int).Div(big.NewInt(params.Ether), big.NewInt(10)) // Fund it with 0.1 ETH
+		fund := new(big.Int).Div(big.NewInt(params.Ether), big.NewInt(2)) // Fund it with 0.5 ETH
```
kevinhalliday, line 126: Note I needed to bump ping pong funding, because we are charging real fees now. 

Could use some more visibility into actual fees paid - added a [ticket](https://app.asana.com/0/1206208509925075/1207199096986774) to add fees to XMsg events.

ttarsi (approved):
nice!

e2e/netman/manager.go
```diff
@@ -353,6 +359,19 @@ func (m *manager) deployIfNeeded(ctx context.Context, chain types.EVMChain, back
...
 
+// destChainIDs returns all configured destination chain ids for a given source.
+func (m *manager) destChainIDs(srcChainID uint64) []uint64 {
+	var destChainIDs []uint64
+
+	for id := range m.Portals() {
+		if id != srcChainID {
+			destChainIDs = append(destChainIDs, id)
+		}
+	}
```
pavelnikolov, line 370: Is there a case where we don't want some chain to have another chain as destination apart from `srcID == destID`? Is it possible that this is required in future?

lib/xfee/manager.go
```diff
@@ -0,0 +1,123 @@
+package xfee
```
pavelnikolov, line 1: package level comment required

lib/price/provider.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"context"
+
+	"github.com/omni-network/omni/lib/tokens"
+)
+
+// Provider is the price provider interface.
+type Provider interface {
+	// GetPriceUSD returns the price of each provided token in USD.
+	GetPriceUSD(ctx context.Context, tokens ...tokens.Token) (map[tokens.Token]float64, error)
```
pavelnikolov, line 12: IMHO, a Get prefix is not needed here. I believe `Price` reads much better. Also, do we every expect other currency other than USD? If yes, the currency needs to be a parameter and the method will probably be called `Convert` and the interface `Converter`. If we are unlikely to use a different currency, I'd just leave it `Pricer` and `Price`.

lib/tokens/tokens.go
```diff
@@ -0,0 +1,42 @@
...
+		OMNI: "omni-network",
+		ETH:  "ethereum",
+	}
+)
+
+func (t Token) String() string {
+	return string(t)
+}
+
+func (t Token) CoingeckoID() string {
```
pavelnikolov, line 21: If that's the same as `String` do we really need it here?

lib/tokens/tokens.go
```diff
@@ -0,0 +1,42 @@
...
+	return coingeckoIDs[t]
+}
+
+func FromCoingeckoID(id string) (Token, bool) {
```
pavelnikolov, line 25: Is it possible that in future we use something other than **Coingecko**?

pavelnikolov (approved):

lib/tokens/tokens.go
kevinhalliday: Yes. Tho I have a feeling ids will be the same. Will deal with it then

lib/tokens/tokens.go
kevinhalliday: They're different. Ex "ETH" and "ethereum"

e2e/netman/manager.go
kevinhalliday: Perhaps. But I can't think of a reason why rn

PR #972, opened 2024-04-30
chore(ci): add explorer ui preview

kevinhalliday:
Add explorer ui preview. Uses same pattern as docs preview.

task: none


pavelnikolov (approved):

kevinhalliday:
Marking as draft, and not merging, because it seems there are some more hoops we need to jump through to just get vercel to work with our previews, but not use it for prod deployment

idea404 (approved):

PR #973, opened 2024-04-30
fix(test): use wait groups for flaky test

idea404:
Seeks to address flakiness found in: https://github.com/omni-network/omni/actions/runs/8897284166

task: none

### DIFF
--- a/lib/cchain/provider/provider_test.go
+++ b/lib/cchain/provider/provider_test.go
@@ -16,6 +16,7 @@ func TestProvider(t *testing.T) {
 	t.Parallel()
 
 	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel() // Ensure all paths cancel the context to avoid context leak
 
 	const (
 		errs  = 2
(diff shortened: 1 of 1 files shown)


lib/cchain/provider/provider_test.go
```diff
@@ -16,6 +16,7 @@ func TestProvider(t *testing.T) {
 	t.Parallel()
 
 	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel() // Ensure all paths cancel the context to avoid context leak
```
pavelnikolov, line 19: This makes total sense here, however, I do not see other paths that can result in `cancel` being called. Unless that happens inside of the `p.Subscribe` method.
The wait group makes more sense here.

idea404: I could not find any use of `cancel` within `p.Subscribe`, so I removed the use of `.WithCancel` and `cancel()` here, but then the test was flaky again. So I'll keep the `cancel()` use here since it is being referenced within functions called in `p.Subscribe`

PR #974, opened 2024-04-30
refactor(explorer): explorer refactoring

pavelnikolov:
* refactor Explorer
* rename/remove fields
* add doc comments to GraphQL schema

task: none

### DIFF
--- a/explorer/graphql/app/schema.graphql
+++ b/explorer/graphql/app/schema.graphql
@@ -39,7 +39,7 @@ type XMsg {
   "ID of the XMsg"
   ID: ID!
 
-  "Monotonically incremented offset of Msg in the Steam"
+  "Monotonically incremented offset of Msg in the Stream"
   StreamOffset: BigInt!
 
   "Sender on source chain, set to msg.Sender"
@@ -86,9 +86,6 @@ type XMsg {
 XBlock represents a cross-chain block.
 """
 type XBlock {
-  "UUID of our block"
-  UUID: ID!
-
   "Source chain ID as per https://chainlist.org"
   SourceChainID: BigInt!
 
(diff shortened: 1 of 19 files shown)


explorer/graphql/app/schema.graphql
```diff
@@ -230,24 +226,40 @@ type PageInfo {
...
   xblock(sourceChainID: BigInt!, height: BigInt!): XBlock
+
+  "Retrieve a range of XBlocks by source chain ID and height."
   xblockrange(from: BigInt!, to: BigInt!): [XBlock]!
+
+  "Get the total count of XBlocks."
   xblockcount: BigInt
+
+  "Retrieve a range of XMsgs by source chain ID and height."
   xmsgrange(from: BigInt!, to: BigInt!): [XMsg]!
```
kevinhalliday, line 243: Do we use these? 

`xblockrange` in particular seems a little useless. Like - `from` and `to` what? we have many source chains. We'd need to at least add source chain (also comment isn't representative)

If we don't use any queries I say we remove

kevinhalliday (approved):

explorer/graphql/app/schema.graphql
pavelnikolov: XBlock has height so it's for example from height 5 to height 15. Regardless of source chain id. Does that make sense?

Good catch on the comment - I'll fix it.

pavelnikolov: `xmsgrange` seems to be used by the UI.

idea404: I belive this is used here https://github.com/omni-network/omni/blob/7e228e9a04cd9cc0aae326378d0444ca6c042f65/explorer/ui/app/routes/_index.tsx#L51

I think `to` and `from` are ID fields, so, you get the max count of xmsgs first, then `to` becomes that and `from` is `to` minus the number of messages you want to have

PR #976, opened 2024-05-01
feat(monitor/xfeemngr): add gas price buffer

kevinhalliday:
Add gas price buffered stream to xfeemngr.

### DIFF
--- a/monitor/xfeemngr/gasprice/gasprice.go
+++ b/monitor/xfeemngr/gasprice/gasprice.go
@@ -1,0 +1,110 @@
+package gasprice
+
+import (
+	"context"
+	"sync"
+
+	"github.com/omni-network/omni/lib/log"
+
+	"github.com/ethereum/go-ethereum"
+)
+
+type Buffer struct {
+	mu sync.RWMutex
+
+	// map chainID to buffered gas price (not changed if outside threshold)
+	buffer map[uint64]uint64
+
+	// map chainID to provider
+	pricers map[uint64]ethereum.GasPricer
+
+	// options
+	opts *Opts
+}
+
+// NewBuffer creates a new gas price buffer.
+//
+// A gas price buffer maintains a buffered view of gas prices for multiple
+// chains. Buffered gas prices are not updated unless they are outside the
+// threshold percentage. Start steaming gas prices with Buffer.Stream(ctx).
+func NewBuffer(pricers map[uint64]ethereum.GasPricer, opts ...func(*Opts)) Buffer {
+	return Buffer{
+		buffer:  make(map[uint64]uint64),
+		pricers: pricers,
+		opts:    makeOpts(opts),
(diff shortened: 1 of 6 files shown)


monitor/xfeemngr/gasprice/gasprice.go
```diff
@@ -0,0 +1,149 @@
...
+		// if price is buffed, and not outside threshold, return
+		buffed, ok := b.price(chainID)
+		if ok && !isOutsideThreshold(price, buffed, b.opts.thresholdPct) {
+			return
+		}
+
+		b.setPrice(chainID, price)
+	}
+
+	go subscribe(ctx, stream(ctx, b.pricer(chainID), b.opts.streamInterval), callback)
```
kevinhalliday, line 78: I do not need a separate `stream` and `subcribe` that comm via a channel. We can just use the callback in the polling directly.

Channel was there becaes in previous design I was thinking of multiple subscribers

kevinhalliday: Though I do kinda like the way they work together

monitor/xfeemngr/gasprice/gasprice_test.go
```diff
@@ -0,0 +1,107 @@
...
+		3: mocks[3],
+	}
+
+	thresh := 0.1
+	interval := 100 * time.Millisecond
+
+	b := gasprice.NewBuffer(pricers, gasprice.WithThresholdPct(thresh), gasprice.WithStreamInterval(interval))
+	b.Stream(context.Background())
+
+	time.Sleep(5 * interval)
```
pavelnikolov, line 65: I don't like time sleeps in tests, TBH. Ideally, we need some Clock interface, which implements the `Now()` method. Then we can adjust the clock back and forth instead of having to sleep and wait for something to happen.

pavelnikolov (approved):

monitor/xfeemngr/gasprice/gasprice.go
kevinhalliday: Yeah removing this now - can introduce if we need multiple subscribers

monitor/xfeemngr/gasprice/gasprice_test.go
kevinhalliday: Yeah I also prefer not to time.Sleep in tests but did not think there was a way around int

kevinhalliday: I can refactor out the ticker

kevinhalliday: yeah that's the plan - but refactoring it out so that I can use a mock ticker

kevinhalliday: done

PR #985, opened 2024-05-03
fix(docs/cli): add tools section and fix cli devnet commands

idea404:
This PR adds a tools section to the docs. 

Also fixes omni CLI devnet commands.

task: https://app.asana.com/0/1206208509925075/1207188240612896/f


cli/cmd/devnet.go
```diff
@@ -172,7 +170,6 @@ func devnetDefinition(ctx context.Context) (app.Definition, error) {
 
 	defCfg := app.DefaultDefinitionConfig(ctx)
 	defCfg.ManifestFile = manifestFile
-	defCfg.OmniImgTag = buildinfo.Version()
```
idea404, line 175: I removed this line as the devnet would not start with `"main"` as a value here either. Removing this avoided that. I'd appreciate a second look here

lib/netconf/provider.go
```diff
@@ -81,6 +81,57 @@ func AwaitOnChain(ctx context.Context, netID ID, portalRegistry *bindings.Portal
 	}
 }
 
+// CheckOnChain returns the network configuration if all expected chains are registered in the on-chain registry.
+// It returns an error if the context is canceled or if any of the expected chains are missing.
+func CheckOnChain(ctx context.Context, netID ID, portalRegistry *bindings.PortalRegistry, expected []string) (Network, error) {
+	if netID == Simnet {
```
kevinhalliday, line 87: Do we need this for this func? If yes - maybe just move this dummy simnet config into one place and use it here and in AwaitOnChain?

cli/cmd/devnet.go
kevinhalliday: This would be if building from source no, but should be fine in a release?

kevinhalliday (approved):

cli/cmd/devnet.go
idea404: It should be fine in a release I though too. But then `"main"` should have worked here too as that would mimic the same behaviour as when using the buildinfo tag version here.

lib/netconf/provider.go
idea404: Good one, no this would only be called for devnet.

cli/cmd/devnet.go
idea404: I'll use this line again, will test before publishing release on GH

PR #986, opened 2024-05-03
feat(contracts): add feeoracle manager role

kevinhalliday:
Add feeoracle manager role, separate from owner, that can set on chain fee parameters.

task: https://app.asana.com/0/1206208509925075/1207233529904017


contracts/src/protocol/FeeOracleV1.sol
```diff
@@ -21,6 +21,11 @@ contract FeeOracleV1 is IFeeOracle, IFeeOracleV1, OwnableUpgradeable {
      */
     uint256 public protocolFee;
 
+    /**
+     * @notice Address allow to set gas prices and to-native conversion rates.
```
ttarsi, line 25:
```suggestion
     * @notice Address allowed to set gas prices and to-native conversion rates.
```

contracts/src/protocol/FeeOracleV1.sol
```diff
@@ -93,6 +107,14 @@ contract FeeOracleV1 is IFeeOracle, IFeeOracleV1, OwnableUpgradeable {
         _setProtocolFee(fee);
     }
 
+    /**
+     * @notice Set the manager admin account.
+     */
+    function setManager(address manager_) external onlyOwner {
+        require(manager_ != address(0), "FeeOracleV1: no zero oracle");
```
ttarsi, line 114: no zero `manager`? to align terms

ttarsi (approved):

PR #989, opened 2024-05-06
feat(docs): warning for dynamic code blocks

idea404:
This PR introduces a dynamic warning to code snippets in docs that will warn whenever code snippets do not match with the same code in `main`. 

This also introduces the restriction that code snippets may only be created with URLs containing a commit hash. 

To test this feature, you can add this code to any markdown file (such as for example in a new file `./docs/site/docs/test.md`) and visit it:

```markdown
# 

import GitHubCodeBlock from '@site/src/components/GitHubCodeBlock/GitHubCodeBlock';

## Out of Date Code

<GitHubCodeBlock repoUrl="https://github.com/omni-network/omni/blob/97df20bfb1411f30b01d79411cb8b3d0f9213573/contracts/src/pkg/XApp.sol#L60-L69" />

## Code Line Ref Without Commit Hash

<GitHubCodeBlock repoUrl="https://github.com/omni-network/omni/blob/v0.1.0/contracts/src/pkg/XApp.sol#L60-L69" />
```

task: https://app.asana.com/0/1206208509925075/1207138429056637/f


kevinhalliday (commented):
Did most of review via slack huddle - will summarize here

- don't reinvent fetching async state w/ react, use a tool like react-query
- less closures with side effects defined in component. more pure functions 
- try to have component tell a terse story - "given props, render this"

docs/site/src/components/GitHubCodeBlock/GitHubCodeBlock.tsx
```diff
@@ -1,93 +1,58 @@
...
-        const noLinesRegex = /(https:\/\/github\.com\/)([^\/]+)\/([^\/]+)\/blob\/([^\/]+)\/([^#]+)/;
-        match = repoUrl.match(noLinesRegex);
+    const fetchData = async () => {
+      try {
+        const result = await useGitHubCodeBlock(repoUrl);
+        setData(result);
+      } catch (error) {
+        console.error(error);
       }
+    };
```
kevinhalliday, line 20: A hook is either a base react hook, or a function that uses a base react hook. In this case `useGithubCodeBlock`. It just fetches the data. 

I suggest moving `useEffect` and `useState` logic into `useGithubCodeBlock`, and using that as a hook.

kevinhalliday (commented):
Yep sorry - forgot about docusaurus context when telling you to use react-query.

You can still do it though. Here is an [example how](https://github.com/omni-network/omni/pull/1011).

That example also does some other refactoring - to show some options for ts style coding. And it prefectches known codeblock queries, to show that we can do that if we want.

idea404:
That PR was 🔥 @kevinhalliday ty! 

I looked into adding some form of automation for the `prefetch` url Array but it seems as it would involve adding a step to the build process, or a pre-commit step. For this case it seems a little overkill to me and updating the links manually also adds overhead, so I removed the use of prefetch.

kevinhalliday (approved):

kevinhalliday:
Another thing you can do to reduce loading "flashes" - add expected line heigh that you set upfront for an empty codeblock, that can be used as real code is loading. That way, the height of the codeblock (and placement of rest of elements on the page) does not change when it loads

idea404:
> Another thing you can do to reduce loading "flashes" - add expected line heigh that you set upfront for an empty codeblock, that can be used as real code is loading. That way, the height of the codeblock (and placement of rest of elements on the page) does not change when it loads

Good idea! Sadly, for normal github URLs without line numbers provided the client would have to do an API request to obtain the length, but I used the parser function to size the block for whenever the lines are specified for snippets

PR #996, opened 2024-05-06
feat(halo/cmd): support trusted state sync init

corverroos:
Adds a `--trusted-sync` flag to `halo init` which sets the trusted state-sync fields by querying the omni consensus RPC.

task: none

### DIFF
--- a/halo/cmd/init.go
+++ b/halo/cmd/init.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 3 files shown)


halo/cmd/init.go
```diff
@@ -145,6 +151,25 @@ func InitFiles(ctx context.Context, initCfg InitConfig) error {
 		comet.P2P.Seeds = strings.Join(seeds, ",")
 	}
 
+	if initCfg.TrustedSync && network.IsProtected() {
+		rpcServer := fmt.Sprintf("https://rpc.consensus.%s.omni.network", network)
```
corverroos, line 155: we need to update this once we remove the `rpc` subdomian.

PR #1001, opened 2024-05-06
feat(monitor/xfeemngr): keep on-chain gas prices above live gas prices

kevinhalliday:
Keep on chain gas prices above live gas prices.

By introducing an on chain "sheild", set to the same pct as thegas price buffer threshold,
we ensure that on chain gas prices are always at least as high as live gas prices.

With this implementation, on chain gas prices may be up to 20% greater than, but never less than, live gas prices.
As opposed to before, in which on chain gas prices could up to 10% greater than or 10% less than live gas prices.

task: https://app.asana.com/0/1206208509925075/1207250196592177

### DIFF
--- a/monitor/xfeemngr/xfeemngr.go
+++ b/monitor/xfeemngr/xfeemngr.go
@@ -27,8 +27,20 @@ type Manager struct {
 	ticker  ticker.Ticker
 }
 
-// feeOracleSyncInterval is the interval at which fee oracles syncs buffered gas and token prices with FeeOracle deployments.
-const feeOracleSyncInterval = 5 * time.Minute
+const (
+	// feeOracleSyncInterval is the interval at which fee oracles syncs buffered gas and token prices with FeeOracle deployments.
+	feeOracleSyncInterval = 5 * time.Minute
+
+	// tokenPriceBufferThreshold is the pct threshold at which a new token price is buffered.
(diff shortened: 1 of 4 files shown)


idea404 (approved):

monitor/xfeemngr/xfeemngr.go
```diff
@@ -27,8 +27,20 @@ type Manager struct {
...
+
+	// tokenPriceBufferThreshold is the pct threshold at which a new token price is buffered.
+	tokenPriceBufferThreshold = 0.1
+
+	// gasPriceBufferThreshold is the pct threshold at which a new gas price is buffered.
+	gasPriceBufferThreshold = 0.1
+
+	// GasPriceShield is the pct offset above the buffered gas price the oracle sets on chain
+	// Setting shield == buffer threshold ensures that on chain gas price is always at least as high as the live gas price.
+	GasPriceShield = 0.1
```
pavelnikolov, line 42: Should we call this `GasPriceBuffer` instead?

pavelnikolov (approved):

corverroos (approved):

monitor/xfeemngr/xfeemngr.go
kevinhalliday: I wish, it would be more appropriate, but we already have something called the gas price buffer

PR #1003, opened 2024-05-06
chore(monitor): monitor portal balances

kevinhalliday:
Monitor portal balance - so we can see fees collected

task: none


monitor/xfeemngr/xfeemngr.go
```diff
@@ -56,7 +56,11 @@ func Start(ctx context.Context, network netconf.Network, endpoints xchain.RPCEnd
 		ticker:  ticker.New(ticker.WithInterval(feeOracleSyncInterval)),
 	}
 
-	go m.start(ctx)
+	if err := startMonitoring(ctx, network, endpoints); err != nil {
+		return errors.Wrap(err, "start monitoring")
+	}
+
+	m.start(ctx)
```
kevinhalliday, line 63: We didn't need the `go` here

pavelnikolov (approved):

PR #1011, opened 2024-05-07
chore(docs): refactor gh codeblock to use react-query

kevinhalliday:
Refactoring codeblocks to use react query.

task: none


docs/site/docusaurus.config.ts
```diff
@@ -26,6 +26,8 @@ const config: Config = {
     locales: ["en"],
   },
 
+  clientModules: ["src/client-modules/index.ts"],
```
idea404, line 29: neat, this does now load before the site

docs/site/src/client-modules/react-query-client.ts
```diff
@@ -0,0 +1,20 @@
...
+// prefectch known codeblock queries
+const knownCodeBlocks = [
+  'https://github.com/omni-network/omni/blob/main/contracts/src/interfaces/IOmniPortal.sol',
+  'https://github.com/omni-network/omni/blob/main/contracts/src/pkg/XApp.sol',
+  'https://github.com/omni-network/omni/blob/main/contracts/src/libraries/XTypes.sol',
+  'https://github.com/omni-network/omni-forge-template/blob/main/src/XGreeter.sol',
+  'https://github.com/omni-network/omni/blob/059303647e07fc3481e379b710922e2b84b1827f/contracts/src/pkg/XApp.sol#L56-L65',
+  'https://github.com/omni-network/omni/blob/059303647e07fc3481e379b710922e2b84b1827f/contracts/src/protocol/OmniPortal.sol#L135-L151',
+  'https://github.com/omni-network/omni/blob/059303647e07fc3481e379b710922e2b84b1827f/contracts/src/protocol/OmniPortal.sol#L165-L220',
+]
```
idea404, line 15: Don't think it's easy to maintain adding a `GitHubCodeBlock` in markdown and then adding that same URL here again. Wondering whether we can auto-populate this based on `.md` + `.mdx` files in `./docs`

kevinhalliday: No it's messy just wanted to show

docs/site/src/client-modules/react-query-client.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const prefech = () =>
+  knownCodeBlocks.forEach(url => client.prefetchQuery(queryOpts(url)))
+
+prefech()
```
idea404, line 20: Cool stuff

docs/site/src/client-modules/react-query-client.ts
kevinhalliday: Also wanted to see if it fixed loading flash on codeblocks, which it does

docs/site/src/components/GitHubCodeBlock/useCodeBlock.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const matchesMain = code === codeOnMain
+
+  return {
+    code,
+    matchesMain,
+    mainURL: sourceURL(onMain(file)),
+    sourceURL: sourceURL(file),
+    language: determineLanguage(file.path),
+  }
+}
```
idea404, line 26: really like how neat this is too

idea404 (approved):

PR #1017, opened 2024-05-08
feat(contracts): track xblock offset in portal

kevinhalliday:
Track xblock offset in portal contracts.

The xblock offest for a portal increments once per each 
source block with at least one xmsg.

task: https://app.asana.com/0/1206208509925075/1207265407492966


e2e/app/xregistry.go
```diff
@@ -123,26 +118,33 @@ func (m registryMngr) registerPortal(ctx context.Context, chainID uint64) error
...
-	m.txOpts.Value = nil
+	txOpts, err := m.backend.BindOpts(ctx, m.admin)
+	if err != nil {
+		return errors.Wrap(err, "bind opts")
+	}
+	txOpts.Value = fee
+
+	if name, _ := m.backend.Chain(); name == "omni_evm" {
+		txOpts.GasLimit = 1_000_000
+	}
```
kevinhalliday, line 129: we need to figure out why this is necessary for omni evm xcall txns

contracts/src/protocol/OmniPortal.sol
```diff
@@ -145,11 +145,54 @@ contract OmniPortal is
...
+        (uint32 _xcallOffset, uint64 _xblockOffset, uint128 _sourceBlockNumber) = _parseLastXCallID();
+
+        if (uint128(block.number) > _sourceBlockNumber) {
+            _sourceBlockNumber = uint128(block.number);
+            _xblockOffset = uint64(_xblockOffset + 1);
+            _xcallOffset = 0;
+        } else {
+            _xcallOffset += 1;
+            _xblockOffset = _xblockOffset;
+            _sourceBlockNumber = uint128(block.number);
```
kevinhalliday, line 157: Necessary to keep gas usage between `else` branch equal to `if` branch.

This is messy I don't love it

kevinhalliday:
Closing - deadend

PR #1033, opened 2024-05-10
fix(explorer/graphql): use mock timestamps in the past

pavelnikolov:
Fix the mock data returned by the GraphQL API to return timestamps in the past. Otherwise the "TimeAgo" logic doesn't work in the Explorer and breaks the UI. 

task: none


CaidynLuckyBeard (approved):
looks good

PR #1041, opened 2024-05-10
feat(cli): rollups in devnet

idea404:
adds `mock_op` and `mock_arb` to `devnet0` and makes cli logs info-only, as well as adding a log line when a portal is private-deployed

task: none

### DIFF
--- a/e2e/manifests/devnet0.toml
+++ b/e2e/manifests/devnet0.toml
@@ -2,6 +2,6 @@
 # This is aimed at local dev environments.
 
 network = "devnet"
-anvil_chains = ["mock_l1", "mock_l2"]
+anvil_chains = ["mock_l1", "mock_op", "mock_arb"]
 
 [node.validator01]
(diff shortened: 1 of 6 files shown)


e2e/manifests/devnet0.toml
```diff
@@ -2,6 +2,6 @@
 # This is aimed at local dev environments.
 
 network = "devnet"
-anvil_chains = ["mock_l1", "mock_l2"]
+anvil_chains = ["mock_l1", "mock_op", "mock_arb"]
```
corverroos, line 5: can remove `mock_l1` I think? Not sure devs need it, and it uses additional resources...

lib/log/logger.go
```diff
@@ -94,6 +94,8 @@ func newCLILogger(opts ...func(*options)) *slog.Logger {
 		opt(&o)
 	}
 
+	o.Level = slog.LevelInfo // Only show info and above
```
corverroos, line 97: Move this up to above forloop, this allows overriding if required

corverroos (approved):

e2e/manifests/devnet0.toml
idea404: agreed. have added a conditional to avs deployment step

PR #1045, opened 2024-05-10
feat(halo/evmengine): set apphash as beaconblockroot

corverroos:
Sets cometBFT `AppHash` as the EVM `BeaconBlockRoot` when building EVM blocks.

This required moving optimistic building from `msg_server` to it's own custom hook called after `abci.finaliseblock` so we have access to the final appHash.

Also includes a PoC rollback function, this will be completed and tested in following PR.

task: none

### DIFF
--- a/halo/app/abci.go
+++ b/halo/app/abci.go
@@ -10,11 +10,21 @@ import (
 	abci "github.com/cometbft/cometbft/abci/types"
 )
 
-type loggingABCIApp struct {
+type finaliseCallback func(context.Context, *abci.RequestFinalizeBlock, *abci.ResponseFinalizeBlock) error
+
+type abciWrapper struct {
 	abci.Application
+	finaliseCallback finaliseCallback
 }
 
-func (l loggingABCIApp) Info(ctx context.Context, info *abci.RequestInfo) (*abci.ResponseInfo, error) {
+func newABCIWrapper(app abci.Application, finaliseCallback finaliseCallback) *abciWrapper {
+	return &abciWrapper{
+		Application:      app,
+		finaliseCallback: finaliseCallback,
+	}
+}
+
+func (l abciWrapper) Info(ctx context.Context, info *abci.RequestInfo) (*abci.ResponseInfo, error) {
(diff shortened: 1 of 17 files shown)


halo/app/abci.go
```diff
@@ -69,16 +79,21 @@ func (l loggingABCIApp) ProcessProposal(ctx context.Context, proposal *abci.Requ
 	return resp, err
 }
 
-func (l loggingABCIApp) FinalizeBlock(ctx context.Context, block *abci.RequestFinalizeBlock) (*abci.ResponseFinalizeBlock, error) {
-	resp, err := l.Application.FinalizeBlock(ctx, block)
+func (l abciWrapper) FinalizeBlock(ctx context.Context, req *abci.RequestFinalizeBlock) (*abci.ResponseFinalizeBlock, error) {
```
pavelnikolov, line 82: Does it make sense to change the variable name `l` -> `a`?
```suggestion
func (a abciWrapper) FinalizeBlock(ctx context.Context, req *abci.RequestFinalizeBlock) (*abci.ResponseFinalizeBlock, error) {
```

halo/evmengine/keeper/abci.go
```diff
@@ -145,7 +147,88 @@ func (k *Keeper) PrepareProposal(ctx sdk.Context, req *abci.RequestPreparePropos
...
+	return nil
+}
+
+func (k *Keeper) startOptimisticBuild(ctx context.Context, appHash common.Hash) (engine.ForkChoiceResponse, error) {
+	latestEBlock, err := k.engineCl.HeaderByNumber(ctx, nil)
+	if err != nil {
+		return engine.ForkChoiceResponse{}, errors.Wrap(err, "get head")
+	}
+
+	ts := uint64(time.Now().Unix())
```
pavelnikolov, line 197: If we use `time.Now()` will this be deterministic across all validators or is this executed only by the block proposer?

pavelnikolov (approved):

halo/evmengine/keeper/abci.go
corverroos: only proposer

PR #1053, opened 2024-05-14
feat(halo): do not attest to empty xblocks

corverroos:
Do not attest to empty blocks. See [ADR](https://www.notion.so/omni-network/Don-t-attest-to-empty-xblocks-09249161404b48358d0581a6204491f2).

task: none


halo/app/start_test.go
```diff
@@ -88,15 +88,15 @@ func TestSmoke(t *testing.T) {
 
 	srcChain := netconf.Simnet.Static().OmniExecutionChainID
 	// Ensure all blocks are attested and approved.
-	cprov.Subscribe(ctx, srcChain, 0, "test", func(ctx context.Context, approved xchain.Attestation) error {
+	cprov.Subscribe(ctx, srcChain, 1, "test", func(ctx context.Context, approved xchain.Attestation) error {
```
pavelnikolov, line 91: Why do we switch from 0 to 1?

explorer/graphql/resolvers/messages_test.go
```diff
@@ -90,7 +90,7 @@ func TestXMsgsNoCursor(t *testing.T) {
 							Cursor
 							Node {
 								ID
-								Offset
+								MsgOffset
```
pavelnikolov, line 93: The prefix doesn't need to be added to the front-end API because the nodes here are of type `XMsg`. The second problem is that it actually introduces a breaking changes for the FE and Explorer. You can leave it and I will rebase and change it in my upcoming PR

e2e/types/manifest.go
```diff
@@ -51,8 +51,10 @@ const (
 type Perturb string
 
 const (
+	// PerturbRestart defines a perturbation that restarts a docker container.
+	PerturbRestart Perturb = "restart"
 	// PerturbStopStart defines a perturbation that stops and then starts a docker container.
-	PerturbStopStart = "stopstart"
+	PerturbStopStart Perturb = "stopstart"
```
pavelnikolov, line 57: What is the difference between the above two? To my understanding restart will ALWAYS use the same container and start/stop will potentially use a new container if available, right?
On a side note, what happens during hard VM reset (the process inside of the container doesn't receive a proper signal and a chance to perform a graceful shutdown)?

pavelnikolov (approved):

lib/xchain/provider/fetch.go
```diff
@@ -196,7 +214,12 @@ func (p *Provider) GetBlock(ctx context.Context, chainID uint64, height uint64)
 		Msgs:      msgs,
 		Receipts:  receipts,
 		Timestamp: time.Unix(int64(header.Time), 0),
-	}, true, nil
+	}
+	if resp.ShouldAttest() {
+		resp.BlockOffset = xOffset
```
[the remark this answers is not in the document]
kevinhalliday: Yeah that `blockoffseter` below

relayer/app/state.go
```diff
@@ -43,16 +43,16 @@ func (s *State) Clear(dstID uint64) error {
...
+func (s *State) Persist(dstID, srcID, offset uint64) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
 	srcMap, ok := s.cursors[dstID]
 	if !ok {
 		srcMap = make(map[uint64]uint64)
 	}
-	srcMap[srcID] = height
+	srcMap[srcID] = offset
```
kevinhalliday, line 55: How do we map offest to source block height (which is we're we need to start looking) - ask halo?

kevinhalliday (approved):

halo/app/start_test.go
corverroos: We previously attested to all xblocks, including genesis. And we used block height as the cursor (height==0).
Now, we only attest to non-empty blocks. And we use block offset as cursors, and they start at 1 (offset=1).

explorer/graphql/resolvers/messages_test.go
corverroos: sorry typo

e2e/types/manifest.go
corverroos: See `e2e/app/perturb.go#perturbService`

lib/xchain/provider/fetch.go
```diff
@@ -93,16 +93,25 @@ func (p *Provider) GetSubmittedCursor(ctx context.Context, destChainID uint64, s
...
-	offset, err := caller.InXStreamOffset(&bind.CallOpts{Context: ctx}, sourceChainID)
+	height, err := rpcClient.BlockNumber(ctx)
+	if err != nil {
+		return xchain.StreamCursor{}, false, err
+	}
+
+	callOpts := &bind.CallOpts{Context: ctx, BlockNumber: big.NewInt(int64(height))}
+
+	// TODO(corver): Rename portal variable to InXStreamMsgOffset
+	msgOffset, err := caller.InXStreamOffset(callOpts, sourceChainID)
```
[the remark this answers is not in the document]
corverroos: nice, agreed

relayer/app/state.go
corverroos: yeah, we query `latestAttestation` which has a `height` and `offset`, we start streaming from that point.

e2e/types/manifest.go
corverroos: We need to add some tests for `PurturbKill`

PR #1054, opened 2024-05-14
feat(contracts): allow operators to register independent val keys

kevinhalliday:
Allow operators to register validator pub keys that are 
separate from their operator keys. 

task: https://app.asana.com/0/1206208509925075/1207231357626829


contracts/src/protocol/OmniAVS.sol
```diff
@@ -127,6 +163,29 @@ contract OmniAVS is IOmniAVS, IOmniAVSAdmin, OwnableUpgradeable, PausableUpgrade
...
     }
 
+    /**
+     * @notice Returns the digest hash to be signed by an opeator's validator key on registration.
+     * @param operator      The operator's ethereum address
+     * @param valPubKey     The operator's 64 byte uncompressed secp256k1 validator public key
+     * @param salt          A salt unique to this registration
+     * @param expiry        The timestamp at which this registration expires
+     */
+    function validatorRegistrationDigestHash(address operator, bytes calldata valPubKey, bytes32 salt, uint256 expiry)
```
kevinhalliday, line 173: Went with [EIP-712](https://eips.ethereum.org/EIPS/eip-712) compatible signatures, in the same way eigen's [AVSDirectory](https://github.com/Layr-Labs/eigenlayer-contracts/blob/dev/src/contracts/core/AVSDirectory.sol#L150) does. We may not need "readablity" off chain. But best to follow standards if they're there

contracts/src/protocol/OmniAVS.sol
```diff
@@ -88,26 +107,43 @@ contract OmniAVS is IOmniAVS, IOmniAVSAdmin, OwnableUpgradeable, PausableUpgrade
...
         require(!allowlistEnabled || _allowlist[operator], "OmniAVS: not allowed");
         require(!_isOperator(operator), "OmniAVS: already an operator");
         require(_operators.length < maxOperatorCount, "OmniAVS: max operators reached");
         require(_getTotalDelegations(operator) >= minOperatorStake, "OmniAVS: min stake not met");
 
-        _addOperator(operator, pubkey);
-        _avsDirectory.registerOperatorToAVS(operator, operatorSignature);
+        // verify the the validator signature is not expired or spent
+        require(valSig.expiry >= block.timestamp, "OmniAVS: signature expired");
+        require(!_isValActive[valPubKey], "OmniAVS: pubkey already active");
```
kevinhalliday, line 131: We can enforce this here. But we cannot enforce this across OmniAVS & OmniStake

ttarsi (approved):

PR #1060, opened 2024-05-16
feat(contracts): allow omni stakers to registe indepdent val key

kevinhalliday:
Allow OmniStakers to specify a validator key that is independent from their 
delegator key.

task: https://app.asana.com/0/1206208509925075/1206208765203004


kevinhalliday:
Closing this - will revist with omnistake <> avs syncing

PR #1069, opened 2024-05-17
feat(cli): simulated fee params for devnet when no network is available

idea404:
simulates `feeparams` when starting a local devnet without network connectivity

so folks can code in the plane

task: none


corverroos (changes requested):
This has already been fixed, see `destFeeParams` in main.

PR #1071, opened 2024-05-20
fix(explorer/graphql): allow for null receipt properties

pavelnikolov:
Fixes an edge cases where the receipt doesn't exist yet and we try to view an XMsg.

task: none

### DIFF
--- a/explorer/graphql/data/message_provider.go
+++ b/explorer/graphql/data/message_provider.go
@@ -100,7 +100,7 @@ type dbXMsgRow struct {
 	ReceiptID             sql.NullInt64
 	ReceiptTxHash         *common.Hash
 	ReceiptSuccess        sql.NullBool
-	ReceiptRelayerAddress common.Address
+	ReceiptRelayerAddress *common.Address
 	ReceiptGasUsed        sql.NullInt64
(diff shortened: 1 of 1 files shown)


explorer/graphql/data/message_provider.go
```diff
@@ -100,7 +100,7 @@ type dbXMsgRow struct {
 	ReceiptID             sql.NullInt64
 	ReceiptTxHash         *common.Hash
 	ReceiptSuccess        sql.NullBool
-	ReceiptRelayerAddress common.Address
+	ReceiptRelayerAddress *common.Address
```
pavelnikolov, line 103: All other properties of the receipts are `sqlNull*` or a pointer. This would fail with an sql driver error:
```
scanning row: sql: Scan error on column index 15, name \"receipt_relayer_address\": can't scan <nil> into Address
```
if the receipt doesn't exist (i.e. for pending XMsgs).

pavelnikolov: I'll add it to the e2e tests in a followup PR.

corverroos (approved):
Some unit tests would be awesome 🙏

PR #1077, opened 2024-05-21
chore(halo/evmengine): support failed evm log delivery

corverroos:
Support evm log failed delivery by using a cached multi store.

Also catch EVM log panics (possibly caused by `k1util.PubkeyFromBytes64`).

task: none


contracts/src/protocol/OmniStake.sol
```diff
@@ -23,7 +23,8 @@ contract OmniStake {
     function deposit(bytes memory pubkey) external payable {
         require(msg.value >= 1 ether, "OmniStake: deposit amt too low");
         require(msg.value < type(uint64).max, "OmniStake: deposit amt too high");
-        require(msg.sender == Secp256k1.pubkeyToAddress(pubkey), "OmniStake: pubkey not sender");
+        // TODO(corver): Figure out another way to test failed evm log in cosmos.
```
corverroos, line 26: this contract is going to be refactors in the coming weeks. This is just a temporary workaround in order to test failed EVM logs. The refactored contract will also support failed testing.

contracts/test/stake/OmniStake.t.sol
```diff
@@ -33,33 +33,33 @@ contract OmniStake_Test is Test {
...
+    //        Vm.Wallet memory wallet = vm.createWallet("test val");
+    //        bytes memory pubkey = bytes.concat(hex"04", _pubkey(wallet));
+    //
+    //        uint256 amt = 16 ether;
+    //        vm.deal(wallet.addr, amt);
+    //
+    //        vm.expectRevert("Secp256k1: invalid pubkey length");
+    //        vm.prank(wallet.addr);
+    //        stake.deposit{ value: amt }(pubkey);
+    //    }
```
pavelnikolov, line 62: We should figure out [another way to ignore these tests](https://ethereum.stackexchange.com/questions/141643/foundry-skip-specifics-tests) without commenting out. For example start the test name with `skip_` or add them to no_match pattern.

pavelnikolov (approved):

idea404 (approved):

PR #1080, opened 2024-05-21
feat(cli/docs): introduce devnet quickstart guide

idea404:
Introduces scaffolding `hello-world-template` to CLI as default (and using `--template` flag) and the quickstart page to the docs

task: https://app.asana.com/0/1206208509925075/1207372351528365/f


cli/cmd/develop.go
```diff
@@ -50,8 +59,24 @@ func newForgeProjectTemplate() error {
 		}
 	}
 
+	// Map the template name to the forge template URL.
+	templateURLs := map[string]string{
+		"hello-world": "https://github.com/omni-network/hello-world-template.git",
+		"xgreeter":    "https://github.com/omni-network/omni-forge-template.git",
```
kevinhalliday, line 65: Rather than mapping, we should just allow users to pass in the `template`. That was we can add more without upgrading cli.

If we want better names, we can rename the gitrepos that host them

docs/site/docs/develop/quickstart/quickstart.md
```diff
@@ -0,0 +1,119 @@
...
+omni devnet info
+```
+
+And you the private key value is the first anvil private key, found by running:
+
+```bash
+anvil
+```
+
+These values are kept in `./script/bash/env.sh` and are used to deploy the contracts.
```
kevinhalliday, line 78: Maybe call this `.env`, that's more standard

docs/site/docs/develop/quickstart/quickstart.md
```diff
@@ -0,0 +1,119 @@
...
+<summary>Obtaining Deployment Addresses</summary>
+
+You can obtain the XGreeter deployment addresses from the output of the previous forge script deployment.
+
+Because the devnet has just been started, the addresses will be the same as the ones shown below:
+
+```bash
+omni_evm: 0x0DCd1Bf9A1b36cE34237eEaFef220932846BCD82
+mock_op: 0xa513E6E4b8f2a923D98304ec87F64353C4D5C853
+mock_arb: 0xa513E6E4b8f2a923D98304ec87F64353C4D5C853
```
kevinhalliday, line 100: Maybe worth getting these to be same across each chain - what are we doing differently on omnievm

kevinhalliday: At some point we could also show devs how to use our create3 factories here

pavelnikolov: What would be the usecase?

idea404: I'll use a different anvil key as the default, this is nonce+predeploy related

kevinhalliday: Use case would be so devs can get consistent addrs across chain - using create3 would show them how they could

cli/cmd/flags.go
```diff
@@ -24,6 +24,10 @@ func bindAVSAddress(cmd *cobra.Command, addr *string) {
 	cmd.Flags().StringVar(addr, "avs-address", *addr, "Optional address of the Omni AVS contract")
 }
 
+func bindDeveloperForgeProjectConfig(cmd *cobra.Command, cfg *developerForgeProjectConfig) {
+	cmd.Flags().StringVar(&cfg.templateURL, "template", "https://github.com/omni-network/hello-world-template.git", "URL of the forge template to use")
```
kevinhalliday, line 28: This works - that way they can use any template they like. But we could also just use the name of it, and try to fetch `https://github.com/omni-network/{template}.git`

cli/cmd/develop.go
```diff
@@ -70,3 +87,21 @@ func checkForgeInstalled() bool {
 
 	return err == nil // If there is no error, forge is installed.
 }
+
+// sanitizeInput validates and sanitizes the input to prevent command injection.
+func sanitizeInput(input string) (string, error) {
```
kevinhalliday, line 92: Why do we need to do this? They're running this on their own machine. 

Also if we are askibng for urls shouldnt' ths allow `:`

kevinhalliday (approved):

cli/cmd/develop.go
idea404: true that was me fighting the pre-commit runner without questioning it

PR #1082, opened 2024-05-21
feat(contracts): fix xsubmit provided gas check

kevinhalliday:
Fix check that ensures relayer has provided sufficient gas for all xmsgs.

task: https://app.asana.com/0/1206208509925075/1207372347229489


pavelnikolov (approved):

contracts/src/protocol/OmniPortal.sol
```diff
@@ -280,22 +280,28 @@ contract OmniPortal is
...
+        if (gasLeftAfter <= gasLimit / 63) {
+            // We use a revert with message to improve testability, to differentiate this from
+            // an actual OutOfGas exception.
+            //
+            // We could use invalid opcode to consume all gas and bubble-up the effects, since
+            // and emulate an "OutOfGas" exception
+            // i.e. assembly { invalid() }
+            //
+            // We instead allow our relayer to respond to the message directly
+            revert("OmniPortal: insufficient gas");
```
kevinhalliday, line 301: Could use feedback here. This was helpful to differentiate between this "out of gas" error vs a normal out of gas exception in tests. Using `invalid()` would emulate an "OutOfGas" exception, but I'm not sure we need that. We can handle specific error handling for this revert reason in relayer if we need it

corverroos: Getting a revert reason in golang is tricky. And I would prefer if we don't need specific error handling logic.

corverroos (approved):

contracts/src/protocol/OmniPortal.sol
kevinhalliday: ah yes indeed

contracts/src/protocol/OmniPortal.sol
```diff
@@ -282,22 +282,24 @@ contract OmniPortal is
...
         // solhint-disable-next-line avoid-low-level-calls
         (bool success, bytes memory result) = to.call{ gas: gasLimit }(data);
 
-        gasUsed = gasUsed - gasleft();
+        uint256 gasLeftAfter = gasleft();
 
-        return (success, result, gasUsed);
+        // Esnure relayer sent enough gas for the call
+        // See https://github.com/OpenZeppelin/openzeppelin-contracts/blob/bd325d56b4c62c9c5c1aff048c37c6bb18ac0290/contracts/metatx/MinimalForwarder.sol#L58-L68
+        if (gasLeftAfter <= gasLimit / 63) {
```
arajasek, line 294: This is correct, but why not just check `63 * gasLeftBefore >= 64 * gasLimit`? We're already loading computing `gasLeftBefore` anyway, so I don't see any extra work.

- The math seems simpler / cleaner (no division with remainders being dropped)
- We can make the check _before_ the call, and so fail early if we want to.

PR #1109, opened 2024-05-23
fix(explorer): fix explorer bugs

pavelnikolov:
* properly handle nil receipts when displaying txHash and receipt timestamp
* move block timestamp to source chain and not dest chain to fix confusion

task: none

### DIFF
--- a/explorer/ui/app/components/details/From.tsx
+++ b/explorer/ui/app/components/details/From.tsx
@@ -1,6 +1,7 @@
 import { Link } from '@remix-run/react'
 import React from 'react'
-import { hashShortener } from '~/lib/formatting'
+import { dateFormatter, hashShortener } from '~/lib/formatting'
+import { dateFormatterXMsgPage } from './date.service'
 import { copyToClipboard } from '~/lib/utils'
 
 export const From = ({ xMsgDetails }) => {
@@ -65,7 +66,18 @@ export const From = ({ xMsgDetails }) => {
           onClick={() => copyToClipboard(xMsgDetails?.txHash)}
         />
       </div>
-      {/* Offset */}
+      {/* Block Timestamp */}
+      <div className="flex mt-5 pb-2 border-b-[1px] border-subtle border-solid">
+        <p className="w-[150px] sm:w-48 text-sm">Block Timestamp</p>
+        <p className="flex-1">
+          <span className="text-default">
+            {dateFormatter(new Date(xMsgDetails?.block.timestamp))}{' '}
+          </span>
+         ( {xMsgDetails?.block.timestamp &&
+            dateFormatterXMsgPage(new Date(xMsgDetails?.block.timestamp))})
+        </p>
+      </div>
+      {/* Block height */}
       <div className="flex mt-5 pb-2 border-b-[1px] border-subtle border-solid">
         <p className="w-[150px] sm:w-48 text-sm">Block Height</p>
         <p className="text-default">{parseInt(xMsgDetails?.block.height)}</p>
--- a/explorer/ui/app/components/details/From.tsx
+++ b/explorer/ui/app/components/details/From.tsx
@@ -4,7 +4,7 @@ import { dateFormatter, hashShortener } from '~/lib/formatting'
 import { dateFormatterXMsgPage } from './date.service'
 import { copyToClipboard } from '~/lib/utils'
 
-export const From = ({ xMsgDetails }) => {
+export const From = ({ xmsg }) => {
   return (
     <>
       <h6 className="text-default my-5 text-lg">From</h6>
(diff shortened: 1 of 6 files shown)


explorer/ui/app/components/details/From.tsx
```diff
@@ -65,7 +66,18 @@ export const From = ({ xMsgDetails }) => {
...
+        <p className="w-[150px] sm:w-48 text-sm">Block Timestamp</p>
+        <p className="flex-1">
+          <span className="text-default">
+            {dateFormatter(new Date(xMsgDetails?.block.timestamp))}{' '}
+          </span>
+         ( {xMsgDetails?.block.timestamp &&
+            dateFormatterXMsgPage(new Date(xMsgDetails?.block.timestamp))})
+        </p>
+      </div>
+      {/* Block height */}
```
pavelnikolov, line 80: Move block timestamp to source chain to be next to the other block properties.

explorer/ui/app/components/details/From.tsx
```diff
@@ -65,7 +66,18 @@ export const From = ({ xMsgDetails }) => {
...
-      {/* Offset */}
+      {/* Block Timestamp */}
+      <div className="flex mt-5 pb-2 border-b-[1px] border-subtle border-solid">
+        <p className="w-[150px] sm:w-48 text-sm">Block Timestamp</p>
+        <p className="flex-1">
+          <span className="text-default">
+            {dateFormatter(new Date(xMsgDetails?.block.timestamp))}{' '}
+          </span>
+         ( {xMsgDetails?.block.timestamp &&
+            dateFormatterXMsgPage(new Date(xMsgDetails?.block.timestamp))})
```
kevinhalliday, line 77: Once you check for undefefined / null, you dont' have to check again

explorer/ui/app/components/details/TabList.tsx
```diff
@@ -12,13 +12,17 @@ export const TabList = ({ xMsgDetails }) => {
...
+          <>
+            <span className="text-default">
+              {dateFormatter(new Date(xMsgDetails?.receipt?.timestamp))}{' '}
+            </span>
+            (
+            {xMsgDetails?.receipt?.timestamp &&
+              dateFormatterXMsgPage(new Date(xMsgDetails?.receipt?.timestamp))}
+            )
+          </>
+          ) : '--'}
```
pavelnikolov, line 25: If receipt doesn't exist don't display anything.

explorer/ui/app/components/details/TabList.tsx
```diff
@@ -55,22 +59,18 @@ export const TabList = ({ xMsgDetails }) => {
...
-      {/* Offset */}
       <div className="flex mt-5 pb-2 border-b-[1px] border-subtle border-solid">
         <p className="w-[150px] sm:w-48 text-sm">Stream Offset</p>
         <p className="text-default">{parseInt(xMsgDetails?.offset)}</p>
       </div>
-      {/* Offset */}
       <div className="flex mt-5 pb-2 border-b-[1px] border-subtle border-solid">
         <p className="w-[150px] sm:w-48 text-sm">Gas Used</p>
-        <p className="text-default">{parseInt(xMsgDetails?.receipt?.gasUsed)}</p>
+        <p className="text-default">{xMsgDetails?.receipt ? parseInt(xMsgDetails?.receipt?.gasUsed) : '--'}</p>
```
pavelnikolov, line 68: Don't show `NaN` if receipt doesn't exist. Instead show `--` as we do with other fields in this case.

explorer/ui/app/components/details/From.tsx
kevinhalliday:
```suggestion
         ( {xMsgDetails?.block.timestamp &&
            dateFormatterXMsgPage(new Date(xMsgDetails.block.timestamp))})
```

kevinhalliday: Also nit but I tend to[ prefer ternaries over &&](https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx)
```suggestion
         ( {xMsgDetails?.block.timestamp ? 
            dateFormatterXMsgPage(new Date(xMsgDetails?.block.timestamp)) : null})
```

kevinhalliday (approved):

explorer/ui/app/components/details/TabList.tsx
kevinhalliday: We'll, we do display `--` : could say something like "No Receipt", or actually nothing. Ask Fab

pavelnikolov: This is what we do for tx_hash already for receipts. I copied that behavior. Happy to change it if required.

PR #1127, opened 2024-05-28
fix(e2e): use broadcast omni evm for internal network rpcs

kevinhalliday:
Use BroadcastOmniEVM for internal rcp config.

This prefers omni evms that are up and running at start.

task: none


e2e/app/definition.go
```diff
@@ -470,7 +470,7 @@ func internalEndpoints(def Definition, nodePrefix string) xchain.RPCEndpoints {
 		return endpoints
 	}
 
-	omniEVM := omniEVMByPrefix(def.Testnet, nodePrefix)
+	omniEVM := def.Testnet.BroadcastOmniEVM()
```
corverroos, line 473: lets rather change the following:
```
// omniEVMByPrefix returns a omniEVM from the testnet with the given prefix.
// Or a random omniEVM if prefix is empty.
// Or the only omniEVM if there is only one.
func omniEVMByPrefix(testnet types.Testnet, prefix string) types.OmniEVM {
```
to
```
// Or a broadcast omniEVM if prefix is empty.
```

corverroos: `internalEndpoints` is called in `Setup` and there we need to link the omniEVM to the same halo node directly, even if not available.

corverroos (approved):

PR #1133, opened 2024-05-29
fix(contracts): prevent old vals from exploiting new src chains

kevinhalliday:
Prevent old validator sets from creating xsubmissions for
 newly supported or unsupported source chains.
 
Additionally, require xmsg.sourceChainId matches xsub.blockHeader.sourceChainId

task: https://app.asana.com/0/1206208509925075/1207428329960053


contracts/src/xchain/XRegistry.sol
```diff
@@ -25,7 +25,7 @@ contract XRegistry is Ownable, XRegistryBase {
     /**
      * @notice OmniPortal contract.
      */
-    IOmniPortal public omni;
+    OmniPortal public omni;
```
ttarsi, line 28: why chang from an interface?

contracts/src/xchain/XRegistryBase.sol
```diff
@@ -42,4 +45,8 @@ contract XRegistryBase {
     function _pack(string memory name, address registrant) internal pure returns (bytes32) {
         return keccak256(abi.encodePacked(name, registrant));
     }
+
+    function _isPortal(string memory name, address registrant) internal pure returns (bool) {
```
ttarsi, line 49: can you add a natspec comment here?

ttarsi (approved):

contracts/src/xchain/XRegistry.sol
kevinhalliday: Need `OmniPortal.initSourceChain`, which I didn't think we needed to add to "public facing" omni portal interface

PR #1143, opened 2024-05-29
chore(contracts): store array length in local vars

rokso:
Store array length in local variable in solidity contracts to save gas.

task: none

### DIFF
--- a/contracts/src/xchain/OmniPortal.sol
+++ b/contracts/src/xchain/OmniPortal.sol
@@ -188,7 +188,8 @@ contract OmniPortal is
      *              and a block header and message batch, proven against the attestation root.
      */
     function xsubmit(XTypes.Submission calldata xsub) external whenNotPaused nonReentrant {
-        require(xsub.msgs.length > 0, "OmniPortal: no xmsgs");
+        uint256 numMsgs = xsub.msgs.length;
+        require(numMsgs > 0, "OmniPortal: no xmsgs");
 
         // validator set id for this submission
         uint64 valSetId = xsub.validatorSetId;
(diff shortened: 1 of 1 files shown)


rokso:
Resolved conflicts.

pavelnikolov (approved):

pavelnikolov:
@kevinhalliday LGTM, but I'm not going to run the workflow - will leave that to you.

kevinhalliday (changes requested):
Thanks for the contribution. 

IMO this is a small sacrifice in readablity for smaller savings in gas costs. Generally, I think we should prefer readability, particularly at this stage in development. 

But `xsubmit` is a fine place to save some gas. So let's do it.

Please rename `_length` variable in each place to something more descriptive, so when we see it used later we know what it means.

For example, looking at this loop on it's own at the bottom of the function, it's not immediately clear what we are looping over.

```
  for (uint256 i = 0; i < _length; i++) {
      // ...
  }
```

rokso:
@kevinhalliday I have renamed variable.

kevinhalliday (commented):
Shorten to `numMsgs` and `numVals`?

Our pattern is no underscore prefix needed for local variables

rokso:
> Shorten to `numMsgs` and `numVals`?
> 
> Our pattern is no underscore prefix needed for local variables

Sure I can do that.

rokso:
> Shorten to `numMsgs` and `numVals`?
> 
> Our pattern is no underscore prefix needed for local variables

@kevinhalliday Done

kevinhalliday (approved):

PR #1150, opened 2024-05-30
feat(halo/attest): override fuzzy attestations with finalized

corverroos:
Mark fuzzy attestations as overriden by approval finalized attesattions of same height.

This ensures that we have a sequential stream of "approved" attesations for fuzzy chain versions even though we might not come to consensus on fuzzy chain versions. Finalized attestations therefore override and fill any gaps in fuzzy attestations streams. 

task: none


kevinhalliday (approved):
👏

PR #1164, opened 2024-06-03
feat(contracts): support conf level shards in portal

kevinhalliday:
Support conf level sharding in portal emissions and submissions.

Remaining:

- allow finalized shard to override latest shard
- restrict xcalls to shards supported for that chain
- tests xcalls to different shards in tests

task: https://app.asana.com/0/1206208509925075/1207470304337925


contracts/bindings/portalregistry_storage_layout.json
```diff
@@ -23,49 +23,49 @@
...
+          "astId": 52625,
           "contract": "src/xchain/PortalRegistry.sol:PortalRegistry",
           "label": "chainId",
           "offset": 0,
           "slot": "0",
           "type": "t_uint64"
         },
         {
-          "astId": 52842,
+          "astId": 52627,
```
kevinhalliday, line 52: Gotta figure out a away to get rid of these or stop this from changing

corverroos (approved):

PR #1166, opened 2024-06-03
feat(contracts): allow finalized block to correct latest shards

kevinhalliday:
Allow finalized block to correct latest shard in portal.

To test, use latest conf level in ping pong.

Will clean up xcall interfact w.r.t. conf level & gas limit separately.

task: https://app.asana.com/0/1206208509925075/1207470304337925


contracts/src/xchain/OmniPortal.sol
```diff
@@ -292,8 +294,18 @@ contract OmniPortal is
...
+                require(ConfLevel.Finalized == xsub.blockHeader.confLevel, "OmniPortal: wrong conf level");
+            } else {
+                // else, require block header either matches xmsg conf level, or is finalized
+                // this allows "fuzzy" conf levels to be corrected by finalized blocks
+                require(
+                    xsub.blockHeader.confLevel == uint8(xmsg_.shardId)
+                        || xsub.blockHeader.confLevel == ConfLevel.Finalized,
+                    "OmniPortal: wrong conf level"
+                );
+            }
```
kevinhalliday, line 308: This is core logic change

corverroos (approved):

contracts/src/xchain/OmniPortal.sol
pavelnikolov: Why do we need the `else` clause in the above statement? If the `xmsg._shardId != ConfLevel.Finalized` then any block header conf level is allowed, right? Is is added in case we support more than two confirmation levels in future?

PR #1174, opened 2024-06-05
chore(contracts): cleanup portal logic

kevinhalliday:
Clean up portal logic in the name of readability

One functional fix:
- update xblock offset per xmsg shard, not just per block header conf level

task: none


idea404 (approved):
There's also

pavelnikolov (approved):

PR #1180, opened 2024-06-05
feat(contracts): only expose a single xcall method on portal

kevinhalliday:
Expose a single xcall method on portal with all params, leave defaults to XApp. 

- Allow specification of default conf level in XApp
- Remove default gas limit for xmsgs

task: none


contracts/src/avs/OmniAVS.sol
```diff
@@ -138,6 +139,7 @@ contract OmniAVS is IOmniAVS, IOmniAVSAdmin, OwnableUpgradeable, PausableUpgrade
         Operator[] memory ops = _getOperators();
         omni.xcall{ value: msg.value }(
             omniChainId,
+            ConfLevel.Finalized,
```
kevinhalliday, line 142: Note AVS contract changes here, which differs from version live on Mainnet. We will be upgrading the AVS regardless

contracts/src/pkg/XApp.sol
```diff
@@ -28,8 +45,10 @@ contract XApp {
         delete xmsg;
     }
 
-    constructor(address _omni) {
-        omni = IOmniPortal(_omni);
+    // TODO: write XAppUpgradeable that follows initialize pattern
+    constructor(address omni_, uint8 defaultConfLevel_) {
+        _setOmniPortal(omni_);
+        _setDefaultConfLevel(defaultConfLevel_);
     }
```
kevinhalliday, line 52: Adding state to XApp will require an Upgradeable option. Or just to not have a contructor and leave `_setOmniPortal` and `_setDefaultConfLevel` to users.

Additionally, I changed omni from immutable to storage, because we should leave the option for xapps to change their portal address, to allow us the option to deploy now portals and migrate xapps over.

contracts/src/pkg/XApp.sol
```diff
@@ -3,16 +3,33 @@ pragma solidity ^0.8.12;
...
     /**
      * @notice The OmniPortal contract
      */
-    IOmniPortal internal immutable omni;
+    IOmniPortal public omni;
+
+    /**
+     * @notice Default confirmation level for xcalls
+     */
+    uint8 public defaultConfLevel;
```
idea404, line 32: What do you think of setting a default value here already and providing a value in constructor if you desire to override? Have this be the finalized level by default

idea404 (approved):

idea404:
LGTM, makes sense

contracts/src/pkg/XApp.sol
kevinhalliday: Well I think it needs to be state. Adding another contract to handle something like storing the portal addr or default conf level of xapp deployed separately would make things more complicated

contracts/src/pkg/XApp.sol
kevinhalliday: We can't overload the constructor so the ux for default is awkward. And I think it's actually a reasonable thing to make people think about, at least once

PR #1187, opened 2024-06-06
fix(e2e): temp mac chip fix for explorer images

idea404:
introduces temp fix of platform assignment to `linux/amd64` for all explorer-related images for runs in mac chip machines

task: none


corverroos:
Shouldn't we just build multi-arch images, like we do for go images in goreleaser?

idea404:
> Shouldn't we just build multi-arch images, like we do for go images in goreleaser?

Let's do this instead

PR #1193, opened 2024-06-06
feat(api): add token supply routes

kevinhalliday:
Adds a token supply api, to be deployed with vercel.

task: https://app.asana.com/0/1206208509925075/1207497924242030


ttarsi (changes requested):
discussed async

ttarsi (approved):

PR #1196, opened 2024-06-07
refactor(relayer): remove local cursor state

corverroos:
Remove relayer local state: `relayer-state.json`

This was used to bootstrap relayer on startup to skip a range of empty blocks. But since we don't attest to empty blocks, this is required anymore. It also conflicts with supporting gaps in fuzzy streams. 

task: none


relayer/app/cursors.go
```diff
@@ -82,21 +82,14 @@ func filterMsgs(ctx context.Context, streamID xchain.StreamID, valSetID uint64,
...
 	res := make(map[xchain.ChainVersion]uint64)
 
 	// Initialize all chain versions to start at 1 by default or if local state is present
 	for _, chainVer := range chainVers {
 		res[chainVer] = initialXBlockOffset
-
-		// If local persisted state is higher, use that instead, skipping a bunch of empty blocks on startup.
-		if offset := state.GetOffset(destChainID, chainVer); offset > initialXBlockOffset {
-			res[chainVer] = offset
-		}
```
corverroos, line 99: this use-case isn't required anymore

PR #1198, opened 2024-06-07
refactor(halo/attest): add block stream offsets to attest module

corverroos:
Adds `BlockStreamOffsets` to `BlockHeader` and adds that to `attest` module state.

Had to formalize a `UniqueKey` hash to dedup attestations and votes.

task: none


pavelnikolov (approved):

corverroos:
We have decided to pivot from the solution in favour of off-chain rehydrated state since this introduces a big performance impact on consensus chain with significant increase in attestation load.

PR #1203, opened 2024-06-11
test(e2e): test native token bridge

kevinhalliday:
Test L1 to Native OMNI token bridge in e2e.

task: https://app.asana.com/0/1206208509925075/1207470307350060


lib/contracts/portal/deploy.go
```diff
@@ -36,6 +36,10 @@ const (
 	XReceiptMaxErrorBytes = uint16(256)
 )
 
+func isDeadOrEmpty(addr common.Address) bool {
+	return addr == common.Address{} || addr == common.HexToAddress(eoa.ZeroXDead)
+}
+
```
kevinhalliday, line 42: Saw we use `0x000...dead` in `e2e/app/eoa` so added these checks in a few places to make sure we don't set owner to dead or something like that

corverroos (approved):

PR #1208, opened 2024-06-11
chore(cli): updates scaffolding to use new template

idea404:
updates scaffolding to use git commands instead of forge init command and specifies commit hash of the template repository for compatibility

### DIFF
--- a/cli/cmd/develop.go
+++ b/cli/cmd/develop.go
@@ -68,7 +68,7 @@ func newForgeProjectTemplate(ctx context.Context, cfg developerForgeProjectConfi
 
 	// Check out the specific commit if using the default template
 	if cfg.templateName == defaultTemplate {
-		cmd = exec.CommandContext(ctx, "git", "checkout", "f25e418") // Commit hash of the template, update as needed
+		cmd = exec.CommandContext(ctx, "git", "checkout", "1d0ba3c") // Commit hash of the template, update as needed
 		cmd.Dir = destinationPath
 		cmd.Stdout = nil // Suppress output
 		cmd.Stderr = nil // Suppress output
--- a/cli/cmd/develop.go
+++ b/cli/cmd/develop.go
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	defaultTemplate 	  = "hello-world-template"
+	defaultTemplate       = "hello-world-template"
 	defaultTemplateCommit = "1d0ba3c" // Commit hash of the template, update as needed
 )
 
--- a/cli/cmd/develop.go
+++ b/cli/cmd/develop.go
@@ -1,14 +1,18 @@
 package cmd
 
 import (
-	"bytes"
 	"context"
 	"os"
 	"os/exec"
+	"path/filepath"
 
 	"github.com/omni-network/omni/lib/errors"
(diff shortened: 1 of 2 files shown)


cli/cmd/develop.go
```diff
@@ -47,35 +51,55 @@ type developerForgeProjectConfig struct {
...
+	cmd := exec.CommandContext(ctx, "git", "clone", "https://github.com/omni-network/"+cfg.templateName+".git", destinationPath)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return errors.Wrap(err, "failed to clone repository")
+	}
+
+	// Check out the specific commit if using the default template
+	if cfg.templateName == defaultTemplate {
+		cmd = exec.CommandContext(ctx, "git", "checkout", "1d0ba3c") // Commit hash of the template, update as needed
```
kevinhalliday, line 71: Maybe move this & it's comment to a constant, so it's clear it's a thing

cli/cmd/develop.go
```diff
@@ -47,35 +51,55 @@ type developerForgeProjectConfig struct {
...
+		if err := cmd.Run(); err != nil {
+			return errors.Wrap(err, "failed to checkout commit")
 		}
 	}
 
-	// Execute the `forge init` command.
-	// #nosec G204
-	cmd := exec.Command("forge", "init", "--template", "https://github.com/omni-network/"+cfg.templateName+".git")
+	// Initialize submodules
+	cmd = exec.CommandContext(ctx, "git", "submodule", "update", "--init", "--recursive")
```
kevinhalliday, line 81: `forge` install`? same thing basically. though with git you don't need forge installed which is nice

cli/cmd/develop.go
```diff
@@ -47,35 +51,55 @@ type developerForgeProjectConfig struct {
...
-			Suggest: "You can install foundry by visiting https://github.com/foundry-rs/foundry",
+// newForgeProjectTemplate creates a new project by cloning a repository, checking out a specific commit if necessary, and initializing a new Git repository.
+func newForgeProjectTemplate(ctx context.Context, cfg developerForgeProjectConfig) error {
+	destinationPath, err := os.Getwd()
+	if err != nil {
+		return errors.Wrap(err, "failed to get current working directory")
+	}
+
+	// Clone the repository
+	cmd := exec.CommandContext(ctx, "git", "clone", "https://github.com/omni-network/"+cfg.templateName+".git", destinationPath)
```
kevinhalliday, line 62: Since we are removing `.git` and need no history, do `--depth 1` in clone

pavelnikolov (approved):

cli/cmd/develop.go
idea404: Good idea did this

cli/cmd/develop.go
idea404: Kept the git, as we now can support scaffolding `omni-network` templates that don't use foundry

cli/cmd/develop.go
idea404: good one, added

PR #1211, opened 2024-06-11
chore(halo): revert valsync and split out registry

corverroos:
Split out the `registry` module from `epochsync` and revert back to `valsync`.

Also, in halo attest module: Instead of maintaining a expesive DB index to dedup and group votes per unique attestation, use a `key_hash` that is calculated with available fields.

task: none


e2e/types/chain.go
```diff
@@ -40,7 +40,7 @@ var (
 func OmniEVMByNetwork(network netconf.ID) EVMChain {
 	return EVMChain{
 		Metadata: mustMetadata(network.Static().OmniExecutionChainID),
-		Shards:   allShards,
+		Shards:   []xchain.ShardID{xchain.ShardFinalized0}, // OmniEVM has instant finality, so Latest == Finalized.
```
corverroos, line 43: only support `Finalized` for omniEVM

kevinhalliday (approved):

PR #1223, opened 2024-06-13
chore(buildinfo): bump minor version

idea404:
updates buildinfo minor version, as there are breaking changes in the syntax for `XApp` introduced by confirmation level features (and bumping the major feels wrong before mainnet)


corverroos (changes requested):
Lets keep it simple, just keep incrementing patch versions

PR #1225, opened 2024-06-13
fix(ci): use ref name in tag releases

idea404:
fixes parsing error found in https://github.com/omni-network/omni/actions/runs/9501746339/job/26188033774

### DIFF
--- a/.github/workflows/release-official.yml
+++ b/.github/workflows/release-official.yml
@@ -37,5 +37,5 @@ jobs:
         with:
           context: "{{defaultContext}}:explorer/ui"
           push: true
-          tags: omniops/explorer-ui:main,omniops/explorer-ui:${{ steps.git_ref.outputs.short_sha }},omniops/explorer-ui:${GITHUB_REF##*/}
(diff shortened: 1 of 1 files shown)


corverroos (approved):

.github/workflows/release-official.yml
```diff
@@ -37,5 +37,5 @@ jobs:
         with:
           context: "{{defaultContext}}:explorer/ui"
           push: true
-          tags: omniops/explorer-ui:main,omniops/explorer-ui:${{ steps.git_ref.outputs.short_sha }},omniops/explorer-ui:${GITHUB_REF##*/}
+          tags: omniops/explorer-ui:main,omniops/explorer-ui:${{ steps.git_ref.outputs.short_sha }},omniops/explorer-ui:${{ github.ref_name }}
```
pavelnikolov, line 40: Would this continue to work if we deploy staging env?

pavelnikolov: It's fine to release the `ref_name` but we can leave the short git sha. Tags are cheap - let's push both.

PR #1229, opened 2024-06-13
feat(halo/evmengine): burn evm fees

corverroos:
Burn execution fees by using the default `0xdead` address as `SuggestedFeeRecipient`.

task: none


corverroos:
Redoing after octane move

PR #1232, opened 2024-06-14
feat(contracts): add max xmsg data size

kevinhalliday:
Add max xmsg data size, set to 20,000 bytes,

task: https://app.asana.com/0/1206208509925075/1207573604321174


contracts/src/xchain/OmniPortalStorage.sol
```diff
@@ -38,7 +43,7 @@ abstract contract OmniPortalStorage is IOmniPortal, IOmniPortalAdmin {
     /**
      * @notice Virtual chain ID used in xmsgs from Omni's consensus chain
      */
-    uint64 public omniCChainID;
+    uint64 public omniCChainId;
```
kevinhalliday, line 46: I think I'm gonna go back and make everything `ID` instead of `Id`

pavelnikolov (approved):

PR #1238, opened 2024-06-14
fix(octane/evmengine): handle unknown payload error

corverroos:
Handle edge-case of omni_evm restart after payload was triggered in which case it may return "Unknown Payload". Currently it gets stuck in that loop. Now, it will error and propose an empty block.

Also increase genesis validator power to 100 since that is the minimum stake in the omni staking contract.

task: none


halo/valsync/keeper/keeper.go
```diff
@@ -258,7 +258,7 @@ func (k *Keeper) insertValidatorSet(ctx context.Context, vals []*Validator, isGe
 	// Log a warn if any validator has 1/3 or more of the total power.
 	// This is a potential attack vector, as a single validator could halt the chain.
 	for address, power := range powers {
-		if power >= totalPower/3 {
+		if power >= totalPower/3 && len(powers) > 1 {
```
corverroos, line 261: don't log this in simnet where we only have 1 validator

kevinhalliday (approved):

PR #1250, opened 2024-06-18
fix(halo/attest): ensure validators only include own votes

corverroos:
Ensure that validators only include votes signed by themselves. This mitigates against malicious validators including votes from other validators.

task: none


kevinhalliday (approved):

halo/attest/keeper/valcache.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// valCache is a simple read-through cache for validator eth-to-comet address lookups.
+type valCache struct {
+	sync.RWMutex
+	valsByAddr map[common.Address]*vtypes.Validator
```
arajasek, line 20: why not just have this be a `map[common.Address]crypto.Address`?

halo/attest/keeper/valcache.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	return cmtAddr, true, nil
+}
+
+func (c *valCache) SetAll(vals []*vtypes.Validator) error {
+	c.Lock()
+	defer c.Unlock()
+
+	var valsByAddr = make(map[common.Address]*vtypes.Validator, len(vals))
```
arajasek, line 44: Any reason to wipe the map here? 

Idk how common it's gonna be for validators to drop in and out of the active set, but we could just add new vals to the existing map without dropping any old ones (in case they pop back in later)?

arajasek (commented):
2 quick questions, apologies if I'm off-base

halo/attest/keeper/valcache.go
kirk-baird: If the map isn't cleared then you can make a VE with signatures from validators that have exited. This would be an issues as `PrepareVotes()` does not filter out votes outside the validator set but `ProcessProposal()` will reject a proposal containing votes outside of the set. So a malicious VE would prevent validators creating valid proposals and stall the chain.

Having the `valCache` strictly check the current validator set should prevent this attack vector. However, it would also be worth while having a filter in `halo/attest/keeper/cpayload.go::PrepareVotes()` to skip votes from validators outside the set.

arajasek: Hmm, yeah, relying on this to reject signatures from former (exited) validators seems flimsy, I would agree that we should have that check elsewhere if we don't already. I'm not even sure it works, because we only wipe this cache upon a cache-miss, so exited validators will definitely still be present here at certain points.

Apologies if I'm misreading things!

kirk-baird: Good point, I hadn't accounted for that we only update the set on a cache miss. Then I agree that an alternate fix for this is required. 

I like still the idea of a filter in `PrepareProposal()`, an option could be to force a cache update whenever we call `PrepareProposal()` and then filter based on a fresh DB. 

Another option is to only cache validators for a specific height. That is `k.valCache.GetCometAddress(ethAddr)` passes the context as well and checks the cached height against the context height. If there is a new height we refresh the cache.

halo/attest/keeper/valcache.go
corverroos: agreed, will limit this to the use-case it should be used in.

PR #1262, opened 2024-06-20
refactor(halo/attest): improve vote extension verification

corverroos:
Slight refactoring that limits the scope and responsibility of the `valAddrs` cache used to ensure that VEs only includes votes from the validator itself.

task: none


halo/attest/keeper/keeper.go
```diff
@@ -51,7 +50,7 @@ type Keeper struct {
 	voteExtLimit   uint64
 	trimLag        uint64
 
-	valCache *valCache
+	valAddrs *valAddrs
```
arajasek, line 53: nit: i'd still name the field and the type `valAddrsCache`, just so it's obvious what it's doing

halo/attest/keeper/valaddr.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/cometbft/cometbft/crypto"
+
+	"github.com/ethereum/go-ethereum/common"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// valAddrs is a simple read-through cache for validator comet-to-eth address lookups.
+// It only supports the current validator set; as per cometBFT.
```
arajasek, line 18: I'd drop this line -- from my perspective `valAddrs` doesn't care one way or the other about whether or not its entries correspond  to the current validator set. All it knows is the correct mapping of some Comet addrs to Eth addrs.

arajasek (approved):
Couple nits, but LGTM.

PR #1266, opened 2024-06-20
fix(docs): correct old path in example cmd

arajasek:
Fix the example command, out of date since #160

task: none


pavelnikolov:
@arajasek there is a `verifypr` workflow which was failing. It complained because the `task: ...` line had trailing spaces.

pavelnikolov (approved):

kevinhalliday (approved):

PR #1270, opened 2024-06-21
fix(halo/relayer): increase gas limit of consensus chain submissions

corverroos:
Fix issue introduces by `adding mock portals` that resulted in `out-of-gas` submission reverted issues of large `setNetwork` consensus chain messages by increasing gas limit for ephemeral chains and using proper gas estimation for protected chains. 

Also prevent ephemeral chains from loading mock portals into netconf. 

task: none

### DIFF
--- a/relayer/app/creator.go
+++ b/relayer/app/creator.go
@@ -4,13 +4,15 @@ import (
 	"slices"
 
 	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/netconf"
 	"github.com/omni-network/omni/lib/xchain"
 )
 
 const (
-	subGasBase         uint64 = 500_000
-	subGasXmsgOverhead uint64 = 100_000
-	subGasMax          uint64 = 10_000_000 // Many chains has block gas limit of 30M, so we limit ourselves to 1/3 of that.
+	subGasBase               uint64 = 500_000
+	subGasXmsgOverhead       uint64 = 100_000
+	subGasMax                uint64 = 10_000_000 // Many chains have block gas limit of 30M, so we limit ourselves to 1/3 of that.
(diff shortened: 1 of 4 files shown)


relayer/app/creator.go
```diff
@@ -4,13 +4,15 @@ import (
...
 	"github.com/omni-network/omni/lib/xchain"
 )
 
 const (
-	subGasBase         uint64 = 500_000
-	subGasXmsgOverhead uint64 = 100_000
-	subGasMax          uint64 = 10_000_000 // Many chains has block gas limit of 30M, so we limit ourselves to 1/3 of that.
+	subGasBase               uint64 = 500_000
+	subGasXmsgOverhead       uint64 = 100_000
+	subGasMax                uint64 = 10_000_000 // Many chains has block gas limit of 30M, so we limit ourselves to 1/3 of that.
```
pavelnikolov, line 14: nit: typo "Many chains have"

pavelnikolov (approved):

relayer/app/creator.go
```diff
@@ -24,6 +26,22 @@ func estimateGas(msgs []xchain.Msg) uint64 {
...
+// For protected chains, we rely on proper gas estimation.
+//
+// Proper gas estimation for protected chains is ok since real world consensus chain messages are rare,
+// so the multiple-submissions-per-block-gas-estimation-wrong-offset issue isn't a problem.
+func consensusGasLimit(network netconf.ID) uint64 {
+	if network.IsEphemeral() {
+		return subEphemeralConsensusGas
+	}
+
+	return 0
```
arajasek, line 42: // we return 0, which is the signal to use proper gas estimation

arajasek (approved):

PR #1273, opened 2024-06-22
fix(readme): avs register link

AtomicInnovation321:
Body goes here...

task: none or https://app.asana.com/0/1206208509925075/<task_number>


ttarsi:
Thanks for the contribution! We actually need to update this to "https://docs.omni.network/tools/cli/" though

PR #1274, opened 2024-06-22
fix(attest): two fixes to deletion logic

arajasek:
- If we find something in the cache, we should return true.
- don't delete anything if we don't have our first approved attestation

task: none

### DIFF
--- a/halo/attest/keeper/keeper.go
+++ b/halo/attest/keeper/keeper.go
@@ -757,9 +757,10 @@ func (k *Keeper) deleteBefore(ctx context.Context, height uint64) error {
 	// Cache latest offset by chain version so we don't delete it.
 	// TODO(corver): Add tests for this.
 	latestByChain := make(map[xchain.ChainVersion]uint64)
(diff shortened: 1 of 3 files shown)


arajasek:
Test to come.

arajasek:
> Test to come.

Done.

ttarsi (approved):

kevinhalliday (approved):

corverroos (approved):

halo/attest/keeper/keeper.go
```diff
@@ -796,7 +797,7 @@ func (k *Keeper) deleteBefore(ctx context.Context, height uint64) error {
 		// once we start catching up.
```
corverroos, line 797: maybe add a comment here, something like: `also don't delete anything before we have at least one approved attestation`

corverroos: Other option is to remove the boolean and just return `0` if not found, then we will not delete anything either

arajasek: Comment added, thanks!

PR #1280, opened 2024-06-24
feat(halo): add womni predeploy

kevinhalliday:
Add WOmni predeploy.

task: https://app.asana.com/0/1206208509925075/1207601759688877


contracts/src/token/WOmni.sol
```diff
@@ -0,0 +1,101 @@
...
+
+pragma solidity 0.8.15;
+
+import { IWOmni } from "src/interfaces/IWOmni.sol";
+
+/**
+ * @title WOmni
+ * @notice WOmni is a modifed version of WETH98, is a version of WETH9 upgraded for Solidity 0.8.x.
+ */
+contract WOmni is IWOmni {
```
kevinhalliday, line 28: Should probably be called WOMNI, but we (meaning me) named our token contract Omni, instead of OMNI, so just staying consistent

pavelnikolov (approved):

contracts/src/token/WOmni.sol
```diff
@@ -0,0 +1,101 @@
...
+
+    /// @notice Pipes to deposit.
+    receive() external payable {
+        deposit();
+    }
+
+    /// @notice Pipes to deposit.
+    fallback() external payable {
+        deposit();
+    }
```
ttarsi, line 42: why do we need both receive and fallback?

contracts/src/token/WOmni.sol
```diff
@@ -0,0 +1,101 @@
...
+// Based on WETH9 by Dapphub, and WETH98 by OP Labs.
+// Modified  by Omni Network
+
+pragma solidity 0.8.15;
+
+import { IWOmni } from "src/interfaces/IWOmni.sol";
+
+/**
+ * @title WOmni
+ * @notice WOmni is a modifed version of WETH98, is a version of WETH9 upgraded for Solidity 0.8.x.
```
ttarsi, line 26: maybe comment what the changes are from the original? and link the original

ttarsi (approved):

PR #1284, opened 2024-06-25
fix(lib/txmgr): fix txmgr intrinsic gas too low err

pavelnikolov:
* Retry with a fee bump on "_Intrinsic gas too low_" error
* Log an error instead of a warn on "_ Unknown error publishing transaction_"

task: none

### DIFF
--- a/lib/txmgr/txmgr.go
+++ b/lib/txmgr/txmgr.go
@@ -21,9 +21,19 @@ import (
 	"github.com/ethereum/go-ethereum/params"
 )
 
+// TxError is a typed error for transaction errors.
+type TxError string
+
+func (e TxError) Error() string {
+	return string(e)
+}
+
 const (
 	// PriceBump geth requires a minimum fee bump of 10% for regular tx resubmission.
 	PriceBump int64 = 10
+
+	// ErrIntrinsicGasTooLow is the error message returned by json-rpc when the intrinsic gas is too low.
+	ErrIntrinsicGasTooLow TxError = "intrinsic gas too low"
 )
 
 // TxManager is an interface that allows callers to reliably publish txs,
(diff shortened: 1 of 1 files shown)


lib/txmgr/txmgr.go
```diff
@@ -361,8 +370,11 @@ func (m *simple) publishTx(ctx context.Context, tx *types.Transaction, sendState
 		case errStringMatch(err, txpool.ErrUnderpriced):
 			log.Warn(ctx, "Transaction is underpriced", err)
 			continue // retry with fee bump
+		case errStringMatch(err, ErrIntrinsicGasTooLow):
+			log.Warn(ctx, "Intrinsic gas too low", err)
+			continue // retry with fee bump
 		default:
-			log.Warn(ctx, "Unknown error publishing transaction", err)
+			log.Error(ctx, "Unknown error publishing transaction", err)
```
pavelnikolov, line 377: Make it more obvious that we are getting na unknown error here.

idea404 (approved):

lib/txmgr/txmgr.go
```diff
@@ -21,9 +21,19 @@ import (
...
+func (e TxError) Error() string {
+	return string(e)
+}
+
 const (
 	// PriceBump geth requires a minimum fee bump of 10% for regular tx resubmission.
 	PriceBump int64 = 10
+
+	// ErrIntrinsicGasTooLow is the error message returned by json-rpc when the intrinsic gas is too low.
+	ErrIntrinsicGasTooLow TxError = "intrinsic gas too low"
```
corverroos, line 36: no need for this error, please use geth [error](https://github.com/ethereum/go-ethereum/blob/master/core/error.go#L81) instead

PR #1289, opened 2024-06-25
feat(cprovider): query historical state for attestations

arajasek:
- adds new `EarliestAttestationInState` method to Attest keeper's query interface
  - [ ] We could instead modify `LatestAttestation` to be `EarliestAndLatestAttestations`, but we will likely still need 2 db reads. See review comments.
- implements historical walkback for the cchain provider's `fetch` function
  - in the event that we don't find the requested attestation offset, we exponentially backoff going back in time
  - we stop when we find an earlier attestation, and then perform binary search (if necessary) to find the requested attestation offset
- NOTE: I've confirmed that all our tests pass with this change AND the keeper's [trimLag](https://github.com/omni-network/omni/blob/main/halo/attest/keeper/keeper.go#L51) set to 1

task: https://app.asana.com/0/1206208509925075/1207131279471036

### DIFF
--- a/e2e/test/attestations_test.go
+++ b/e2e/test/attestations_test.go
@@ -28,7 +28,7 @@ func TestApprovedAttestations(t *testing.T) {
 		ctx := context.Background()
 		for _, portal := range portals {
 			for _, chainVer := range portal.Chain.ChainVersions() {
-				atts, err := fetchAllAtts(ctx, cprov, chainVer)
+				atts, err := fetchAllAtts(ctx, cprov, chainVer, node.StartAt > 0)
 				require.NoError(t, err)
 
 				// Only non-empty blocks are attested to, and we don't know how many that should be, so just ensure it isn't 0.
@@ -40,7 +40,7 @@ func TestApprovedAttestations(t *testing.T) {
 		consChain, ok := network.OmniConsensusChain()
 		require.True(t, ok)
 		for _, chainVer := range consChain.ChainVersions() {
-			atts, err := fetchAllAtts(ctx, cprov, chainVer)
+			atts, err := fetchAllAtts(ctx, cprov, chainVer, node.StartAt > 0)
 			require.NoError(t, err)
 			require.NotEmpty(t, len(atts))
 		}
@@ -102,11 +102,17 @@ func TestApprovedValUpdates(t *testing.T) {
 	})
 }
 
-func fetchAllAtts(ctx context.Context, cprov cchain.Provider, chainVer xchain.ChainVersion) ([]xchain.Attestation, error) {
+func fetchAllAtts(ctx context.Context, cprov cchain.Provider, chainVer xchain.ChainVersion, nodeIsDelayed bool) ([]xchain.Attestation, error) {
 	fromOffset := uint64(1) // Start at initialXBlockOffset
 	var resp []xchain.Attestation
 	for {
 		atts, err := cprov.AttestationsFrom(ctx, chainVer, fromOffset)
+		if provider.IsErrHistoryPruned(err) && nodeIsDelayed {
+			// For delayed nodes, we might not have the first attestations.
+			// Just move on and return what we do find in state.
+			fromOffset++
+			continue
+		}
 		if err != nil {
 			return nil, err
 		}
--- a/halo/attest/keeper/keeper.go
+++ b/halo/attest/keeper/keeper.go
@@ -469,7 +469,45 @@ func (k *Keeper) latestAttestation(ctx context.Context, version xchain.ChainVers
 	}
 
 	if iter.Next() {
-		return nil, false, errors.New("multiple attestation found")
+		return nil, false, errors.New("multiple attestations found")
+	}
+
+	// If this attestation is overridden by a finalized attestation, return that instead.
+	if att.GetFinalizedAttId() != 0 {
+		att, err := k.attTable.Get(ctx, att.GetFinalizedAttId())
+		if err != nil {
+			return nil, false, errors.Wrap(err, "get finalized attestation")
+		}
+
+		return att, true, nil
+	}
+
+	return att, true, nil
+}
+
+// earliestAttestation returns the earliest approved attestation for the given chain currently found in state,
+// or false if none is found.
+func (k *Keeper) earliestAttestation(ctx context.Context, version xchain.ChainVersion) (*Attestation, bool, error) {
+	defer latency("earliest_attestation")()
+
+	idx := AttestationStatusChainIdConfLevelBlockOffsetIndexKey{}.WithStatusChainIdConfLevel(uint32(Status_Approved), version.ID, uint32(version.ConfLevel))
+	iter, err := k.attTable.List(ctx, idx, ormlist.DefaultLimit(1))
+	if err != nil {
+		return nil, false, errors.Wrap(err, "list")
+	}
+	defer iter.Close()
+
+	if !iter.Next() {
+		return nil, false, nil
+	}
+
+	att, err := iter.Value()
+	if err != nil {
+		return nil, false, errors.Wrap(err, "value")
+	}
+
+	if iter.Next() {
+		return nil, false, errors.New("multiple attestations found")
 	}
 
 	// If this attestation is overridden by a finalized attestation, return that instead.
--- a/halo/attest/keeper/query.go
+++ b/halo/attest/keeper/query.go
@@ -49,6 +49,28 @@ func (k *Keeper) LatestAttestation(ctx context.Context, req *types.LatestAttesta
 	return &types.LatestAttestationResponse{Attestation: AttestationFromDB(att, sigs)}, nil
 }
 
+func (k *Keeper) EarliestAttestation(ctx context.Context, req *types.EarliestAttestationRequest) (*types.EarliestAttestationResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+
+	chainVer := xchain.ChainVersion{ID: req.ChainId, ConfLevel: xchain.ConfLevel(req.ConfLevel)}
+
+	att, ok, err := k.earliestAttestation(ctx, chainVer)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	} else if !ok {
+		return nil, status.Error(codes.NotFound, "no approved attestations for chain")
+	}
+
+	sigs, err := k.getSigs(ctx, att.GetId())
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	return &types.EarliestAttestationResponse{Attestation: AttestationFromDB(att, sigs)}, nil
+}
+
 func (k *Keeper) WindowCompare(ctx context.Context, req *types.WindowCompareRequest) (*types.WindowCompareResponse, error) {
 	if req == nil {
 		return nil, status.Error(codes.InvalidArgument, "invalid request")
--- a/halo/attest/types/query.proto
+++ b/halo/attest/types/query.proto
@@ -16,6 +16,9 @@ service Query {
   // LatestAttestation queries halo for the latest approved attestations for the given chain_id.
   rpc LatestAttestation(LatestAttestationRequest) returns (LatestAttestationResponse) {}
 
+  // EarliestAttestation queries halo for the earliest approved attestation in consensus chain state for the given chain_id.
+  rpc EarliestAttestation(EarliestAttestationRequest) returns (EarliestAttestationResponse) {}
+
   // WindowCompare compares the given chain height to the vote window.
   // It returns whether the request is behind (-1), or in (0), or after (1) the vote window.
   // The vote window is a configured number of blocks around the latest approved attestation.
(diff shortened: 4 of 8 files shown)


arajasek:
- adds new `EarliestAttestationInState` method to Attest keeper's query interface
  - [ ] We could instead modify `LatestAttestation` to be `EarliestAndLatestAttestations`, but we will likely still need 2 db reads. See review comments.
- implements historical walkback for the cchain provider's `fetch` function
  - in the event that we don't find the requested attestation offset, we exponentially backoff going back in time
  - we stop when we find an earlier attestation, and then perform binary search (if necessary) to find the requested attestation offset

halo/attest/keeper/keeper.go
```diff
@@ -462,7 +462,46 @@ func (k *Keeper) latestAttestation(ctx context.Context, version xchain.ChainVers
...
+
+		return att, true, nil
+	}
+
+	return att, true, nil
+}
+
+// earliestAttestationInState returns the earliest approved attestation for the given chain currently found in state,
+// or false if none is found.
+func (k *Keeper) earliestAttestationInState(ctx context.Context, version xchain.ChainVersion) (*Attestation, bool, error) {
```
arajasek, line 484: Instead of adding a new method here, we could combine this into `latestAttestation` and save some query calls. However, I'm guessing what we _really_ want to conserve is database reads, and I think we'll still need two reads (one reversed and one not).

halo/attest/keeper/query.go
```diff
@@ -49,6 +49,28 @@ func (k *Keeper) LatestAttestation(ctx context.Context, req *types.LatestAttesta
 	return &types.LatestAttestationResponse{Attestation: AttestationFromDB(att, sigs)}, nil
 }
 
+func (k *Keeper) EarliestAttestationInState(ctx context.Context, req *types.EarliestAttestationInStateRequest) (*types.EarliestAttestationInStateResponse, error) {
```
arajasek, line 52: Planning on adding unit tests in a separate PR, but feel free to say "no you have to do it now" :P

halo/attest/keeper/keeper.go
```diff
@@ -462,7 +462,46 @@ func (k *Keeper) latestAttestation(ctx context.Context, version xchain.ChainVers
 	}
 
 	if iter.Next() {
-		return nil, false, errors.New("multiple attestation found")
+		return nil, false, errors.New("multiple attestations" +
+			" found")
```
pavelnikolov, line 466: nit: Why is this error msg broken in multiple lines?

pavelnikolov (approved):

halo/attest/keeper/keeper.go
arajasek: fixed, thanks

lib/cchain/provider/abci.go
```diff
@@ -138,15 +139,119 @@ func newABCIFetchFunc(cl atypes.QueryClient) fetchFunc {
...
-		atts, err := atypes.AttestationsFromProto(resp.Attestations)
+		if len(resp.Attestations) != 0 {
+			atts, err := atypes.AttestationsFromProto(resp.Attestations)
+			if err != nil {
+				return nil, errors.Wrap(err, "attestations from proto")
+			}
+
+			return atts, nil
+		}
+
```
corverroos, line 153: can we extract the searching steps into a separate function maybe?
```
height, ok, err := searchOffsetInHistory()
if err != nil {
  return nil, err
} else if !ok {
  return nil, nil
}

return attsFromAtHeight(ctx, req, height)
```

lib/cchain/provider/abci.go
```diff
@@ -138,15 +139,119 @@ func newABCIFetchFunc(cl atypes.QueryClient) fetchFunc {
 		})
 		if err != nil {
 			incQueryErr(endpoint)
-			return nil, errors.Wrap(err, "abci query approved-from")
+			return nil, errors.Wrap(err, "abci query attestations-from")
 		}
 
-		atts, err := atypes.AttestationsFromProto(resp.Attestations)
+		if len(resp.Attestations) != 0 {
```
corverroos, line 145: What the probability of findings attestations at the latest height? Why do we first try and avoid searching by querying latest height, and if we don't find anything, then we start searching? We could also just always first search.

lib/cchain/provider/abci.go
```diff
@@ -138,15 +139,119 @@ func newABCIFetchFunc(cl atypes.QueryClient) fetchFunc {
...
+			}
+
+			// Otherwise, keep moving back
+			endHeightIndex = queryHeight
+			lookback *= 2
+			queryHeight -= lookback
 		}
 
-		return atts, nil
+		log.Debug(ctx, "Cchain fetch lookback done ", "steps", lookbackStepsCounter)
```
corverroos, line 197: lets add metrics for the searching steps: lookback_steps and search_steps histogram.

lib/cchain/provider/abci.go
corverroos: We need metrics that show this optimisation is worth the extra query.

e2e/test/attestations_test.go
```diff
@@ -102,11 +103,17 @@ func TestApprovedValUpdates(t *testing.T) {
 	})
 }
 
-func fetchAllAtts(ctx context.Context, cprov cchain.Provider, chainVer xchain.ChainVersion) ([]xchain.Attestation, error) {
+func fetchAllAtts(ctx context.Context, cprov cchain.Provider, chainVer xchain.ChainVersion, nodeIsDelayed bool) ([]xchain.Attestation, error) {
 	fromOffset := uint64(1) // Start at initialXBlockOffset
 	var resp []xchain.Attestation
 	for {
 		atts, err := cprov.AttestationsFrom(ctx, chainVer, fromOffset)
+		if err != nil && strings.Contains(err.Error(), "no commit info found") && nodeIsDelayed {
```
corverroos, line 111: feels like we can improve the interface or error which indicates "attestation pruned from historical state". Having to check for "no commit info found" isn't obvious.

What about some helper method like:
```
if cchain.IsErrHistoryPruned(err) {
  fromOffset++
}
```

lib/cchain/provider/abci.go
```diff
@@ -138,15 +139,119 @@ func newABCIFetchFunc(cl atypes.QueryClient) fetchFunc {
...
+			return nil, errors.Wrap(err, "abci query info")
+		}
+		startHeightIndex := 0
+		endHeightIndex := int(info.Response.LastBlockHeight)
+		lookback := 1
+		lookbackStepsCounter := 0 // For metrics only
+		queryHeight := endHeightIndex - lookback
+		for queryHeight > 0 {
+			lookbackStepsCounter++
+			earliestAtt, ok, err := queryEarliestAttestation(ctx, cl, chainVer, queryHeight)
```
corverroos, line 179: queryHeight is exponentially jumping backwards, if it crosses the prune height this will error out. But the DB might contain the offset at a highter height.

```
----|--------|--|-----|---->
    A        B  C     D
```
A: current queryHeight
B: pruned height
C: height containing attestation
D: previous iteration queryHeight

We need to catch and handle `if cchain.IsErrHistoryPruned`

corverroos (approved):

halo/attest/types/query.proto
```diff
@@ -16,6 +16,9 @@ service Query {
   // LatestAttestation queries halo for the latest approved attestations for the given chain_id.
   rpc LatestAttestation(LatestAttestationRequest) returns (LatestAttestationResponse) {}
 
+  // EarliestAttestationInState queries halo for the earliest approved attestation in consensus chain state for the given chain_id.
+  rpc EarliestAttestationInState(EarliestAttestationInStateRequest) returns (EarliestAttestationInStateResponse) {}
```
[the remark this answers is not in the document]
arajasek: Yeah, you're right. I was worried about it being misconstrued as the earliest attestation for the chain in question, but that's just attestation 1 always. Thank you!

lib/cchain/provider/abci.go
arajasek: Agreed. My instinct is that when things are working normally (relayer is keeping sync well), we'll mostly be querying at the latest height. But you're right, metrics are needed.

lib/cchain/provider/abci.go
```diff
@@ -138,15 +139,119 @@ func newABCIFetchFunc(cl atypes.QueryClient) fetchFunc {
...
+		// Exponentially backoff to find a good start point for binary search, this prefers more recent queries
+		info, err := client.ABCIInfo(ctx)
+		if err != nil {
+			return nil, errors.Wrap(err, "abci query info")
+		}
+		startHeightIndex := 0
+		endHeightIndex := int(info.Response.LastBlockHeight)
+		lookback := 1
+		lookbackStepsCounter := 0 // For metrics only
+		queryHeight := endHeightIndex - lookback
```
corverroos, line 176: Be careful when subtracting. This can be negative, what happens then? Lets handle that case explicitly.

lib/cchain/provider/abci.go
arajasek: Good catch, thank you!

lib/cchain/provider/provider.go
```diff
@@ -147,3 +148,11 @@ func (p Provider) Subscribe(in context.Context, chainVer xchain.ChainVersion, xB
 		}
 	}()
 }
+
+func IsErrHistoryPruned(err error) bool {
```
corverroos, line 152: add a comment please

lib/cchain/provider/abci.go
```diff
@@ -280,3 +315,194 @@ func (a rpcAdaptor) Invoke(ctx context.Context, method string, req, resp any, _
 func spanName(endpoint string) string {
 	return "cprovider/" + endpoint
 }
+
+// Finds the height at which the requested offset can be found in state.
+func searchOffsetInHistory(ctx context.Context, client ABCIClient, cl atypes.QueryClient, chainVer xchain.ChainVersion, fromOffset uint64) (int, error) {
+	// TODO: Aayush: Is it fine to be capturing a second thing here?
```
corverroos, line 321: sure

lib/cchain/provider/abci.go
```diff
@@ -280,3 +315,194 @@ func (a rpcAdaptor) Invoke(ctx context.Context, method string, req, resp any, _
...
+	const endpoint = "search_offset"
+	defer latency(endpoint)
+
+	// Exponentially backoff to find a good start point for binary search, this prefers more recent queries
+	info, err := client.ABCIInfo(ctx)
+	if err != nil {
+		return -1, errors.Wrap(err, "abci query info")
```
corverroos, line 328: always prefer to return zero values along with errors, not -1.

lib/cchain/provider/abci.go
```diff
@@ -280,3 +315,194 @@ func (a rpcAdaptor) Invoke(ctx context.Context, method string, req, resp any, _
 func spanName(endpoint string) string {
 	return "cprovider/" + endpoint
 }
+
+// Finds the height at which the requested offset can be found in state.
+func searchOffsetInHistory(ctx context.Context, client ABCIClient, cl atypes.QueryClient, chainVer xchain.ChainVersion, fromOffset uint64) (int, error) {
```
corverroos, line 320: godoc comments have a strict format, you must start with the name of the function/type:

```
// searchOffsetInHistory searches the consensus state history and 
// returns the historical consensus block height that contains an approved attestation
// for the provided chain version and fromOffset.
```

corverroos: we always use `uint64` for heights and offsets, lets stick to that #consistencyFTW

lib/cchain/provider/abci.go
```diff
@@ -280,3 +315,194 @@ func (a rpcAdaptor) Invoke(ctx context.Context, method string, req, resp any, _
...
+			break
+		}
+
+		// Otherwise, keep moving back
+		endHeightIndex = queryHeight
+		lookback *= 2
+		if queryHeight > lookback {
+			queryHeight -= lookback
+		} else {
+			queryHeight = 1
```
corverroos, line 377: lets just break out of this loop if queryHeight reaches zero, since using 0 for start of binary search is the default.

lib/cchain/provider/abci.go
```diff
@@ -280,3 +315,194 @@ func (a rpcAdaptor) Invoke(ctx context.Context, method string, req, resp any, _
...
+		}
+	}
+
+	// We now have reasonable start and end indices for binary search
+	binarySearchStepsCounter := 0 // For metrics only
```
corverroos, line 382: we prefer to initialise zero variables using the `var foo int` pattern. As per [Practical Go](https://dave.cheney.net/practical-go/presentations/gophercon-singapore-2019.html#_use_a_consistent_declaration_style)

lib/cchain/provider/abci.go
```diff
@@ -280,3 +385,43 @@ func (a rpcAdaptor) Invoke(ctx context.Context, method string, req, resp any, _
 func spanName(endpoint string) string {
 	return "cprovider/" + endpoint
 }
+
+func queryEarliestAttestation(ctx context.Context, cl atypes.QueryClient, chainVer xchain.ChainVersion, height int) (xchain.Attestation, bool, error) {
```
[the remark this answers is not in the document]
corverroos: prefer to godoc style of comments: 

```
queryEarliestAttestation returns the earliest approved attestation at the provided historical consensus chain height. If height is 0, latest height is used.
```

corverroos (commented):
Looks good.

In general, add more function and type comments, especially for this type of code that is quite involved and not super obvious.

lib/cchain/provider/abci.go
arajasek: Yes, definitely -- thank you!

corverroos (approved):

PR #1292, opened 2024-06-26
fix(lib/fireblocks): handle errors and increase query interval

pavelnikolov:
* Handle non-json body errors properly
* Increase retry interval from 1s to 5s to avoid getting rate limited

task: none

### DIFF
--- a/lib/fireblocks/config.go
+++ b/lib/fireblocks/config.go
@@ -8,8 +8,8 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 )
 
-// options houses parameters for altering the behavior of a SimpleTxManager.
-type options struct {
+// Config houses parameters for altering the behavior of a SimpleTxManager.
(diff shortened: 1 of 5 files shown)


idea404 (approved):

lib/fireblocks/config.go
```diff
@@ -8,8 +8,8 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 )
 
-// options houses parameters for altering the behavior of a SimpleTxManager.
-type options struct {
+// Config houses parameters for altering the behavior of a SimpleTxManager.
+type Config struct {
```
pavelnikolov, line 12: This was falsely named `options` - renaming it to fix the mistake. An option in this context is something which changes the config - addressed by introducing an `Option` type below and making it public so that it can be used by other packages, e.g. e2e.

PR #1294, opened 2024-06-26
Added Base Sepolia config to manifests, types and evmchain

brid:
Adds configuration to support the Base Sepolia chain.

Uses the same pattern as per other chains.

Guidance from @kevinhalliday


e2e/manifests/staging.toml
```diff
@@ -1,5 +1,5 @@
 network = "staging"
-public_chains = ["op_sepolia"]
+public_chains = ["op_sepolia", "base_sepolia"]
```
ttarsi, line 2: let's just add it to omega, no need for staging to include it imo

ttarsi: we don't include arb sepolia

brid: Done!

pavelnikolov (approved):

PR #1298, opened 2024-06-27
chore(contracts): rename testnet to omega

pavelnikolov:
Rename some leftovers of testnet to omega.

task: none

### DIFF
--- a/lib/contracts/address.go
+++ b/lib/contracts/address.go
@@ -39,7 +39,7 @@ func MainnetCreate3Factory() common.Address {
 	return crypto.CreateAddress(eoa.MustAddress(netconf.Mainnet, eoa.RoleCreate3Deployer), 0)
 }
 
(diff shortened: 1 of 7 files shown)


idea404 (commented):
I believe the logic for these contracts can indeed be applied to any testnet not only Omega

pavelnikolov:
Correct, it can be applied to other testnets in future, however, if you pay attention inside of funcs like:

```go
func omegaCfg() DeploymentConfig {
	return DeploymentConfig{
		Deployer: eoa.MustAddress(netconf.Omega, eoa.RoleCreate3Deployer),
	}
}
```
we call `netconf.Omega` which applies specifically to this network. Right now there is only one testnet on the horizon and this code will need to be refactored in case of a new one.

corverroos (approved):

PR #1300, opened 2024-06-27
test(e2e): remove base from omega manifest

pavelnikolov:
Remove base from omega manifest for test purposes.

task: none


idea404 (approved):
approve for test

PR #1320, opened 2024-07-01
fix(halo/voter): verify state loaded from disk

corverroos:
Verify voter state loaded from disk to avoid panics due to corrupted state.

Fixes OMP-20 as per Sigma Prime audit.

task: none

### DIFF
--- a/halo/attest/voter/voter.go
+++ b/halo/attest/voter/voter.go
@@ -549,6 +549,22 @@ func loadState(path string) (stateJSON, error) {
 		return stateJSON{}, errors.Wrap(err, "unmarshal state path")
 	}
 
+	verify := func(voteSets ...[]*types.Vote) error {
+		for _, votes := range voteSets {
+			for _, vote := range votes {
+				if err := vote.Verify(); err != nil {
+					return errors.Wrap(err, "verify vote")
+				}
+			}
+		}
+
+		return nil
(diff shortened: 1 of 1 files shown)


pavelnikolov (approved):
LGTM. Do we need a unit tests that actually fails with this err?

halo/attest/voter/voter.go
```diff
@@ -549,6 +549,22 @@ func loadState(path string) (stateJSON, error) {
 		return stateJSON{}, errors.Wrap(err, "unmarshal state path")
 	}
 
+	verify := func(voteSets ...[]*types.Vote) error {
+		for _, votes := range voteSets {
```
arajasek, line 553: If it's invalid for any of  `Latest` / `Proposed` etc. to be nil, check that here (but I don't know whether that would be invalid, it seems like it would be fine to me).

corverroos: Not sure what you mean. 

A nil and empty slice is the same, and empty slices are allowed.
Nil votes are invalid, but that is handled in `vote.Verify()` (it doesn't panic)

arajasek: Yeah, my question is basically whether it's valid for any of the voteSets to be nil, and it sounds like that's a yes.

PR #1330, opened 2024-07-02
fix(lib/txmgr): handle network errors while querying receipts

corverroos:
Treats all network errors while querying receipts as temporary; retrying fetching receipts, don't give up.

This fixes the issue where a network error occurs, but the tx is mined, txmgr gives up querying receipt, and then gets stuck trying to bump fees and resubmit the tx (getting ErrNonceToLow). 

This caused flapping e2e tests when perturbations restart the EVM. 

task: https://app.asana.com/0/1206208509925075/1207634096794277

### DIFF
--- a/lib/txmgr/txmgr.go
+++ b/lib/txmgr/txmgr.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"log/slog"
 	"math/big"
+	"net"
 	"strings"
 	"sync"
 	"time"
@@ -18,6 +19,7 @@ import (
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/txpool"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/p2p/netutil"
 	"github.com/ethereum/go-ethereum/params"
 )
 
@@ -411,6 +413,8 @@ func (m *simple) waitMined(ctx context.Context, tx *types.Transaction,
 	const logFreqFactor = 10 // Log every 10th attempt
 	attempt := 1
 
+	netErrBackoff := expbackoff.New(ctx) // Additional backoff on network errors.
+
 	queryTicker := time.NewTicker(m.cfg.ReceiptQueryInterval)
 	defer queryTicker.Stop()
 	for {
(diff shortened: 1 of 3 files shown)


pavelnikolov (approved):

lib/txmgr/txmgr.go
```diff
@@ -677,3 +686,14 @@ func maybeSetTimeout(ctx context.Context, timeout time.Duration) (context.Contex
...
+
+// isNetworkError returns true if the error is a network error.
+func isNetworkError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	opErr := new(net.OpError)
+
+	return errors.As(err, &opErr)
```
arajasek, line 698: Any reason we're using `As` instead of `Is` here?

lib/txmgr/txmgr_internal_test.go
```diff
@@ -804,6 +814,40 @@ func TestWaitMinedReturnsReceiptAfterNotFound(t *testing.T) {
 	require.Equal(t, receipt.TxHash, txHash)
 }
 
+func TestWaitMinedRetiesNetError(t *testing.T) {
```
arajasek, line 817:
```suggestion
func TestWaitMinedRetriesNetError(t *testing.T) {
```

arajasek (approved):
Nits, but LGTM

lib/txmgr/txmgr.go
corverroos: Because it doesn't work.

```
func TestScratch(t *testing.T) {
	e1 := error(&net.OpError{Op: "dial", Net: "tcp", Source: nil, Addr: nil, Err: io.EOF})
	e2 := fmt.Errorf("wrapped: %w", e1)
	e3 := errors.Wrap(e2, "wrapped2")

	zeroOpErr := new(net.OpError)

	fmt.Printf("🔥!! =%v\n", errors.Is(e3, e1))
	fmt.Printf("🔥!! =%v\n", errors.Is(e3, e2))
	fmt.Printf("🔥!! =%v\n", errors.Is(e3, e3))
	fmt.Printf("🔥!! =%v\n", errors.Is(e3, zeroOpErr))
	fmt.Printf("🔥!! =%v\n", errors.As(e3, &zeroOpErr))
}
```


Results in:
```
🔥!! =true
🔥!! =true
🔥!! =true
🔥!! =false
🔥!! =true

```

corverroos: TL;DR:
 - `As`: check for types of errors
 - `Is`: checks for instances of errors.

arajasek: thanks for the explanation, i think i finally understand this :D

PR #1342, opened 2024-07-02
feat(attest): detect reorgs based on BlockOffset

### LINKED ISSUE #1312: feat(halo/attest): detect reorgs by streamOffset
We currently only [detect reorgs](https://github.com/omni-network/omni/blob/f7a6719a2913da6d9e96f0dedb9f6148ff3e5dbf/halo/attest/voter/voter.go#L515) based on "is parentBlockHash == prevBlockHash". We can harden this by also checking the StreamOffset for the messages in the blocks.

arajasek:
This is another check to harden against reorgs. Fixes #1312 

task: none


arajasek:
Hmm, actually, I'm not sure this is correct. Closing for now.

PR #1353, opened 2024-07-03
refactor(relayer): estimate pending gas instead of gas model

corverroos:
wip


e2e/app/run.go
```diff
@@ -91,6 +91,10 @@ func Deploy(ctx context.Context, def Definition, cfg DeployConfig) (*pingpong.XD
 		return nil, err
 	}
 
+	if err := waitForEVMs(ctx, networkFromDef(def), def.Backends()); err != nil {
```
corverroos, line 94: had to add this for some reason, mock_l1 startup has slowed down due to some other reason, haven't looked into it much, just implemented a workaround for now.

PR #1357, opened 2024-07-03
feat(xchain): fetch XMsgLogs and XMsgReceipts by blockhash

### LINKED ISSUE #1313: feat(xprovider): fetch by blockhash, not height
We should [query for msgs](https://github.com/omni-network/omni/blob/cbc8df960d848fad213c26c7568f2597a2f646da/lib/xchain/provider/fetch.go#L297) and receipts by blockhash, not height. This gives us a bit more confidence in what we query and is more reorg-resistant.

arajasek:
This gives us more reorg-resistance, and ensures that the header, XMsgs, and XReceipts returned by the provider's `GetBlock` are all consistent.

issue: #1313

### DIFF
--- a/lib/xchain/provider/fetch.go
+++ b/lib/xchain/provider/fetch.go
@@ -12,7 +12,9 @@ import (
 	"github.com/omni-network/omni/lib/tracer"
 	"github.com/omni-network/omni/lib/xchain"
 
+	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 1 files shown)


kevinhalliday (approved):

lib/xchain/provider/fetch.go
```diff
@@ -433,3 +422,26 @@ func headTypeFromConfLevel(conf xchain.ConfLevel) (ethclient.HeadType, bool) {
...
 	return chainVersion{ID: req.ChainID, ConfLevel: req.ConfLevel}
 }
+
+func getLogs(ctx context.Context, rpcClient ethclient.Client, contractAddr common.Address, blockHash common.Hash, topicName string) ([]types.Log, error) {
+	metadata, err := bindings.OmniPortalMetaData.GetAbi()
+	if err != nil {
+		return nil, errors.Wrap(err, "get abi")
+	}
+
+	topics, err := abi.MakeTopics([]any{metadata.Events[topicName].ID})
```
corverroos, line 432: rather use the topic inside generated bindings, see [example](https://github.com/omni-network/omni/blob/main/halo/registry/keeper/logeventproc.go#L41)

lib/xchain/provider/fetch.go
```diff
@@ -433,3 +422,26 @@ func headTypeFromConfLevel(conf xchain.ConfLevel) (ethclient.HeadType, bool) {
...
+	if err != nil {
+		return nil, errors.Wrap(err, "make topics")
+	}
+
+	logs, err := rpcClient.FilterLogs(ctx, ethereum.FilterQuery{
+		BlockHash: &blockHash,
+		Addresses: []common.Address{contractAddr},
+		Topics:    topics,
```
corverroos, line 440: note we could combine xmsgs and xreceipt into a single request and then parse logs based on topic. But maybe this isn't worth it...

corverroos (approved):

lib/xchain/provider/fetch.go
arajasek: Yeah, doesn't seem super worth it to me.

PR #1358, opened 2024-07-03
feat(attest): detect reorgs based on XMsg StreamOffsets

arajasek:
Keep an up-to-date map of offset based on StreamID, and report a reorg if the next block's XMsgs don't apply consecutively on our map's offsets.

issue: #1312

### DIFF
--- a/halo/attest/voter/voter.go
+++ b/halo/attest/voter/voter.go
@@ -191,6 +191,7 @@ func (v *Voter) runOnce(ctx context.Context, chainVer xchain.ChainVersion) error
 		Offset:    fromOffset,
 	}
 
+	cursors := make(map[xchain.StreamID]uint64)
 	var prevBlock xchain.Block
 
 	return v.provider.StreamBlocks(ctx, req,
@@ -199,7 +200,7 @@ func (v *Voter) runOnce(ctx context.Context, chainVer xchain.ChainVersion) error
 				return errors.New("not a validator anymore")
 			}
 
-			if err := detectReorg(chainVer, v.network.ChainVersionName(chainVer), prevBlock, block); err != nil {
+			if err := detectReorg(chainVer, v.network.ChainVersionName(chainVer), prevBlock, block, cursors); err != nil {
 				// Restart stream, recalculating block offset from finalized version.
 				return err
 			}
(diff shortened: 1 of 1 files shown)


halo/attest/voter/voter.go
```diff
@@ -191,6 +191,7 @@ func (v *Voter) runOnce(ctx context.Context, chainVer xchain.ChainVersion) error
 		Offset:    fromOffset,
 	}
 
+	cursors := make(map[xchain.StreamID]uint64, len(v.network.Chains))
```
arajasek, line 194: As far as I can tell, we shouldn't need a lock over this, because only one process can ever access this map. Happy to throw one on anyway if we think it's advisable.

arajasek, line 194: The relayer's `msgCursorFilter` that I based this off has a dedicated struct for the entries in this map (see `streamCursor` [here](https://github.com/omni-network/omni/blob/3aa61db7c4295185cb3a1f9bed5d1ce483ed72de/relayer/app/cursors.go#L131-L133)). I wasn't sure if this was considered better practice, and didn't want to blindly copy it, so I just have a `uint64`. 

Feedback welcome!

corverroos: Correct, lock is not required.

`len(v.network.Chains)` is incorrect, since we have multiple shards per destination chain per source chain. Best avoid setting length in this case.

halo/attest/voter/voter.go
```diff
@@ -516,6 +517,16 @@ func detectReorg(chainVer xchain.ChainVersion, chainVerName string, prevBlock xc
 		return errors.New("consecutive block height mismatch [BUG]", "prev_height", prevBlock.BlockHeight, "new_height", block.BlockHeight)
 	}
 
+	for _, xmsg := range block.Msgs {
+		cursor, ok := cursors[xmsg.StreamID]
+		if ok && xmsg.StreamOffset != cursor+1 {
+			return errors.New("consecutive message offset mismatch", "prev_offset", cursor, "new_offset", xmsg.StreamOffset)
```
corverroos, line 523: maybe add `xmsg.StreamID` to error args as well

halo/attest/voter/voter.go
corverroos: this is great #lessismore

corverroos (approved):

PR #1377, opened 2024-07-05
fix(xchain/provider): fix concurrent offset tracker issue

### LINKED ISSUE #1372: Race condition in xprovider offset tracker
The xprovider [tracker](https://github.com/omni-network/omni/blob/corver/monitorhaloflag/lib/xchain/provider/provider.go#L162) assumes that `fetchBatch` is called sequentially. This isn't the case. There are actually multiple concurrent workers calling it concurrently. 

This is especially the case for arb_sepolia, which has very fast blocks and uses 4 workers.

This results in `"get unexpected old height [BUG]"` see example [logs](https://omniops.grafana.net/explore?schemaVersion=1&panes=%7B%22lla%22:%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22builder%22,%22expr%22:%22%7Bnetwork%3D%5C%22staging%5C%22%7D%20%7C%3D%20%60BUG%60%20%7C%20logfmt%20%7C%20line_format%20%60%7C%20%7B%7B%20alignLeft%2040%20.msg%20%7D%7D%7B%7B%20range%20$k,%20$v%20:%3D%20.%7D%7D%7B%7Bif%20eq%20$k%20%5C%22msg%5C%22%20%5C%22time%5C%22%20%5C%22source%5C%22%20%5C%22stream%5C%22%20%5C%22job%5C%22%20%5C%22level%5C%22%20%5C%22container%5C%22%20%5C%22instance%5C%22%20%5C%22env%5C%22%20%5C%22stacktrace%5C%22%20%5C%22network%5C%22%20%5C%22service_name%5C%22%7D%7D%7B%7B%20else%20%7D%7D%7B%7B$k%7D%7D%3D%7B%7B$v%7D%7D%20%7B%7B%20end%20%7D%7D%7B%7B%20end%20%7D%7D%60%22,%22queryType%22:%22range%22,%22refId%22:%22A%22%7D%5D,%22range%22:%7B%22from%22:%22now-3h%22,%22to%22:%22now%22%7D%7D%7D&orgId=1).

corverroos:
Fixes concurrent fetch workers causing race in offset tracker. 

Instead, wait for "next height" before assigning the offset.

issue: fix #1372

### DIFF
--- a/lib/xchain/provider/provider.go
+++ b/lib/xchain/provider/provider.go
@@ -154,21 +154,16 @@ func (p *Provider) stream(
 		fromOffset = initialXOffset
 	}
 
-	tracker := newOffsetTracker(fromOffset)
+	tracker := newOffsetTracker(fromOffset, fromHeight, chain.AttestInterval)
 
 	deps := stream.Deps[xchain.Block]{
 		FetchWorkers: workers,
 		FetchBatch: func(ctx context.Context, chainID uint64, height uint64) ([]xchain.Block, error) {
-			offset, err := tracker.getOffset(height)
-			if err != nil {
-				return nil, err
-			}
-
 			fetchReq := xchain.ProviderRequest{
 				ChainID:   chainID,
 				Height:    height,
 				ConfLevel: req.ConfLevel,
-				Offset:    offset,
+				Offset:    0, // Don't assign block offset yet
 			}
 
 			xBlock, exists, err := p.GetBlock(ctx, fetchReq)
@@ -178,10 +173,12 @@ func (p *Provider) stream(
 				return nil, nil
 			}
 
-			if xBlock.ShouldAttest(chain.AttestInterval) {
-				if err := tracker.assignOffset(height); err != nil {
-					return nil, err
-				}
+			// Get block offset from tracker.
+			// Possibly waiting for another fetch worker to fetch "next height";
+			// since offsets need to be assigned in sequential order.
+			xBlock.BlockOffset, err = tracker.awaitOffset(ctx, xBlock)
+			if err != nil {
+				return nil, err
 			}
 
 			return []xchain.Block{xBlock}, nil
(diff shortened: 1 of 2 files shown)


pavelnikolov (approved):

lib/xchain/provider/provider.go
```diff
@@ -247,68 +244,83 @@ func (p *Provider) getEVMChain(chainID uint64) (netconf.Chain, ethclient.Client,
...
-	currentOffset uint64
-	prevHeight    *uint64
-}
-
-func newOffsetTracker(from uint64) *offsetTracker {
-	return &offsetTracker{
-		enabled:       true,
-		currentOffset: from,
+	attestInterval uint64
+	enabled        bool
```
arajasek, line 250: What is the purpose of this `enabled` field? It looks like it might be legacy -- we just set it to true on creation.

Can we just drop it?

lib/xchain/provider/provider.go
```diff
@@ -247,68 +244,83 @@ func (p *Provider) getEVMChain(chainID uint64) (netconf.Chain, ethclient.Client,
...
+	defer c.mutable.Unlock()
+
+	// Sanity check that next height is still as expected.
+	if c.mutable.nextHeight != block.BlockHeight {
+		return 0, errors.New("unexpected next height [BUG]")
 	}
-	if *c.prevHeight == height {
-		return c.currentOffset - 1, nil
+
+	// Get block offset to return, maybe increment next.
```
arajasek, line 306:
```suggestion
	// Get block offset to return, maybe increment nextOffset.
```

arajasek: just for clarity

lib/xchain/provider/provider.go
```diff
@@ -247,68 +244,83 @@ func (p *Provider) getEVMChain(chainID uint64) (netconf.Chain, ethclient.Client,
...
-}
-
-func newOffsetTracker(from uint64) *offsetTracker {
-	return &offsetTracker{
-		enabled:       true,
-		currentOffset: from,
+	attestInterval uint64
+	enabled        bool
+
+	// mutable defines all mutable fields protected by a RWMutex.
```
arajasek, line 252: Can we just use two atomic ints here? It would look a lot cleaner.

Are you concerned about a race between when nextOffset and nextHeight are incremented?

arajasek (approved):
Nits / suggestions, but fine to land as-is.

lib/xchain/provider/provider.go
corverroos: I prefer locks, it is much easier to reason about.

corverroos: with locks it feels more "obviously no bugs", while with atomic ints it feels like a race is still possible.

PR #1383, opened 2024-07-06
fix(e2e): lower parallelization when starting contracts

arajasek:
We group the edges by the source, and deploy in parallel across each source-group. However, for a specific source-group, we start them in series so as to avoid blitzing Fireblocks. 

issue: none


arajasek:
The entire pingpong starting process takes about 10 minutes now (with 4 chains). It used to take about 2 mins (with 3 chains, so significantly fewer operations).

corverroos (approved):

PR #1385, opened 2024-07-08
chore(halo/voter): improve reorg detection logs and metrics

corverroos:
Improve voter reorg detection logs and metrics and logic. 

Also add `height` log attr to abci context.

issue: none

### DIFF
--- a/halo/attest/voter/voter.go
+++ b/halo/attest/voter/voter.go
@@ -153,7 +153,7 @@ func (v *Voter) runForever(ctx context.Context, chainVer xchain.ChainVersion) {
 			return // Don't log or sleep on context cancel.
 		}
 
-		log.Warn(ctx, "Vote runner failed (will retry)", err)
+		log.Warn(ctx, "Vote runner failed (will retry)", err, "chain", v.network.ChainVersionName(chainVer))
 		backoff()
 	}
 }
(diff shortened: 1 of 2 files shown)


halo/attest/voter/voter.go
```diff
@@ -506,33 +513,35 @@ func (v *Voter) instrumentUnsafe() {
 	count(v.proposed, proposedCount)
 }
 
-// detectReorg returns an error if the previous block doesn't match the new block's parent hash.
-// This indicates that a reorg occurred.
-func detectReorg(chainVer xchain.ChainVersion, chainVerName string, prevBlock xchain.Block, block xchain.Block, cursors map[xchain.StreamID]uint64) error {
-	if prevBlock.BlockHash == (common.Hash{}) {
```
corverroos, line 512: this skipped omni consensus chain reorgs completely. We can still ensure height and message offsets are consecutive.

halo/attest/voter/voter.go
```diff
@@ -506,33 +513,35 @@ func (v *Voter) instrumentUnsafe() {
...
 	}
 
+	if block.BlockHash == (common.Hash{}) {
+		return nil // Skip consensus chain blocks without block hashes.
+	}
 	if prevBlock.BlockHash == block.ParentHash {
 		return nil // No reorg detected.
 	}
 
-	reorgTotal.WithLabelValues(chainVerName).Inc()
```
corverroos, line 534: this was only incremented for block hash reorgs, not stream offset or height issues.

pavelnikolov (approved):

PR #1387, opened 2024-07-08
chore(e2e/eoa): decrease large threshold within sane max

corverroos:
Decrease large eoa fund threshold to within sane max so that these accounts can be funded and alerts cleared.

issue: none


brid (approved):
LGTM

PR #1403, opened 2024-07-08
chore(contracts): separate avs into its own forge project

kevinhalliday:
Separate AVS into its own forge project.

This is a prerequisite to upgrading openzeppelin in the core. 

issue: #1310


pavelnikolov (approved):

contracts/core/script/genesis/AllocPredeploys.s.sol
```diff
@@ -158,7 +158,7 @@ contract AllocPredeploys is Script {
         bytes32 ownerSlot = bytes32(0);
 
         vm.store(Predeploys.ProxyAdmin, ownerSlot, bytes32(uint256(uint160(cfg.admin))));
-        vm.etch(Predeploys.ProxyAdmin, vm.getDeployedCode("out/ProxyAdmin.sol/ProxyAdmin.0.8.24.json"));
+        vm.etch(Predeploys.ProxyAdmin, vm.getDeployedCode("out/ProxyAdmin.sol/ProxyAdmin.json"));
```
ttarsi, line 161: can we confirm that this uses our `ProxyAdmin` and not Open Zeppelines?

ttarsi (approved):

contracts/core/script/genesis/AllocPredeploys.s.sol
kevinhalliday: good call yeah it should

kevinhalliday: confirmed it was - we only have a single ProxyAdmin artifact. But going to update the string we use here to be explicit to ours

kevinhalliday: though we will change when we upgrade oz - because they added owner to constructor

PR #1407, opened 2024-07-08
chore(contracts/core): upgrade open zeppelin to v5

kevinhalliday:
Upgrade to openzeppelin v5.

Things of note:

- oz v4 remains, to keep in in token source code, as that was what was deployed
- ~open zeppelin v5 deploys a new proxy admin for each transparent proxy. this 
  does not fit our use case. both in genesis, and for other deployments, as we want a 
  single proxy admin per chain. this, unfortunately, requires us copy and slightly modify
  their transparent proxy contract~ We've decided to use the new pattern given to use by 
  open zeppelin. This then requires changes in our e2e deployment flow - as we no longer 
  need a separate ProxyAdmin deploy flow / address management

issue: #1310


.github/workflows/soltest.yml
```diff
@@ -41,10 +41,10 @@ jobs:
...
       - name: Check fmt
         run: forge fmt --check
+      - name: Run tests
+        run: forge test -vv
       - name: Check snapshot
         # fuzz tests snapshots are different in CI, so we ignore them
         run: forge snapshot --check --tolerance 5 --no-match-test testFuzz_syncWithOmni
         env:
             FOUNDRY_FUZZ_SEED: "0x6eed"
-      - name: Run tests
```
kevinhalliday, line 49: Changed order so tests with logs run first (snapshot will fail if tests fail)

contracts/core/foundry.toml
```diff
@@ -20,7 +20,7 @@ fs_permissions = [
 ]
 
 allow_paths = ["../avs"]
-
+gas_limit = 9223372036854775807 # prev foundry default
```
kevinhalliday, line 23: Recent foundry nightly version changed gas limit it seems. Tests were failing

contracts/core/test/genesis/AllocPredeploys.t.sol
```diff
@@ -26,24 +30,56 @@ contract AllocPredeploys_Test is Test, AllocPredeploys {
...
+        // preinstalls
+        expected += 15;
+
+        // 4788 deployer account (nonce set to 1)
+        expected += 1;
+
+        // AllocPredeploys.deployer, account used to deploy temp proxies, which deploy ProxyAdmins left in genesis
+        // nonce left to ensure no new deployments conflict with genesis deployments - though this will likely never
+        // happen, as the deployer address does not have a known private key
+        expected += 1;
```
kevinhalliday, line 55: @ttarsi probably a could section to review carefully

contracts/core/foundry.toml
ttarsi: why were the tests failing / why is it okay to not use the new default?

contracts/Makefile
```diff
@@ -14,8 +14,7 @@ build: version ## Build contracts.
 
 CORE_CONTRACTS := OmniPortal FeeOracleV1 Create3 TransparentUpgradeableProxy \
 			Staking Slashing OmniBridgeL1 OmniBridgeNative Omni WOmni \
-			PortalRegistry AllocPredeploys PingPong \
-			core/src/deploy/ProxyAdmin.sol:ProxyAdmin \
+			PortalRegistry AllocPredeploys PingPong ProxyAdmin \
```
ttarsi, line 17: how/where do we specify that `ProxyAdmin` is v5 instead of v4

ttarsi (approved):
sheesh - is the +128k locs because we're not deploying proxyadmins for every contract?

contracts/Makefile
kevinhalliday: in core directory we only have v5 bindings

kevinhalliday:
sheesh - is the +128k locs because we're not deploying proxyadmins for every contract?

It's because we are deploying proxy admins for every proxy in the predeploy namspaces.

Yeah all the lines of change come down to additional allocs in `contracts/allocs/...` (also in test golden file)

We could choose to remove these from vcs, and instead generate them on command, but this would place requirement of having forge installed where you run that

contracts/core/foundry.toml
kevinhalliday: Tests were failing because they ran out of gas - not sure what they changed it too in new nightly release, it's not in docs. might be a bug, or they just decreased defaults

Issue #1314, opened 2024-06-27
feat(attest): add ListAllAttestations endpoint to query pending atts


ttarsi:
Add a `ListAllAttestations` query endpoint to attest module using index `status,chain_id,conf_level,block_offset`. The aim to be able to query pending attestations in order to debug state if required.

PR #1420, opened 2024-07-09
feat(attest): add ListAllAttestations query

arajasek:
Lists all attestations of the given status found. Expected to be mostly used for debugging.

issue: #1314

### DIFF
--- a/halo/attest/keeper/keeper.go
+++ b/halo/attest/keeper/keeper.go
@@ -367,31 +367,12 @@ func (k *Keeper) ListAttestationsFrom(ctx context.Context, chainID uint64, confL
 			}
 		}
 
-		pbsigs, err := k.getSigs(ctx, att.GetId())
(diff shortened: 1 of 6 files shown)


halo/attest/keeper/keeper.go
```diff
@@ -522,6 +503,41 @@ func (k *Keeper) earliestAttestation(ctx context.Context, version xchain.ChainVe
...
+
+	var resp []*types.Attestation
+	next := blockOffset
+	for iter.Next() {
+		att, err := iter.Value()
+		if err != nil {
+			return nil, errors.Wrap(err, "value")
+		}
+
+		if att.GetBlockOffset() != next {
```
arajasek, line 525: I copy-pasta-ed this from `listAttestationsFrom` above, but I'm actually not sure why we break here instead of failing?

PR #1422, opened 2024-07-10
chore(monitor/xchain): start populating emit cache from head-128

corverroos:
When populating the emitcursorcache, don't start from DeployHeight or LatestAttestation, since that can esily be more than 128 block old. Our fullnodes only store 128 blocks with of historical state, so just start there.

issue: none


brid (approved):
LGTM

pavelnikolov (approved):

pavelnikolov (approved):

PR #1425, opened 2024-07-10
fix(monitor): fix nil pointer in xmonitor

corverroos:
Replace incorrect `errors.Wrap` with `errors.New` causing nil pointer.

issue: none


brid (approved):
LGTM

Issue #1393, opened 2024-07-08
feat(e2e): add flows for portal admin actions


kevinhalliday:
Actions to include:

- pause portals
- unpause portals
- upgrade portals

PR #1429, opened 2024-07-10
feat(e2e): fireblocks rpc proxy

kevinhalliday:
Add a fireblocks rpc proxy that enables raw signing.

The proxy intercepts `eth_sendTransaction` requests, signs them,
and forwards an `eth_sendRawTransaction` request.

Prerequisite for running admin forge scripts (#1393)

issue: #1393

### DIFF
--- a/e2e/fbproxy/app/middleware.go
+++ b/e2e/fbproxy/app/middleware.go
@@ -21,10 +21,10 @@ type txSigner interface {
 //   - intercepts eth_sendTransaction requests, signs them `txsigner`
 //     and replaces them with eth_sendRawTransaction requests
 //   - leaves all other requests unmodified
-func NewSendTxMiddleware(txsigner txSigner, chainID *big.Int) Middleware {
-	// signer used to create signaure hash and recover tx sender,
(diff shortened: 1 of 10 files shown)


e2e/fbproxy/app/middleware.go
```diff
@@ -0,0 +1,78 @@
...
+
+type txSigner interface {
+	Sign(ctx context.Context, digest common.Hash, signer common.Address) ([65]byte, error)
+}
+
+// NewSendTxMiddleware returns a middleware func that
+//   - intercepts eth_sendTransaction requests, signs them `txsigner`
+//     and replaces them with eth_sendRawTransaction requests
+//   - leaves all other requests unmodified
+func NewSendTxMiddleware(txsigner txSigner, chainID *big.Int) Middleware {
```
kevinhalliday, line 24: Exported for tests - should we just keep private and use `_internal_test` pattern?

arajasek (approved):
Slightly weak approval here -- nothing seems horribly wrong, and it's a relatively simple change, but a second more competent review would be good.

corverroos (approved):

PR #1430, opened 2024-07-10
fix(monitor): correctly start fromHeight to start streaming from

arajasek:
For very new chains (in particular, the omni_evm and omni_consensus chains), the `latest` height is ~0, and we're underflowing.

issue: none

### DIFF
--- a/monitor/xmonitor/emitcursorcache.go
+++ b/monitor/xmonitor/emitcursorcache.go
@@ -67,7 +67,10 @@ func startEmitCursorCache(
 			return nil, errors.Wrap(err, "latest height", "chain", chain.Name)
 		}
 
-		fromHeight := latest - cacheStartLag
+		var fromHeight uint64
+		if latest > cacheStartLag {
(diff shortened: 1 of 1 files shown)


monitor/xmonitor/emitcursorcache.go
```diff
@@ -67,7 +67,10 @@ func startEmitCursorCache(
 			return nil, errors.Wrap(err, "latest height", "chain", chain.Name)
 		}
 
-		fromHeight := latest - cacheStartLag
+		var fromHeight uint64
+		if latest > cacheStartLag {
+			fromHeight = latest - cacheStartLag
+		}
```
pavelnikolov, line 73: Do we need a unit test for that?

pavelnikolov (approved):

monitor/xmonitor/emitcursorcache.go
arajasek: I don't really think so, no.

PR #1432, opened 2024-07-11
refactor(halo/attest): harden vote verification

corverroos:
Hardens vote verification in multiple places:
- Detect duplicate votes in `AggVote`
- Prevent naive local validator double signing 
- Harden `verifyAggVotes`:
  - Ensure vote extension limit is not exceeded
  - Ensure no duplicate aggregates
- Refactor sig DB table:
  - Remove unnecessary index
  - Add double sign detection index
- Check for double signing when inserting sigs into DB.
- Add a metric to detect any double signing (we need to know if this happens)

This addresses issues raised by Zellic and Sigma Prime audit.

issue: none


halo/attest/keeper/keeper.go
```diff
@@ -688,19 +710,27 @@ func (k *Keeper) VerifyVoteExtension(ctx sdk.Context, req *abci.RequestVerifyVot
 		log.Warn(ctx, "Rejecting invalid vote extension", err)
 		return respReject, nil
 	} else if !ok {
-		log.Debug(ctx, "Accepting nil vote extension") // This can happen if no non-empty blocks are present.
```
corverroos, line 691: this happens a lot, no need to log it.

arajasek (approved):

PR #1447, opened 2024-07-11
feat(e2e): add pause portal admin command

kevinhalliday:
Add e2e admin pause-portal command.
Also, set up boilerplate code to be used by other admin commands.

issue: #1393


e2e/app/admin/pauseportal.go
```diff
@@ -0,0 +1,71 @@
...
+
+	if err := pausePortalForge(ctx, c.rpc, s.admin, c.PortalAddress); err != nil {
+		return errors.Wrap(err, "pause portal")
+	}
+
+	return nil
+}
+
+func pausePortalForge(ctx context.Context, rpc string, sender common.Address, portal common.Address) error {
+	calldata, err := pausePortalABI.Pack("run", portal)
```
ttarsi, line 58: if we're using the ABI like this, rather than the script mode from forge, could we just have all of our admin actions in a single contract and share all the go boilerplate instead of having a separate go file per action?

ttarsi: oh nvm, this is just encoding args

ttarsi (approved):

e2e/app/admin/pauseportal.go
kevinhalliday: No but we def can combine everything into a single script was thinking that too

kevinhalliday: Just different entry points

PR #1455, opened 2024-07-12
feat(contracts): add pause portal admin script

kevinhalliday:
Add e2e admin pause-portal command.

DRAFT

This is a draft PR that implements the pause portal script outside the e2e runner.

This should be compared to https://github.com/omni-network/omni/pull/1447, which
implements this feature within the e2e runner.

Both offer boilerplate / setup code that will be used for future admin actions.


kevinhalliday:
Closing to prefer bash over go

PR #1457, opened 2024-07-12
ci(e2e): bump omega halo

corverroos:
Upgrade omega version by 1 commit to restart the halo instances to fix the stalled chain.

Also, don't upgrade jwtsecret since halo isn't always restarted on upgrade. Which caused the current chain stall.

issue: none


brid (approved):
LGTM

Issue #1329, opened 2024-07-01
chore(contracts): remove fast and safe


ttarsi:
Remove the `fast` and `safe` confirmation levels, as they are unused and currently only add confusion for devs

PR #1460, opened 2024-07-12
chore(contracts): remove fast and safe conf levels

kevinhalliday:
Remove fast and safe conf levels from ConfLevel lib.

issue: #1329


contracts/core/src/libraries/ConfLevel.sol
```diff
@@ -5,24 +5,15 @@ pragma solidity ^0.8.12;
...
  * @dev We prefer explicit constants over Enums, because we want uint8 values to start at 1, not 0, as they do in
- *      lib/xchain/types.go, such that 0 can represent "unset".
+ *      lib/xchain/types.go, such that 0 can represent "unset". Note only latest and finalized levels are supported
+ *      on-chain.
  */
 library ConfLevel {
     /**
      * @notice XMsg confirmation level "latest", last byte of xmsg.shardId.
      */
     uint8 internal constant Latest = 1;
```
ttarsi, line 15: can we maybe use an enum for this?

contracts/core/src/libraries/ConfLevel.sol
```diff
@@ -5,24 +5,15 @@ pragma solidity ^0.8.12;
...
-     * @notice XMsg confirmation level "fast", last byte of xmsg.shardId.
-     */
-    uint8 internal constant Fast = 2;
-
-    /**
-     * @notice XMsg confirmation level "safe", last byte of xmsg.shardId.
-     */
-    uint8 internal constant Safe = 3;
-
     /**
```
ttarsi, line 17: latest and finalized to 1 & 2?

ttarsi (approved):
approving with Qs

contracts/core/src/libraries/ConfLevel.sol
kevinhalliday: They are still 1 & 4 in lib/xchain/types.go

contracts/core/src/libraries/ConfLevel.sol
kevinhalliday: Enum are zero indexed and we start with one. And they always increment montonically. So the translation to lib/xchain/types.go values is just a little annoying. I'd rather just set constants explicitly

contracts/core/src/libraries/ConfLevel.sol
kevinhalliday: If we change them there I'm down

PR #1462, opened 2024-07-13
ci(github): add uint subtraction codeql query

corverroos:
Adds a `uint-subtract` CodeQL query to detect uint subtraction. Also add CodeQL action.

issue: none


corverroos:
Replaced by https://github.com/omni-network/omni/pull/1463

PR #1470, opened 2024-07-15
feat(e2e/admin): add upgrade portal action

kevinhalliday:
Add upgrade portal action. 

Additional:

- refactor to reduce code duplication and add tests

issue: #1393


contracts/core/script/admin/Admin.s.sol
```diff
@@ -11,21 +15,47 @@ import { IOmniPortalAdmin } from "src/interfaces/IOmniPortalAdmin.sol";
...
-     * @param portal The address of the portal contract.
+     * @param admin     The owner of the portal contract.
+     * @param portal    The address of the portal contract.
      */
-    function unpausePortal(address portal) public {
-        vm.startBroadcast();
-        IOmniPortalAdmin(portal).pause();
+    function unpausePortal(address admin, address portal) public {
+        vm.startBroadcast(admin);
+        OmniPortal(portal).unpause();
```
ttarsi, line 34: 👀

contracts/core/script/admin/Admin.s.sol
```diff
@@ -11,21 +15,47 @@ import { IOmniPortalAdmin } from "src/interfaces/IOmniPortalAdmin.sol";
...
+    function upgradePortal(address admin, address deployer, address portal, bytes calldata data) public {
+        // deploy new implementation
+        vm.startBroadcast(deployer);
+        address impl = address(new OmniPortal());
+        vm.stopBroadcast();
+
+        // upgrade proxy
+        vm.startBroadcast(admin);
+        address proxyAdmin = EIP1967Helper.getAdmin(portal);
+        ProxyAdmin(proxyAdmin).upgradeAndCall(ITransparentUpgradeableProxy(portal), impl, data);
```
ttarsi, line 54: would we ever use upgrade and call? or is that only used to initialize

ttarsi (approved):

e2e/fbproxy/app/proxy.go
```diff
@@ -58,11 +58,29 @@ func (p *proxy) Proxy(w http.ResponseWriter, r *http.Request) {
...
+	for {
+		if err == nil {
+			return false
+		}
+		if errors.Is(err, context.Canceled) {
+			return true
+		}
+		err = errors.Unwrap(err)
+	}
+}
```
corverroos, line 82: `errors.Is(err, context.Canceled)` should be sufficient, it does check the whole chain, no need to `Unwrap` yourself.

corverroos (approved):

contracts/core/script/admin/Admin.s.sol
kevinhalliday: gonna add some tests for these

PR #1510, opened 2024-07-16
ci(e2e): configure external advertised P2P IP

corverroos:
Use externalIP as "advertisingIP" for seeds and fullnodes on protected chains. The rest use internal IPs since they don't support external inbound connections.

issue: none


brid (approved):
LGTM

PR #1512, opened 2024-07-16
feat(contracts/core): admin controlled valset cutoff

kevinhalliday:
Make xsubValsetCutoff admin controlled state. 

issue: #1416


ttarsi (approved):
nice!

Issue #1465, opened 2024-07-15
emitcursorcache not populated errors


corverroos:
Problem to solve:
 - `monitor` service has lots of `emitcursorcache not populated`
 - This is causing loss of metrics and visibility and observability in staging/omega.

Proposed solution:
```
Okay, I think I have an explanation here. Tl;dr: the emitCursorCache needs to factor in the latest attestation heights when trimming.
Currently, we write every single attestation-worthy block into the emit cursor cache, and delete everything older than 10000 epochs from our latest block.
Our reads, however, are for whatever block the latest attestation corresponds to.
The latest attestation doesn’t _need_ to correspond to something in the past 10k epochs, though, and generally isn’t so for the omni consensus chain.
As a result, we just keep looking for a block that we’ve long since deleted.
I’m not 100% confident this is the cause of all the monitor issues we’re seeing, but it’s definitely causing some of them. Let’s fix it and see where we get.
```

PR #1513, opened 2024-07-17
fix(monitor/emitcache): store zero cursors

corverroos:
Store zero cursors to ditinguish between "not populated" and "nothing there yet". This should reduce errors by a lot.

issue: #1465


arajasek (approved):
Confused by the PR title -- the fix is in the monitor, not the cprovider?

PR #1518, opened 2024-07-17
feat(e2e): add support for testnet token and bridge

kevinhalliday:
Add support for deploying token & l1 bridge during omega deploy.

It also updates bridge deploy flow to support multiple testnet versions, and mainnet.
Though mainnet bridge deployment config is not introduced, and would therefore error.

issue: #1414

### DIFF
--- a/e2e/app/tokenbridge.go
+++ b/e2e/app/tokenbridge.go
@@ -31,11 +31,6 @@ func setupTokenBridge(ctx context.Context, def Definition) error {
 		return nil
 	}
 
-	if !def.Testnet.Network.IsEphemeral() {
-		log.Warn(ctx, "Skipping token bridge setup", errors.New("only ephemeral networks"))
-		return nil
-	}
-
 	omniEVM, ok := network.OmniEVMChain()
 	if !ok {
 		return errors.New("no omni evm chain")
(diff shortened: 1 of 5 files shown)


e2e/app/tokenbridge.go
```diff
@@ -60,21 +55,29 @@ func setupTokenBridge(ctx context.Context, def Definition) error {
...
+	tokenAddr, receipt, err := omnitoken.DeployIfNeeded(ctx, networkID, l1Backend)
 	if err != nil {
 		return errors.Wrap(err, "deploy omni token")
 	}
 
-	log.Info(ctx, "Deployed Omni Token", "chain", l1.Name, "addr", tokenAddr.Hex(), "block", receipt.BlockNumber)
+	if receipt != nil {
+		log.Info(ctx, "Deployed Omni Token", "chain", l1.Name, "addr", tokenAddr.Hex(), "block", receipt.BlockNumber)
+	} else if tokenAddr != networkID.Static().TokenAddress {
+		log.Warn(ctx, "Omni token already deployed, but not in network static", errors.New("missing static token addr"), "addr", tokenAddr.Hex())
```
arajasek, line 66: Wanna just fail the deployment in this case, since afaic tell it means we're doing something wrong?

lib/contracts/l1bridge/deploy.go
```diff
@@ -68,9 +68,26 @@ func getDeployCfg(network netconf.ID) (DeploymentConfig, error) {
...
 	}
 
+	if network == netconf.Omega {
+		return omegaCfg(), nil
+	}
+
 	return DeploymentConfig{}, errors.New("unsupported network", "network", network)
 }
 
+func omegaCfg() DeploymentConfig {
```
arajasek, line 78: Can we just have a static `map[netconf.ID]DeploymentConfig`? That will make a lot of this cleaner (fewer methods, no switching on the network ID in `getDeployCfg` above, etc.).

lib/contracts/l1bridge/deploy.go
```diff
@@ -97,6 +114,54 @@ func devnetCfg() DeploymentConfig {
...
+		return contracts.StagingL1Bridge(), true
+	case netconf.Devnet:
+		return contracts.DevnetL1Bridge(), true
+	default:
+		return common.Address{}, false
+	}
+}
+
+// IsDeployed returns true if the token contract is already deployed to its expected address.
+func IsDeployed(ctx context.Context, network netconf.ID, backend *ethbackend.Backend) (bool, common.Address, error) {
```
arajasek, line 133: Make private?

arajasek (approved):
Misc suggestions / nits, but nothing merge-blocking.

e2e/app/tokenbridge.go
kevinhalliday: Hmm. It may not necessarily be wrong. Might just be a reminder to go put it in static, cause you forgot to last time. Useful in case where omega deploy fails, but some contracts were deployed.

lib/contracts/l1bridge/deploy.go
kevinhalliday: Yeah think that's probably cleaner. 

These funcs (seen elsewhere in lib/contracts) were initially used to avoid global vars, cause linter didn't like. But we've removed that rule and use global vars in a number of places. I can clean up and introduce in a separate pr

PR #1522, opened 2024-07-18
chore(lib/contracts): cleanup deployment configs

kevinhalliday:
Cleanup deployment config code. 

Just use a mapping of network ID to config. 

issue: none

### DIFF
--- a/lib/contracts/l1bridge/deploy.go
+++ b/lib/contracts/l1bridge/deploy.go
@@ -59,60 +59,9 @@ func (cfg DeploymentConfig) Validate() error {
 	return nil
 }
 
-var configs = map[netconf.ID]DeploymentConfig{
-	netconf.Omega: {
-		Create3Factory:  contracts.OmegaCreate3Factory(),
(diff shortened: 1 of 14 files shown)


corverroos (approved):

lib/contracts/l1bridge/deploy.go
```diff
@@ -59,82 +59,9 @@ func (cfg DeploymentConfig) Validate() error {
...
-		Portal:          contracts.DevnetPortal(),
-		Token:           contracts.DevnetToken(),
-		ExpectedAddr:    contracts.DevnetL1Bridge(),
-	}
-}
-
-func AddrForNetwork(network netconf.ID) (common.Address, bool) {
-	switch network {
-	case netconf.Mainnet:
-		return contracts.MainnetToken(), true
```
corverroos, line 120: this was incorrect right?

kevinhalliday: It seems so - tests required perhaps.

corverroos: I'll add some

Issue #1384, opened 2024-07-06
fix(attest): don't delete finalized attestations needed by fuzzy chain


arajasek:
There's a dangerous edge-case in the deletion logic due to the inter-dependence of fuzzy and finalized chains. The deletion logic is careful not to delete the latest approved attestation for all chains. However, in an edge-case, the latest approved attestation for a fuzzy chain might be overridden by a finalized attestation that gets deleted. Consider the following case:

- We have a pending fuzzy attestation `A` for block offset 10
- A finalized attestation `B` for block offset 10 is approved, we update `A` to be "overridden" by `B`
- A  finalized attestation `C` for block offset 11 is approved
- Deletion logic runs (assume all attestations were created long enough ago to be considered for deletion)

The latest attestation offsets we will calculate are 10 for the fuzzy chain, and 11 for the finalized chain. As a result, we will not delete `A` (it's a fuzzy attestation with offset = 10), and will not delete `C` (it's a finalized attestation with offset = 11), but we WILL delete `B` (since, from the perspective of the finalized chain, it's stale). This is a bug, since `A` links to `B`, but `B` has now been deleted. It leads to crashes such as this:

```
4-06-25 17:13:44.263 ERRO Finalize req failed [BUG]                height=29 err="latest approved: get finalized attestation: not found" stacktrace="[errors.go:39 keeper.go:473 keeper.go:258 keeper.go:560 module.go:85 module.go:799 app.go:170 baseapp.go:792 abci.go:819 abci.go:884 cmt_abci.go:44 abci.go:94 local_client.go:185 app_conn.go:104 execution.go:213 state.go:1771 state.go:1682 state.go:1617 state.go:1655 state.go:2334 state.go:2066 state.go:929 state.go:836 asm_arm64.s:1222]"
24-06-25 17:13:44.263 ERRO error in proxyAppConn.FinalizeBlock      module=state err="latest approved: get finalized attestation: not found" stacktrace="[errors.go:39 keeper.go:473 keeper.go:258 keeper.go:560 module.go:85 module.go:799 app.go:170 baseapp.go:792 abci.go:819 abci.go:884 cmt_abci.go:44 abci.go:94 local_client.go:185 app_conn.go:104 execution.go:213 state.go:1771 state.go:1682 state.go:1617 state.go:1655 state.go:2334 state.go:2066 state.go:929 state.go:836 asm_arm64.s:1222]"
[...]

PR #1531, opened 2024-07-18
refactor(halo/attest): dont delete dependant attestations

corverroos:
Avoid deleting attestations with dependencies.

See [ADR](https://www.notion.so/omni-network/Keep-fewer-attestations-in-state-62f28f9367f24782a06af256c2348224#7c3ac622641543c2985a8e9c802fea79) for details.

No tests yet, will add them once we are happy with the solution direction.

issue: #1384


halo/attest/keeper/keeper.go
```diff
@@ -906,16 +903,37 @@ func (k *Keeper) deleteBefore(ctx context.Context, height uint64) error {
...
+		}
+
+		var fuzzyDepsFound bool
+		for _, fuzzyVer := range fuzzyDependents(att.XChainVersion()) {
+			// For finalized attestations (with fuzzy dependents), only delete them AFTER all fuzzy attestations have been deleted.
+			// This avoids deleting finalized overrides for current or future fuzzy attestations.
+			earliestFuzzy, ok, err := earliestOffset(ctx, fuzzyVer)
+			if err != nil {
+				return err
+			} else if !ok && att.GetBlockOffset() <= earliestFuzzy {
```
arajasek, line 938: I think the logic here is incorrect. I think you want:
- `||` instead of `&&`
- `att.GetBlockOffset() >= earliestFuzzy`

arajasek (commented):
I think this looks pretty good overall, and are good ways to meet the spec's restrictions on when it's not safe to delete an attestation.

- I'm definitely happy with the "earliest fuzzy" check, it's very neat.
- I'm not thrilled with the way we're handling the consensus chain / valset issue. It feels a little hacky, and I don't love leaning into the argument of "if we haven't delivered this in 10k epochs, we've got bigger problems". Having said that, I think it's reasonable, and I'm probably being a little overzealous.

halo/attest/keeper/keeper.go
corverroos: inverse yes

halo/attest/keeper/keeper.go
```diff
@@ -906,16 +903,37 @@ func (k *Keeper) deleteBefore(ctx context.Context, height uint64) error {
...
 		// Never delete anything after the last approved attestation offset per chain,
 		// even if it is very old. Otherwise, we could introduce a gap
 		// once we start catching up.
 		// Also, don't delete anything if we don't have an approved attestation yet (leave all pending attestations).
-		if latest, ok, err := getLatest(att.XChainVersion()); err != nil {
+		if latest, ok, err := latestOffset(ctx, att.XChainVersion()); err != nil {
 			return err
 		} else if !ok || att.GetBlockOffset() >= latest {
-			continue
+			continue // Skip deleting finalized attestation.
```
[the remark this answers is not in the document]
corverroos: Yeah, it would only be pendings

arajasek (approved):
LGTM, thank you!

PR #1533, opened 2024-07-18
fix(e2e/admin): return error on any failures

kevinhalliday:
Return error if any run on any fails. 

issue: none

### DIFF
--- a/e2e/app/admin/common.go
+++ b/e2e/app/admin/common.go
@@ -117,11 +117,15 @@ func getChains(network netconf.Network, chain string) []string {
 // They take a chain name as input, and return cli output.
 type runResults = forkjoin.Results[string, string]
 
-// report logs the results of a forkjoin, returning an error only if all runs failed.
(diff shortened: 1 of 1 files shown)


e2e/app/admin/common.go
```diff
@@ -131,15 +135,17 @@ func report(ctx context.Context, results runResults) error {
 		}
 	}
 
-	if len(failed) == len(results) {
+	if len(failed) == runs {
```
arajasek, line 138: Not clear what we're gaining by introducing `runs`, seems like `len(results)` was achieving the same thing?

arajasek (approved):

e2e/app/admin/common.go
kevinhalliday: `len(results)` was always zero. I assume cause it's something like an iterator that's been spent? though haven't confirmed

arajasek: (discussed offline)

ah, it's a channel, that makes sense

PR #1534, opened 2024-07-19
feat(lib/netconf): update omega version

brid:
bumping Omega version for a new deployment, and enable deployment

issue: none


lib/netconf/static.go
```diff
@@ -164,7 +164,7 @@ var statics = map[ID]Static{
 	},
 	Omega: {
 		Network:              Omega,
-		Version:              "v0.0.7",
+		Version:              "v0.0.8",
 		AVSContractAddress:   omegaAVS,
 		OmniExecutionChainID: evmchain.IDOmniOmega,
 		MaxValidators:        maxValidators,
```
corverroos, line 170: We need to also clear the portals: `Portals: nil,`

corverroos (changes requested):
Requesting changes since auto-merge enabled and we shouldn't merge this yet

corverroos (approved):

PR #1544, opened 2024-07-19
fix(cchain): sanity check binary search start index

arajasek:
This allows us to fail earlier and with a useful error type.

issue: none

### DIFF
--- a/halo/app/start.go
+++ b/halo/app/start.go
@@ -95,10 +95,8 @@ func Start(ctx context.Context, cfg Config) (<-chan error, func(context.Context)
 		return nil, nil, err
 	}
 
-	if cfg.Network != netconf.Simnet {
-		if err := enableSDKTelemetry(); err != nil {
-			return nil, nil, errors.Wrap(err, "enable cosmos-sdk telemetry")
-		}
+	if err := enableSDKTelemetry(cfg.Network); err != nil {
+		return nil, nil, errors.Wrap(err, "enable cosmos-sdk telemetry")
 	}
 
 	privVal, err := loadPrivVal(cfg)
@@ -311,7 +309,12 @@ func newEngineClient(ctx context.Context, cfg Config, network netconf.ID, pubkey
 }
 
 // enableSDKTelemetry enables prometheus based cosmos-sdk telemetry.
-func enableSDKTelemetry() error {
+func enableSDKTelemetry(id netconf.ID) error {
+	// Skip telemetry for simnet, because it uses globals which conflict when running tests in parallel.
+	if id == netconf.Simnet {
+		return nil
+	}
+
 	const farFuture = time.Hour * 24 * 365 * 10 // 10 years ~= infinity.
 
 	_, err := sdktelemetry.New(sdktelemetry.Config{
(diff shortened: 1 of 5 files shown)


lib/cchain/provider/provider.go
```diff
@@ -155,6 +155,8 @@ func (p Provider) Subscribe(in context.Context, chainVer xchain.ChainVersion, xB
 	}()
 }
 
+var ErrHistoryPruned = errors.New("no commit info found (history pruned)")
```
corverroos, line 158: could we maybe add a comment please?

lib/cchain/provider/abci.go
```diff
@@ -373,7 +373,17 @@ func searchOffsetInHistory(ctx context.Context, client rpcclient.Client, cl atyp
 		endHeightIndex = queryHeight
 		lookback *= 2
 		if queryHeight < lookback {
-			startHeightIndex = 1
+			earliestStoreHeight, err := getEarliestStoreHeight(ctx, cl, chainVer, 1)
```
corverroos, line 376: what about rather setting `queryHeight=1` and looping back to `for queryHeight > 0 {`?

no additional logic required in that case, we rerun existing logic.

corverroos: Also nit: `for queryHeight > 0` is always true, not sure that condition is actually every triggered? can just do `for {` instead?

arajasek: Yup, both of those seem reasonable to me, thanks!

halo/app/start.go
```diff
@@ -95,8 +95,10 @@ func Start(ctx context.Context, cfg Config) (<-chan error, func(context.Context)
 		return nil, nil, err
 	}
 
-	if err := enableSDKTelemetry(); err != nil {
-		return nil, nil, errors.Wrap(err, "enable cosmos-sdk telemetry")
+	if cfg.Network != netconf.Simnet {
```
arajasek, line 98: I had to do this in order to not get the test suite to fail when it's run (individual tests are fine). Any better suggestions?

corverroos (approved):
Nice, love that test!

lib/cchain/provider/abci.go
```diff
@@ -373,11 +373,11 @@ func searchOffsetInHistory(ctx context.Context, client rpcclient.Client, cl atyp
 		endHeightIndex = queryHeight
 		lookback *= 2
 		if queryHeight < lookback {
-			startHeightIndex = 1
-			break
+			// Query from the start, but don't break out yet -- we need to find the earliest height that we have state for
+			queryHeight = 1
+		} else {
+			queryHeight -= lookback
 		}
```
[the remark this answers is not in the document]
arajasek: Good catch, thank you. I don't think we want to use 0, since that defaults to using the `latest` height. So I think what we want is actually:

```
		if queryHeight <= lookback {
			// Query from the start, but don't break out yet -- we need to find the earliest height that we have state for
			queryHeight = 1
		} else {
			queryHeight -= lookback
		}
```

and we can do this at the top of the loop.

Issue #1433, opened 2024-07-11
Encode merkle leaves using EIP-712


corverroos:
We currently use abi encoding of our merkle tree leaves.

Different leaf types (XBlockHeader vs XMsg) can result in the same leaf hash if their structure and data matches. This introduces an attack surface.

This was raised by the Sigma Prime audit can they suggest encoding leaves using EIP-712 which uses domain separation by type to avoid this issue.

corverroos:
From Sigma Prime

There's an interesting edge case for OMP-01 [#1252](https://github.com/omni-network/omni/pull/1252) . The merkle trees are no longer enforced to have the block header in index 0 of the tree. That is because the merkle tree [hashPair()](https://github.com/omni-network/omni/blob/b7745f13ba5cb5846da7cb69ebbfec125726acad/lib/merkle/core.go#L201) will sort the leaves and then perform hashing. This is the correct behaviour as it matches OZs Solidity implementation.
However, there is a side effect, this isn't exploitable at present but would be very unpleasant if it is. The simplest case would be if you have an xblock with a single message and block header and the goal is to switch them. This would only work if you could have an valid encoding of a block header match a valid encoding of a message and vice versa. Currently a XTypes.BlockHeader encoding is fixed as 4 slots and XTypes.Msg is at least 8 slots, so it would not be possible to have a blockheader be a valid encoding of message.

If we were to somehow create a valid encoding in the future, say the fields of block header change. Then the attack vector would be to switch the blockheader and messages in the tree. Thus, interpret a blockheader as a message and message as a blockheader.

The recommendation is to add some domain separation when hashing leaves. On-chain could look like

```
    function _msgLeaves(XTypes.Msg[] calldata msgs) private pure returns (bytes32[] memory) {
        bytes32[] memory leaves = new bytes32[](msgs.length);

        for (uint256 i = 0; i < msgs.length; i++) {
            leaves[i] = _leafHash(abi.encode(0x02, msgs[i])); // @audit add 0x02 domain separator, should be a named constant
        }

        return leaves;
    }

    function _blockHeaderLeaf(XTypes.BlockHeader calldata blockHeader) private pure returns (bytes32) {
        return _leafHash(abi.encode(0x01, blockHeader)); // @audit 0x01 domain separator, better to use a named constant rather than 0x01
    }
 ```
 
Alternatively, you could make functions _leafHeaderHash() and _leafMsgHash() which add domain separation.

corverroos:
[...]

PR #1545, opened 2024-07-19
feat(contracts): include DSTs when hashing leaves

arajasek:
Adds one-byte DSTs to our hashing, so far using 1 for XBlockHeaders and 2 for XMsgs.

Note that this does not implement full EIP-712, because it's hard to shoehorn what we need into that format (eg. what's the Chain ID to use?).

issue: #1433


lib/xchain/merkle.go
```diff
@@ -7,6 +7,11 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 )
 
+const (
+	DomainSeparationTagBlockHeader merkle.DomainSeparationTag = 1 + iota
+	DomainSeparationTagMessage
```
corverroos, line 12: lets explicitly number these enums. I suggest to only use `iota` for in-memory enums. When persisting enums to disk, it is better to be super explicit about there values, this avoid accidental changing of values due reordering or deprecation. Also improves lookup. Don't need to count in your head.

TL;DR: only use iota if the actual number doesn't matter.

corverroos: Can maybe make names more concise, especially since the type is already full qualified.
```
const (
    DSTUnknown  merkle.DomainSeparationTag = 0
    DSTBlockHeader merkle.DomainSeparationTag = 1
    DSTMessage merkle.DomainSeparationTag = 2
)
```

corverroos (approved):

contracts/core/src/libraries/XBlockMerkleProof.sol
```diff
@@ -9,6 +9,10 @@ import { XTypes } from "./XTypes.sol";
  * @dev Library for verifying XBlock merkle proofs
  */
 library XBlockMerkleProof {
+    uint8 internal constant DST_XBLOCK_HEADER = 1;
+
+    uint8 internal constant DST_XMSG = 2;
+
```
kevinhalliday, line 15: Let's add some natspec comments here.

Also thinking - these aren't exaclty "domain separators" (is DST == domain separator? or domain separator type?)  They are kinda like the type hashes used in EIP-712 domain separators. But they also aren't type hashes. 

Is there another name we can give these? like `_TYPE_ID`? or maybe `_HASH_PREFIX`?

kevinhalliday (commented):
Just need to update the ts test data generator

contracts/core/src/libraries/XBlockMerkleProof.sol
arajasek: Comments added.

I think these are domain separators (DST = domain separation tag), unless I'm very mistaken.

kevinhalliday: I guess my understanding of the semantics of "domain separators" was that they were more about verifying chain & contract. but types are a domain in themselves. so this works

kevinhalliday (approved):

PR #1559, opened 2024-07-23
ci(lib/anvil): swap anvil for anvilproxy

corverroos:
Instead of using `anvil:latest` use our own `anvilproxy:main` that we know if working and should be faster to pull 🤞 .

This aims to fix failing CI which is timing out downloading `anvil:latest`.

issue: none


corverroos:
This is failing in e2e for some reason and original error seems to have gone away, so closing for now

Issue #1543, opened 2024-07-19
Replay Protection: include consensus chain ID in siggys


arajasek:
We should consider including the consensus chain ID in signatures. This would serve as replay protection between our different networks, though that would only matter if we had the same validator address on 2 different networks.

kirk-baird:
# The issue 

This is a good point. There should be some sort of domain separation that prevents signature replays on different networks. To do this we need to sign over something that is unique to the entire network. Then we need to validate the unique tag is correct during signature validation.

If there is a fork or some reason we have multiple `OmniPortal` contracts operating on a single chain they will both produce `XBlocks`. Any validator that is using the same key for both instances of `OmniPortal` can have their signature replayed on the other network. 

# My proposed solution is:

a) Include `omniChainId` in each `XBlockHeader`. Then add validation in the smart contracts in `xsubmit()` and consensus code for each signature verification step. Reject signatures with a different `omniChainId`. 

This should work as `omniChainId` would be unique per network fork including testnets and mainnets.

c) Discourage the use of validator keys on multiple networks wherever possible.

PR #1567, opened 2024-07-23
feat(*): include consensus chain ID in XBlockHeaders

arajasek:
This gets signed as part of the Attestation, and verified in the Portal Contract.

issue: #1543


contracts/core/src/xchain/OmniPortalStorage.sol
```diff
@@ -46,7 +46,7 @@ abstract contract OmniPortalStorage is IOmniPortal, IOmniPortalAdmin {
     uint64 public omniChainId;
 
     /**
-     * @notice Virtual chain ID used in xmsgs from Omni's consensus chain
+     * @notice Chain ID of Omni's consensus chain
```
kevinhalliday, line 49: Maybe we can leave "virtual"? or something like it

the "virtual" wording was here because it's not actually the chain id of the consensus chain. consensus chain is a cosmos chain, and does not have an evm chain id. We use a uint representation of the Omni's consensus chain  id for consensus xblocks.

halo/attest/types/tx.proto
```diff
@@ -54,11 +54,12 @@ message Vote {
 
 // BlockHeader uniquely identifies a cross chain block/attestation.
 message BlockHeader {
-  uint64 chain_id    = 1; // Chain ID as per https://chainlist.org
-  uint32 conf_level  = 2; // Confirmation level (aka version) of the cross-chain block/attestation.
-  uint64 offset      = 3; // Offset of the cross-chain block
-  uint64 height      = 4; // Height of the source-chain block
-  bytes  hash        = 5; // Hash of the source-chain block
+  uint64 source_chain_id      = 1; // Source Chain ID as per https://chainlist.org
+  uint64 consensus_chain_id   = 2; // Consensus Chain ID as per https://chainlist.org, for replay protection.
```
kevinhalliday, line 58: I wouldn't add the chainlist.org bit here.

It won't be there. We don't get to reserve this chain ID cause it's not an evm chain. Someone else could reserve. And we wouldn't be able to support them

kevinhalliday (approved):
Some comments on comments. The rest looks good

corverroos (commented):
It doesn't look like the current implementation verifies the vote/attestation stored in the consensus chain. Malicious validators can therefore create votes with invalid consensus chain IDs. This should be prevented.

Suggestions:
 - Add `cChainID` to `halo/attest/types#BlockHeader.Verify` (this is probably safest and most explicit, but cascades a bit).
 - Other option is to add adhoc checks in attest keeper. This has higher risk of bypassing the verification, but has smaller footprint.

halo/attest/types/tx.proto
arajasek: I made this change in light of https://github.com/omni-network/ops/issues/198, but it's not super relevant to this PR anyway. I'll simplify for now regardless.

lib/xchain/abi_internal_test.go
```diff
@@ -42,11 +42,12 @@ func TestEncodeHeader(t *testing.T) {
...
-		SourceChainID: 1,
-		ConfLevel:     1,
-		BlockOffset:   100,
-		BlockHeight:   99, // Ignored
-		BlockHash:     common.HexToHash("0x412d62a6a3115ab5a0e0cae9d63082ff8dfb002a98cc889d06dc986a9461586b"),
+		SourceChainID:    1,
+		ConsensusChainID: 1,
+		ConfLevel:        1,
+		BlockOffset:      100,
+		BlockHeight:      99, // Ignored
```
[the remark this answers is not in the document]
arajasek: Making the change.

arajasek:
> It doesn't look like the current implementation verifies the vote/attestation stored in the consensus chain. Malicious validators can therefore create votes with invalid consensus chain IDs. This should be prevented.

Is this worth doing? I'm not sure what it achieves -- a malicious validator can always lie about things, so I don't really see any concern here. 

I guess if over 2/3rds  of validator power is participating on two Omni chains, we might attest to blocks from the other chain?

contracts/core/src/xchain/OmniPortal.sol
```diff
@@ -173,6 +173,7 @@ contract OmniPortal is
         XTypes.BlockHeader calldata xheader = xsub.blockHeader;
         uint64 valSetId = xsub.validatorSetId;
 
+        require(xheader.consensusChainId == omniCChainId, "OmniPortal: wrong consensus chain");
```
kevinhalliday, line 176: Can we make this reason string one character smaller? It's at 33 we want 32 so it fits into a single 32 byte slot.

halo/attest/types/tx.proto
kevinhalliday: I mean the "Consensus chain as per http://chainlist.org"

contracts/core/src/xchain/OmniPortal.sol
kevinhalliday: And let's add a test case

halo/attest/types/tx.proto
arajasek: Yeah, I believe we're planning on registering the consensus chain ID with chainlist too ([thread](https://omni-chain.slack.com/archives/C02HCJN6KNE/p1721318167863919?thread_ts=1721249907.530149&cid=C02HCJN6KNE)), but regardless dropped from this PR.

contracts/core/src/xchain/OmniPortal.sol
arajasek: Done, and done! Thank you!

PR #1568, opened 2024-07-23
fix(lib): endpoint HeaderByNumber might return nil

arajasek:
The result of a call to `eth_getBlockByNumber` can be `nil`, with a `nil` error. This most often happens when calling the `Pending` block, but it can happen in other situations (see [here](https://github.com/ethereum/go-ethereum/blob/ed8fd0ac0919ee954cfa621d16a5d0cfbf30c96d/core/blockchain.go#L929)). geth [checks for the returned block ](https://github.com/ethereum/go-ethereum/blob/2e8e35f2ada60e8e90aff8f6374ea7cd7da16116/ethclient/ethclient.go#L217-L219)being `nil` before returning, throwing an error in this case. We should do the same.

This PR refactors things to use geth's `HeaderByNumber` method, which will perform the check for us.

issue: none

### DIFF
--- a/lib/ethclient/ethclient.go
+++ b/lib/ethclient/ethclient.go
@@ -2,6 +2,7 @@ package ethclient
 
 import (
 	"context"
+	"math/big"
 
 	"github.com/omni-network/omni/lib/errors"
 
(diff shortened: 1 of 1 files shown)


lib/ethclient/ethclient.go
```diff
@@ -96,20 +97,18 @@ func (w Wrapper) HeaderByType(ctx context.Context, typ HeadType) (*types.Header,
...
+	var bn rpc.BlockNumber
+	if err := bn.UnmarshalJSON([]byte(typ.String())); err != nil {
+		return nil, errors.Wrap(err, "unmarshal head type")
+	}
+
+	header, err := w.cl.HeaderByNumber(ctx, big.NewInt(int64(bn)))
 	if err != nil {
 		incError(w.chain, endpoint)
-		return nil, errors.Wrap(err, "get block")
+		err = errors.Wrap(err, "json-rpc", "endpoint", endpoint)
```
arajasek, line 108: following the pattern in the generated wrappers for consistency

corverroos: nice

corverroos (approved):
LGTM 🚀

Issue #1541, opened 2024-07-19
Can't join omega network


corverroos:
Joining omega using `omni operators init-nodes` fails with the attached logs. 

To reproduce:
```
make build-halo-relayer
go install github.com/omni-network/omni/cli/cmd/omni && omni operator init-nodes --network=omega --moniker=foo --clean
cd ~/.omni/omega
docker compose up
```

corverroos:
[Uploading logs.txt…]()

PR #1600, opened 2024-07-29
ci(e2e): improve seed node p2p config

corverroos:
Improve seed node p2p again:
 - Add normal persisted peers (since staging had a blip where seednode didn't have any nodes)
 - Prepopulate addressbook with persisted peers.

issue: #1541

### DIFF
--- a/e2e/app/definition.go
+++ b/e2e/app/definition.go
@@ -238,12 +238,6 @@ func adaptNode(ctx context.Context, manifest types.Manifest, testnet *e2e.Testne
 		node.Seeds = append(node.Seeds, testnet.LookupNode(seed))
 	}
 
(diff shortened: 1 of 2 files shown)


e2e/app/definition.go
```diff
@@ -238,12 +238,6 @@ func adaptNode(ctx context.Context, manifest types.Manifest, testnet *e2e.Testne
 		node.Seeds = append(node.Seeds, testnet.LookupNode(seed))
 	}
 
-	if node.Mode == types.ModeSeed {
```
corverroos, line 241: this resulted in staging seed node struggling to find peers. It even had a time where it didn't have any nodes. This is bad.

ttarsi (approved):

PR #1606, opened 2024-07-30
ci(e2e): disable seedmode

corverroos:
Disable `seedmode` since it doesn't seem to work well in small networks. This aims to fix joining staging issues.

issue: #1541


brid (approved):
LGTM

PR #1610, opened 2024-07-30
ci(e2e): add staging static seeds

corverroos:
Staging only has a single seednode, add the fullnode to the static list of seed so that we at least have 2 nodes to work with. Sometimes, node addressbooks are empty if they only have outgoing peers.

Also add long-lived static keys for staging nodes (aligns staging with omega).

Also add basic omni_evm monitoring in halo: syncing, peers, height.

Also, when joining, configure seeds as persisted peers, that is more robust for small networks.

issue: #1541


brid (approved):
LGTM

Issue #1526, opened 2024-07-18
Improve halo metrics


corverroos:
While debugging issues on staging we identified the following metrics to improve monitoring:

1. Is Halo connected to EVM?
- 1/0 whether the halo instance is connected to engine API.
2. Is Halo a validator?
- 1/0 whether the halo instance is a validator

PR #1611, opened 2024-07-30
chore(halo): monitor comet syncing

corverroos:
Adds logs and metrics for halo/geth syncing, also geth peers. 

Don't initialise voter until the node is a validator. This avoid issues with partial rpc endpoint config in non-validators.

issue: #1526


brid (approved):
LGTM

PR #1613, opened 2024-07-30
chore(halo/genutil): add testnet faucet allocs

kevinhalliday:
Add testnet faucet allocs.

Fund 10 accounts, derived from a single faucet mnemonic.

issue: none


brid (approved):
LGTM

PR #1615, opened 2024-07-30
fix(e2e): relayer should use an archive node

arajasek:
The relayer has to use an archive node after #1612, else it might not be able to fetch necessary attestations.

issue: none

### DIFF
--- a/e2e/manifests/devnet0.toml
+++ b/e2e/manifests/devnet0.toml
@@ -1,7 +1,10 @@
-# Devnet0 is a tiny devnet. Only a single validator and two anvils.
+# Devnet0 is a tiny devnet. Only a single validator, an archive node (for the relayer), and two anvils.
 # This is aimed at local dev environments.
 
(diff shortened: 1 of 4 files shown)


e2e/manifests/devnet0.toml
```diff
@@ -1,7 +1,10 @@
-# Devnet0 is a tiny devnet. Only a single validator and two anvils.
+# Devnet0 is a tiny devnet. Only a single validator, an archive node (for the relayer), and two anvils.
 # This is aimed at local dev environments.
 
 network = "devnet"
 anvil_chains = ["mock_op", "mock_arb"]
 
 [node.validator01]
+
+[node.fullnode01]
```
arajasek, line 9: We always need an archive node now for a meaningful network.

kevinhalliday (approved):

brid (approved):
Makes sense to me

Issue #1591, opened 2024-07-26
turn off emit cursor cache and point relayer to archive halo


ttarsi:
*Problem to Solve*

- emit cursor cache is buggy

*Proposed Solution*

- turn it off and point relayer to archive node

PR #1628, opened 2024-08-01
chore(monitor/emitcache): fallback to archive node

corverroos:
Workaround cache not populated errors by falling back to querying (archive) nodes. Also add metrics with the aim of debugging and fixing the emitcache issues in the long run.

Also refactor docker to use `docker compose` instead of `docker-compose` which isn't supported on github runners anymore it seems.

issue: #1591


brid (approved):
LGTM

Issue #1631, opened 2024-08-01
Voter to use consensus registry instead of evm registry


corverroos:
*Problem to Solve*
The voter (and other processes) uses `netconf.AwaitOnChain` that queries the evm portal registry for supported chains. While cosmos onchain verification uses the consensus chain registry module supported chains. They are not identical, with consensus chain registry lagging by 1 block. This causes voter to submit votes too early, resulting in "check supported chain: missing conf levels" failures.

*Proposed Solution*
Refactor `netconf.AwaitOnChain` to use the consensus chain registry instead.
Instead of `*bindings.PortalRegistry` use the cprovider that queries `registry` module state.

arajasek:
This requires adding some missing fields to the registry module, and is therefore now consensus critical.

PR #1469, opened 2024-07-15
feat(contracts): add attestInterval, blockPeriod, and name to registry

kevinhalliday:
Add attestInterval, blockPeriod, and name to PortalRegistry.

Update netconf provider accordingly.

Also addresses #1446.

Additional:
- update some comments

issue: #1323

PR #1633, opened 2024-08-01
fix(halo): add missing fields to registry

arajasek:
Also use it when loading the voter to avoid raciness.

issue: #1631


lib/netconf/provider.go
```diff
@@ -113,6 +163,40 @@ func networkFromPortals(ctx context.Context, network ID, portals []bindings.Port
...
+
+		chains = append(chains, Chain{
+			ID:            portal.ChainId,
+			Name:          namer(portal.ChainId),
+			PortalAddress: common.BytesToAddress(portal.GetAddress()),
+			DeployHeight:  portal.DeployHeight,
+			// TODO: How can I get this info?
+			BlockPeriod: 0,
+			Shards:      toShardIDs(portal.GetShardIds()),
+			// TODO: How can I get this info?
```
arajasek, line 186: @corverroos Any ideas here?

corverroos: ah damn, when we added there fields to the execution portal registry contract, [PR](https://github.com/omni-network/omni/pull/1469), we forgot to also add them to the consensus portal registry. We should add them since they are consensus critical 🤦

arajasek:
Okay, this is now ready for review and merge.

halo/registry/keeper/registry.proto
```diff
@@ -23,4 +23,7 @@ message Portal {
   bytes           address            = 2; // Portal contract address
   uint64          deploy_height      = 3; // Height this portal contract was deployed at
   repeated uint64 shard_ids          = 4; // Shard IDs supported by this portal
+  uint64          attest_interval    = 5; // The interval, in blocks, at which validators must attest, even if empty
+  uint64          block_period       = 6; // The block period of the chain deployed to, in milliseconds.
```
corverroos, line 27: common practice is to add the unit as a suffix, like `_ms`?

corverroos (approved):

halo/registry/keeper/registry.proto
arajasek: Done, thanks.

Issue #1415, opened 2024-07-09
add retries to avoid relatively benign WARN logs


arajasek:
Problem: there are 2 aspects of the system that can generate WARN logs a little too zealously. These are the Xchain Provider's `FetchBatch` method, which may fail due to being very slightly behind another node, and the relayer's send method, which might have submissions be reverted on very dynamic chains due to out of gas (eg. [here](https://omniops.grafana.net/goto/8fhASflIg?orgId=1)). In both cases, the components upstream will retry anyway, and likely succeed, but still emit a failure log / metric.

Solution: given that this is relatively benign, let's retry a few times within these methods before erroring.

PR #1634, opened 2024-08-02
feat(xchain): retry FetchBatch a few times before erroring

arajasek:
XBlock fetcher tries 5 times (with 100 ms pauses) before giving up and returning an error.

issue: #1415


lib/xchain/provider/provider.go
```diff
@@ -150,14 +150,21 @@ func (p *Provider) stream(
 				ConfLevel: req.ConfLevel,
 			}
 
-			xBlock, exists, err := p.GetBlock(ctx, fetchReq)
-			if err != nil {
-				return nil, err
-			} else if !exists {
-				return nil, nil
+			var lastErr error
+			retryCount := 5
```
corverroos, line 154: can make this a inline `const`.

lib/xchain/provider/provider.go
```diff
@@ -150,14 +150,21 @@ func (p *Provider) stream(
...
+			for i := 0; i < retryCount; i++ {
+				xBlock, exists, err := p.GetBlock(ctx, fetchReq)
+				if err != nil {
+					lastErr = err
+					time.Sleep(100 * time.Millisecond)
```
corverroos, line 159: we could also stick to `backoff := expbackoff.New(ctx, expbackoff.WithPeriodic(time.Milli*100)` then all our backoff logic follows the same pattern. Benefit is that sleep is aborted when context is cancelled.

corverroos (approved):

Issue #1629, opened 2024-08-01
Add pinning for Relayer and Monitor in deployments


brid:
Ensure Relayer and Monitor are pinnable in the same way Halo is

PR #1635, opened 2024-08-02
feat(e2e): allow the pinning of relayer and monitor

brid:
Lets us pin the versions of relayer and monitor, in addition to halo.

Renames "omniTag" to "haloTag" as it feels like a generic naming feels less relevant now we have the ability to pin most of our applications specifically.

A follow up PR in the ops repo adds the modifications to the GH action deploy command.

issue: #1629


e2e/manifests/staging.toml
```diff
@@ -3,6 +3,8 @@ public_chains = ["op_sepolia","base_sepolia","arb_sepolia"]
 anvil_chains = ["slow_l1"]
 multi_omni_evms = true
 prometheus = true
+pinned_relayer_tag = "2a0a9bb"
```
corverroos, line 6: We shouldn't pin staging, can just use latest in all cases

e2e/cmd/flags.go
```diff
@@ -20,7 +20,9 @@ func bindDefFlags(flags *pflag.FlagSet, cfg *app.DefinitionConfig) {
 	flags.StringVar(&cfg.DeployKeyFile, "deploy-key", cfg.DeployKeyFile, "path to deploy private key file")
 	flags.StringVar(&cfg.FireAPIKey, "fireblocks-api-key", cfg.FireAPIKey, "FireBlocks api key")
 	flags.StringVar(&cfg.FireKeyPath, "fireblocks-key-path", cfg.FireKeyPath, "FireBlocks RSA private key path")
-	flags.StringVar(&cfg.OmniImgTag, "omni-image-tag", cfg.OmniImgTag, "Omni docker images tag (halo, relayer). Defaults to working dir git commit.")
+	flags.StringVar(&cfg.HaloImgTag, "halo-image-tag", cfg.HaloImgTag, "Halo docker image tag. Defaults to working dir git commit.")
+	flags.StringVar(&cfg.RelayerImgTag, "relayer-image-tag", cfg.RelayerImgTag, "Relayer docker image tag. Defaults to working dir git commit.")
```
corverroos, line 24: don't think we need flag. Pinning is only needed vi manifest (omega.toml).

cli/cmd/devnet.go
```diff
@@ -177,7 +177,9 @@ func devnetDefinition(ctx context.Context) (app.Definition, error) {
 		imgTag = "main"
 	}
 
-	defCfg.OmniImgTag = imgTag
+	defCfg.HaloImgTag = imgTag
+	defCfg.RelayerImgTag = imgTag
+	defCfg.MonitorImgTag = imgTag
```
corverroos, line 182: lets stick to existing two different patterns:
 - `omniTag` is overrride for all containers, specified in flags.
 - `pinned_*_tag` is specified in manifest only.

corverroos (changes requested):

e2e/manifests/staging.toml
brid: Yep. Only added that so we could test it. But after that has been verified we should remove it.

PR #1642, opened 2024-08-05
fix(contracts): fix native bridge total supply

kevinhalliday:
Fix OmniBridgeNative total supply allloc.

issue: none


kevinhalliday:
Lot of bindings changed - investigating. Maybe someone else updated in a PR with diff version of forge. 

This should be locked before public omega deploy. Will handle

corverroos (approved):

PR #1649, opened 2024-08-06
ci(e2e): pin monitor and relayer images

corverroos:
Support pinning of monitor and relayer images.

issue: https://github.com/omni-network/omni/issues/1629


brid (approved):
LGTM

PR #1653, opened 2024-08-06
chore(halo): instrument validator status

corverroos:
Adds the `halo_comet_validator` constant gauge metric that tracks whether the local halo node is a cometBFT validator (1) or not (0). Se should add an alert if any `validator` halo instance is not a validator.

issue: #1526


brid (approved):
LGTM

Issue #1626, opened 2024-07-31
Fix e2e log pusher


ttarsi:
*Problem to Solve*

When `e2etest.yml` and `pr-e2etest.yml fail, we have a step to push logs to loki. That steps fails:

<img width="742" alt="Screenshot 2024-07-31 at 4 07 10 PM" src="https://github.com/user-attachments/assets/698ff555-95c3-4c25-b350-c696aed990c6">

*Proposed Solution*

Either fix this or remove it (not sure we've ever used it)

PR #1654, opened 2024-08-06
chore(github): remove pushing e2e logs to loki

corverroos:
Remove github action step that pushes e2e logs to loki since it has never worked for a long time.

issue: #1626


brid (approved):
LGTM

Issue #1595, opened 2024-07-28
Create an XAppUpgradeable type


ttarsi:
*Problem to Solve*

The current XApp contract doesn't follow the initialize or general upgradeability / storage alloc pattern (this is fine, it's not designed to be upgradeable)

*Proposed Solution*

Following OpenZeppelin's naming patterns (alongside their upgradeability patterns) create `XAppUpgradable`

PR #1679, opened 2024-08-08
feat(contracts): add an XAppUpgradeable contract

kevinhalliday:
Add an XAppUpgradeable contract

Additionally, add XApp and XAppUpgradeable tests.
Mark each as abstract.
Fix small bug in MockPortal.

issue: #1595


ttarsi (approved):
Love it

Issue #1701, opened 2024-08-13
fix account monitor frequent alerts


ttarsi:
#### Problem to Solve

Account monitor is frequently firing. This means that one or more of our accounts are frequently running out of funds. This means that either 1) that account's funding level is set too low, resulting in frequent alerts or 2) that account's funding level is fine, and something is wrong with how much it's spending.

#### Proposed Solution

Collect more data:
- which accounts are running out of funds
- on which chains?
- is it consistently the same chain/account?

corverroos:
Staging chain stalled due to `old val set` issue since we didn’t fund it. I’m going to reduce self-delegation to allow more “relayer downtime”. Also going to increase relayer funding from 20 to 50.

PR #1702, opened 2024-08-13
ci(e2e): address relayer funds issue

corverroos:
Relayer on staging is running out of 20ETH balance fast: 
 - Increase relayer funding from 20 to 50
 - Decrease parallel ping pongs from 10 to 3
 - Increase self-delegation period from 1H to 6H (allowing more relayer downtime before `old valset` issue is triggered).

issue: #1701


brid (approved):
LGTM

PR #1706, opened 2024-08-13
feat(monitor/xfeemngr): manage fee params for chain not in network

kevinhalliday:
Support management of onchain fee params for a chain not in the network.

Testnet sepolia chains post tx data to sepolia. Their xcall fees should therefore 
be calculated using sepolia gas prices. Sepolia is not in the omega network, but we
need to manage it's fee params on chain.

Requires https://github.com/omni-network/ops/pull/385

issue: none

### DIFF
--- a/lib/evmchain/evmchain.go
+++ b/lib/evmchain/evmchain.go
@@ -20,14 +20,14 @@ const (
 	// Testnets.
 	IDOmniOmega   uint64 = 164
 	IDHolesky     uint64 = 17000
+	IDSepolia     uint64 = 11155111
 	IDArbSepolia  uint64 = 421614
 	IDOpSepolia   uint64 = 11155420
 	IDBaseSepolia uint64 = 84532
 
 	// Ephemeral.
 	IDOmniEphemeral uint64 = 1651
-	IDMockL1Fast    uint64 = 1652
-	IDMockL1Slow    uint64 = 1653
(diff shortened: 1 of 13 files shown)


kevinhalliday:
Currently not setting fee params for chains not in network on deploy. So we'd need to wait for xfeemngr to set it before sending xcalls to sepolia l2s. TBD if this would be an issue for e2e deploy, it may cause ping pong starts to fail. Staging deploy will tell us. If it's an issue we can add. Maybe worth adding anyways

lib/evmchain/evmchain.go
```diff
@@ -102,64 +102,61 @@ var static = map[uint64]Metadata{
...
 	},
-	IDMockL1Fast: {
-		ChainID:     IDMockL1Fast,
+	IDMockL1: {
+		ChainID:     IDMockL1,
 		Name:        "mock_l1",
 		BlockPeriod: time.Second,
 		NativeToken: tokens.ETH,
 	},
-	IDMockL1Slow: {
```
kevinhalliday, line 138: We weren't using slow l1 anywhere so i removed

corverroos (changes requested):
Just blocking this until omega either deployed or until we decide to delay omega until this is properly tested and confirmed on staging.

corverroos (approved):

PR #1707, opened 2024-08-13
fix(e2e): reduce admin script timeout risk

kevinhalliday:
Reduce timeout risk in admin forge scripts

- Increase fbproxy write timeout to allow for mpc fireblocks signing.
- Run admin scripts in sequence, not parallet

issue: none


kevinhalliday:
Requires https://github.com/omni-network/ops/pull/385

corverroos (approved):

Issue #1731, opened 2024-08-19
Incorrect dedup logic in attest verifyAggVotes


corverroos:
#### Problem to Solve
The attestmodule `verifyAggVotes` has incorrect dedup logic which resulted in all validator proposing the same valid blocks that are rejected by this logic. This resulted in the chain stalling.

```
msg="Rejecting process proposal" err="execute message: verify votes: invalid duplicate aggregate votes"
```

The agg votes are for the attest header, but for different blocks (fuzzy votes). This is valid.

#### Proposed Solution
Fix the logic and add tests

PR #1732, opened 2024-08-19
fix(halo/attest): fix duplicate agg vote logic

corverroos:
Fixes attest module `verifyAggVotes` dedup logic to align with `aggregateVotes` logic that uses `attestation root` as dedup key. 

issue: #1731


brid (approved):

arajasek:
LGTM, thanks for the test.

PR #1753, opened 2024-08-22
chore(xchain): support querying submission from tx

corverroos:
Adds support for `GetSubmission` to xprovider with the aim of querying AttestConfLevel for submissions using the originating submission since we are going to remove `Receipt.ConfLevel` since it isn't supported by portals.

issue: none


corverroos:
See #1755

kevinhalliday (approved):

PR #1755, opened 2024-08-22
chore(xchain): remove unused conf level from receipt

corverroos:
Removes `Receipt.ConfLevel` as this was never used and isn't supported by portals.

issue: none


corverroos:
See https://github.com/omni-network/omni/pull/1753

kevinhalliday (approved):

Issue #1686, opened 2024-08-09
Allow users to exchange ETH for gas on OMNI


kevinhalliday:
*Problem To Solve*

We ask devs to build apps that maintain global state on omni. To allow users to act on that state, they need native OMNI. The only way to get this is via the L1 native bridge, which is slow and expensive.

*Proposed Solution*

A pragmatic xchain exchange on each supported rollup that allows users to exchange small amounts of ETH for gas on Omni. We can use the FeeOracleV1 contracts, which track OMNI <> ETH conversion rates. This exchange should have safeguards to avoid misuse - such as global rate limits, and a potential percent haircut on conversion.

PR #1756, opened 2024-08-22
feat(contracts): allow xapps to swap eth for gas on omni

kevinhalliday:
Allow xapps to swap eth for gas on omni.

Components:
- OmniXFund: contract that pays out OMNI on Omni
- OmniGasPump: contract on each rollups that accepts ETH payments, and directs OmniFund to pay out
- XGasPump: abstract contract that can be used to extend you xapp with helpers to fund accounts on Omni

issue: https://github.com/omni-network/omni/issues/1686


contracts/core/src/token/OmniExchange.sol
```diff
@@ -0,0 +1,209 @@
...
+     */
+    mapping(address => uint256) public swapped;
+
+    constructor() {
+        _disableInitializers();
+    }
+
+    function initialize(address fund_, address portal, address owner) external initializer {
+        _setFund(fund_);
+        _setMaxSwap(0.02 ether);
```
ttarsi, line 59: should avoid hardcoded params like this, since changing the default will change the bytecode

contracts/core/src/token/OmniExchange.sol
```diff
@@ -0,0 +1,209 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity 0.8.24;
+
+import { OwnableUpgradeable } from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import { XAppUpgradeable } from "src/pkg/XAppUpgradeable.sol";
+import { FeeOracleV1 } from "src/xchain/FeeOracleV1.sol";
+import { ConfLevel } from "src/libraries/ConfLevel.sol";
+import { OmniFund } from "./OmniFund.sol";
+
+contract OmniExchange is XAppUpgradeable, OwnableUpgradeable {
```
ttarsi, line 10: description here on where this is deployed would be nice, same with OmniFund

contracts/core/src/token/OmniFund.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity 0.8.24;
+
+import { ConfLevel } from "src/libraries/ConfLevel.sol";
+import { XAppUpgradeable } from "src/pkg/XAppUpgradeable.sol";
+import { OwnableUpgradeable } from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+
+contract OmniFund is XAppUpgradeable, OwnableUpgradeable {
```
ttarsi, line 8: pausability in case of emergencies

ttarsi (commented):
Let's consider emergencies, how would we halt this if we

kevinhalliday:
Yes do not you worry - still a WIP (saying this here even tho we already discussed in slack to save face in the open source community)

contracts/core/src/examples/Funder.sol
```diff
@@ -0,0 +1,44 @@
...
+
+import { XApp } from "src/pkg/XApp.sol";
+import { XGasPump } from "src/pkg/XGasPump.sol";
+import { ConfLevel } from "src/libraries/ConfLevel.sol";
+
+/**
+ * @title Funder
+ * @notice Example contract that shows how to use XGasPump
+ */
+contract Funder is XApp, XGasPump {
```
kevinhalliday, line 12: Going to remove this example before mering

ttarsi: why? examples are good!

contracts/core/src/token/OmniGasPump.sol
```diff
@@ -0,0 +1,204 @@
...
+    event MaxSwapSet(uint256 max);
+    event FundSet(address fund);
+    event PctCutSet(uint256 pct);
+    event Refilled(address indexed recipient, uint256 paid, uint256 fee, uint256 cut, uint256 amtOMNI);
+
+    /// @notice Gas limit passed to OmniXFund.settleUp xcall
+    uint64 internal constant SETTLE_GAS = 100_000;
+
+    /// @notice Denominator for pct cut
+    uint256 internal constant PCT_DENOM = 1000;
```
ttarsi, line 25: 1_000

contracts/core/src/token/OmniGasPump.sol
```diff
@@ -0,0 +1,204 @@
...
+
+    /// @notice Address of OmniXFund on Omni
+    address public fund;
+
+    /// @notice Max amt (in native token) that can be swapped in a single tx
+    uint256 public maxSwap;
+
+    /// @notice Percentage cut (over PCT_DENOM) taken by this contract for each swap
+    ///     Used to disencentivize spamming
+    uint256 public pctCut;
```
ttarsi, line 35: don't love this name, it's not very descriptive imo

what about deduction or fee?

contracts/core/src/token/OmniXFund.sol
```diff
@@ -0,0 +1,61 @@
...
+        __Ownable_init(owner);
+    }
+
+    /**
+     * @notice Settle up with a recipient. If `total` is more than what they've been
+     *         funded, transfer the difference.
+     * @param recipient     Address to receive the funds
+     * @param total         Total (historical) amount requested for `recipient`
+     */
+    function settleUp(address recipient, uint256 total) external xrecv whenPaused {
```
ttarsi, line 35: `whenNotPaused`?

contracts/core/src/token/OmniXFund.sol
```diff
@@ -0,0 +1,61 @@
...
+        require(isXCall() && authed[xmsg.sourceChainId][xmsg.sender], "OmniFund: unauthorized");
+        require(total >= funded[recipient][xmsg.sourceChainId], "OmniFund: already funded");
```
ttarsi, line 37: what about only allowing one call per source chain? users should really only use this once, if they're using it multiple times its a red flag

ttarsi: eh i'll take this back, a user may want to fund their wallet if they run out

contracts/core/src/token/OmniGasPump.sol
```diff
@@ -0,0 +1,204 @@
...
+
+    struct InitParams {
+        address fund;
+        address portal;
+        address owner;
+        uint256 maxSwap;
+        uint256 pctCut;
+    }
+
+    function initialize(InitParams calldata p) external initializer {
```
ttarsi, line 52: I think we should probably keep global params (i.e. pctCut) on Omni, rather than on rollups. Global state! Only keep parameters in 1 place if possible

Max swap makes sense to keep here ofc, and the others are required. But no need to track pct cut in many places

kevinhalliday: Not a bad call. 

Thought part of the design is that this `OmniXFund` can be used for any app we have that ends up funding people with OMNI. OmniGasPump would just be one of them. We can add more by just `auth`ing new contracts

But then again, it's hard to build for the future. Sometimes better to just solve the problem at hand and nothing more

contracts/core/src/token/OmniXFund.sol
kevinhalliday: Yeah I think multiple times is good cause we're doing small swap amounts

contracts/core/src/token/OmniGasPump.sol
kevinhalliday: - fee is tricky cause we use if for xcall stuff
- IMO deduction means the same thing as cut

I agree I don't  _love_ the name. But it's not really user facing. 

Maybe toll?

contracts/core/src/examples/Funder.sol
kevinhalliday: guess it doesn't hurt

contracts/core/src/token/OmniGasPump.sol
kevinhalliday: Sorry did mention that this was a design goal anywhere ^

kevinhalliday: Also doing all the cuts before coming to a total "owed" value on the source chain is useful for the  accounting we do

contracts/core/src/token/OmniGasPump.sol
kevinhalliday: Form fmt does not request underscores for this so gonna leave it

contracts/core/src/token/OmniGasPump.sol
ttarsi: Toll is great

contracts/core/src/token/OmniGasPump.sol
ttarsi: Ehh I think we should not try to optimize for something we don't know about. You can make the call here, maybe you have something in mind that I don't. My only request is that you consider maintenance simplicity, I have not thought through what is necessary in each case, so will leave that to you

kevinhalliday: If we do make this "not a generic fund" we can rename it to OmniGasStation which I've been wanting to do

kevinhalliday: Wrapping this up. (copying from slack for posterity)

Tried moving toll to the singleton contract on Omni. Unfortunately it does make the book keeping trickier.

Because we're working with "total" amounts - total owed vs total settled - moving the toll to the gas station means that changing the toll changes the toll for all historical swaps.

kevinhalliday: TLDR: going to keep toll on gas pumps

contracts/core/src/token/OmniGasPump.sol
```diff
@@ -0,0 +1,225 @@
...
+     *
+     *      To retry (if OmniGasStation transfer fails), call swap() again with the
+     *      same `recipient`, and msg.value == swapFee().
+     *
+     * @param recipient Address on Omni to send OMNI to
+     */
+    function fillUp(address recipient) public payable whenNotPaused returns (uint256) {
+        // take xcall fee
+        uint256 f = xfee();
+        require(msg.value >= f, "OmniGasPump: insufficient fee");
```
ttarsi, line 88: can be strictly greater than, because if it's equal, they receive nothing

contracts/core/src/token/OmniGasPump.sol
```diff
@@ -0,0 +1,225 @@
...
+     * @param toll      Toll taken by this contract, denominated in ETH
+     * @param amtOMNI   Amount of OMNI swapped for
+     */
+    event FillUp(address indexed recipient, uint256 owed, uint256 amtETH, uint256 fee, uint256 toll, uint256 amtOMNI);
+
+    /// @notice Gas limit passed to OmniGasStation.settleUp xcall
+    uint64 internal constant SETTLE_GAS = 100_000;
+
+    /// @notice Denominator for percentage calculations (1000 = 100%)
+    uint256 internal constant PCT_DENOM = 1000;
```
ttarsi, line 40: `TOLL_DENOM` now?

contracts/core/src/token/OmniGasPump.sol
```diff
@@ -0,0 +1,225 @@
...
+        uint256 f = xfee();
+        require(msg.value >= f, "OmniGasPump: insufficient fee");
+        uint256 amtETH = msg.value - f;
+
+        // check max
+        require(amtETH <= maxSwap, "OmniGasPump: over max");
+
+        // take toll
+        uint256 t = amtETH * toll / PCT_DENOM;
+        amtETH -= t;
```
ttarsi, line 96: could collapse into 1 line

ttarsi (approved):
Left a bunch of nits but awesome stuff LGTM

contracts/core/src/token/OmniGasStation.sol
```diff
@@ -0,0 +1,86 @@
...
+        uint256 settled = fueled[recipient][xmsg.sourceChainId];
+
+        // If already settled, revert
+        require(owed >= settled, "Station: already funded");
+
+        // Transfer the difference
+        (bool success,) = recipient.call{ value: owed - settled }("");
+
+        // Update books. We do not bother doing so pre-transfer, because isXCall() prevents reentrancy
+        if (!success) fueled[recipient][xmsg.sourceChainId] = owed;
```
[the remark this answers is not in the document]
kevinhalliday: yes - tests should flush more of these things out when we add

contracts/core/src/token/OmniGasPump.sol
kevinhalliday: no we want to allow retry, where `msg.value == fee`

contracts/core/src/token/OmniGasPump.sol
kevinhalliday: We need to keep `t` as a variable to use later

contracts/core/src/token/OmniGasPump.sol
kevinhalliday: shore

contracts/core/src/token/OmniGasStation.sol
```diff
@@ -0,0 +1,86 @@
...
+     * @param recipient Address
+     * @param chainID   ChainID of the pump
+     * @param owed      Total amount owed to `recipient` by way of `chainID`'s OmniGasPump
+     * @param fueled    Total amount sent to `recipient` by this contract.
+     * @param success   True if the transfer was successful (if true, owed == fueled)
+     */
+    event SettleUp(address indexed recipient, uint64 indexed chainID, uint256 owed, uint256 fueled, bool success);
+
+    /// @notice Emitted when a OmniGasPump is set for a chain
+    event PumpSet(uint64 indexed chainID, address addr);
```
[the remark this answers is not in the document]
kevinhalliday: shore

PR #1758, opened 2024-08-23
feat(contracts): sync l1 bridge balance on each withdraw

kevinhalliday:
Sync l1 bridge balance with native bridge on each withdraw.

This will allow us to fund accounts at genesis, and account for this in the total supply but 
transfering tokens to the l1 bridge contract. The native bridge genesis balance will need to
be decreased for every other account we fund at genesis.

Additional:

- refactor to use `encodeCall` instead of `encodeWithSelector`

issue: none


contracts/core/src/token/OmniBridgeNative.sol
```diff
@@ -83,15 +84,19 @@ contract OmniBridgeNative is OwnableUpgradeable {
...
-    function withdraw(address payor, address to, uint256 amount) external {
+    function withdraw(address payor, address to, uint256 amount, uint256 l1Balance) external {
         XTypes.MsgContext memory xmsg = omni.xmsg();
 
         require(msg.sender == address(omni), "OmniBridge: not xcall"); // this protects against reentrancy
         require(xmsg.sender == l1Bridge, "OmniBridge: not bridge");
         require(xmsg.sourceChainId == l1ChainId, "OmniBridge: not L1");
 
-        l1BridgeBalance += amount;
+        l1BridgeBalance = l1Balance;
```
kevinhalliday, line 99: This is really the whole fix (plus the corresponding change on the l1 bridge)

The rests is just tests / comments / refactoring out encodeWithSelector

Issue #1537, opened 2024-07-19
Add backwards compatibility e2e test


corverroos:
Problem to be solved:
 - We need to know when we introduce backwards incompatible changes.
 - We need to know what versions/commits are backwards compatible with `main`.
 - We need to know when `main` breaks backwards compatibility with `omega` pinned version.
 
 Proposed solution:
  - Add a new e2e manifest that uses omega pinned version for half the validators (on startup).
  - Upgrade them to `current` in the e2e test (see cometBFT logic for upgradeVersion)

PR #1762, opened 2024-08-26
ci(e2e): add backwards compatibility test

corverroos:
Adds a `backwards.toml` that ensures backwards compatibility of `main` with `v0.4.0` our first omega LTS version. 

issue: #1537


ttarsi (approved):
Nice, only Q is should this be `v0.5.0` instead of `v0.4.0`?

kevinhalliday (approved):

PR #1773, opened 2024-08-26
chore(docs): rewrite build docs

kevinhalliday:
Rewrite build docs.

Goals:

- lean more "chain abstracted" than "cross chain app"
- have main example reflect value prop
- less is more (remove showcase for now)

issue: none


ttarsi (approved):
Love it! Simple + precise

Issue #1759, opened 2024-08-23
Allow pausing of native bridge


kevinhalliday:
*Problem To Solve*

Currently, the native bridge is not pausable. We should be able to pause it.

*Proposed Solution*

Add pausability (standard contracts stuff). Optionally pausing withdraws / deposits individually.

PR #1785, opened 2024-08-28
feat(contracts): allow pausing bridge contracts

kevinhalliday:
Add pausability to native bridge.

Allow pausing withdraws & claims separately

Additional:
- some file / constant name cleanup

issue: #1759


contracts/core/src/token/OmniBridgeL1.sol
```diff
@@ -14,7 +16,7 @@ import { OmniBridgeNative } from "./OmniBridgeNative.sol";
  * @notice The Ethereum side of Omni's native token bridge. Partner to OmniBridgeNative, which is
  *         deployed to Omni's EVM.
  */
-contract OmniBridgeL1 is OwnableUpgradeable {
+contract OmniBridgeL1 is OwnableUpgradeable, PausableUpgradeable, OmniBridgeCommon {
```
ttarsi, line 19: `OmniBridgeCommon` is both `OwnableUpgradeable` and `PausableUpgradeable`, so we shouldn't need to include those inheritances here as well

ttarsi (commented):
Nice, only feedback is that solidity inheritance is bleh so it could be worth considering just duplicating a little code rather than have 2 layers of inheritance

ttarsi (approved):

contracts/core/src/token/OmniBridgeL1.sol
kevinhalliday: We don't need to but I like showing it explicitly here

kevinhalliday: eh guess it looks nice with just one - ill think about your top level comment too

Issue #1442, opened 2024-07-11
test unjailing in e2e


corverroos:
(task logged by KH and migrated from Asana)

corverroos:
Tested manually by joining staging with inactive validator, then unjailing a few times.

PR #1790, opened 2024-08-29
feat(cli): add operator unjail command

corverroos:
Adds a `omni operator unjail` command that must be called by the operator-private-key.

issue: #1442


kevinhalliday (approved):
For both this and create-validator, we might want some check for success / failure on the consensus chain. We might need some cosmos event that maps to evm txns.

corverroos:
> For both this and create-validator, we might want some check for success / failure on the consensus chain. We might need some cosmos event that maps to evm txns.

I'll add some initial verification that would ensure that the tx succeeds.

PR #1804, opened 2024-09-03
chore(*): rename BlockPeriod to reflect nanoseconds

arajasek:
The value we set in the e2e app is in nanoseconds, so let's rename the fields to reflect this.

issue: none


kevinhalliday (approved):

corverroos (approved):

corverroos:
Maybe improve the places where we convert between uint64 and time.Time by using something like this?
```
periodNs, err := umath.ToUint64(period.Nanoseconds())
```
and
```
period = time.Duration(periodNs) * time.Nanosecond
```

Issue #1814, opened 2024-09-04
Add e2e cmd for validator allowlist


kevinhalliday:
## Problem to Solve

We need an e2e command (and corresponding forge script) to update the validator allowlist, so that we can allow validators in omega.

## Proposed Solution

The current Staking.sol contract does not allow for idempotent configuration. It only allows adding / removing validators, with little visibility into current validator set. I propose:

- updating Staking.sol, so that it's config can be idempotent
- add an `e2e admin staking upgrade` command, so we can upgrade Staking.sol 
- add static config that lives in monorepo, which we can update and sync on chain with an `e2e admin staking configure`

PR #1815, opened 2024-09-04
feat(e2e): add staking admin commands

kevinhalliday:
Add Staking.sol admin commands.

- rewrite Staking.sol to allow for idompentent config managment
- add upgrade script, so that we can upgrade Staking.sol on Omega
- add idompentent configure script
- add e2e cmd to test all admin commands

Additional:

- namespace admin commands by contract type
- refactor all admin commands, with less restrictive abstractions
- move admin command code to e2e/app root

issue: #1814


e2e/app/admin_staking.go
```diff
@@ -0,0 +1,165 @@
...
+	contract, err := bindings.NewStaking(common.HexToAddress(predeploys.Staking), client)
+	if err != nil {
+		return errors.Wrap(err, "new staking contract")
+	}
+
+	cfg, ok := staking.ConfigByNetwork(s.network.ID)
+	if !ok {
+		log.Info(ctx, "No static staking config, skipping", "network", s.network.ID)
+		return nil
+	}
```
kevinhalliday, line 80: Right now just logging and returning nil. Though we can configure these for devnet / staging & use them in our e2e flow. Right now just do it off of [genesis alloc accounts](https://github.com/omni-network/omni/blob/main/e2e/app/portalregistry.go#L265).

Another thing I need to add is how to tests - devnet will just deploy what's in binary. And this command uses what's in binary. So need to add an override for tests.

contracts/core/src/octane/Staking.sol
```diff
@@ -85,39 +110,50 @@ contract Staking is OwnableUpgradeable {
...
+        _allowlist = validators;
+
+        // set new
         for (uint256 i = 0; i < validators.length; i++) {
-            isAllowedValidator[validators[i]] = false;
+            isAllowedValidator[validators[i]] = true;
         }
+
+        emit AllowlistSet(validators);
     }
```
kevinhalliday, line 158: Changes here amount to:

- replace `disableAllowlist / enableAllowlist` with `setAllowlistEnabled`
- replace  `allowValidators / disallowValidators` with `setAllowlist`

ttarsi:
Closing for #1823

PR #1818, opened 2024-09-05
feat(cometbft): mitigate slow blocks due to inactive validators

corverroos:
Decreases default cometBFT consensus config param `timeout_propose` from 3s to 1s.

This aim to mitigate slow blocks when the proposing validator is offline/inactive. Currently, when 1 of 3 validators are offline, the block times on staging decreases from 1s to 4s.

I had a look at some other cosmos chains and aligning timeout_commit and timeout_propose seems to be common practice:
 - [Sei](https://github.com/sei-protocol/sei-chain/blob/main/app/params/config.go#L98-L103): `timeout_commit=200ms, timeout_propose=300s`
 - [Moonbridge](https://services.moonbridge.team/testnet/kopi/install/) and [kopi](https://docs.kopi.money/nodes.html#setting-up-a-node): `timeout_commit=1s, timeout_propose=1s`

This change aims to mitigate the affect of offline validators on block times.

issue: none or #12345


corverroos:
See block times slow down when 1of3 validators were offline in staging
image
image

brid (approved):

PR #1879, opened 2024-09-16
fix(contracts/core): avoid var name shadowing

kevinhalliday:
Avoid shadowing storage var with local var name.

issue: #1875


contracts/core/src/xchain/OmniPortal.sol
```diff
@@ -366,17 +366,17 @@ contract OmniPortal is
 
         uint64 totalPower;
         XTypes.Validator memory val;
-        mapping(address => uint64) storage valSet = valSet[valSetId];
+        mapping(address => uint64) storage set = valSet[valSetId];
```
ttarsi, line 369: Approving but don't love `set`, since that's an overloaded term, maybe consider underscores

ttarsi (approved):

PR #1883, opened 2024-09-16
fix(contracts): fix typos

kevinhalliday:
Additional:
- rename XReceipt error to err

issue: #1875


ttarsi (approved):

contracts/core/src/xchain/OmniPortalStorage.sol
```diff
@@ -77,13 +77,13 @@ abstract contract OmniPortalStorage is IOmniPortal, IOmniPortalAdmin {
     mapping(uint64 => mapping(uint64 => uint64)) public outXMsgOffset;
 
     /**
-     * @notice Offset of the last outbound XMsg that was sent to destChainId in shardId
-     *         Maps destChainId -> shardId -> offset.
+     * @notice Offset of the last inbout XMsg that was sent from sourceChainId in shardId
```
Saw-mon-and-Natalie, line 80:
```suggestion
     * @notice Offset of the last inbound XMsg that was sent from sourceChainId in shardId
```

PR #1914, opened 2024-09-18
feat(monitor): add contract balance monitoring

kevinhalliday:
Add contract balance monitoring.

issue: none

### DIFF
--- a/monitor/contract/monitor.go
+++ b/monitor/contract/monitor.go
@@ -1,0 +1,93 @@
+package contract
+
+import (
+	"context"
+	"time"
+
+	"github.com/omni-network/omni/lib/contracts"
+	"github.com/omni-network/omni/lib/ethclient"
+	"github.com/omni-network/omni/lib/log"
+	"github.com/omni-network/omni/lib/netconf"
+
+	"github.com/ethereum/go-ethereum/params"
+)
+
+// StartMonitoring starts the monitoring goroutines.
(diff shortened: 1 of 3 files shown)


monitor/contract/monitor.go
```diff
@@ -0,0 +1,93 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			err := monitorContractOnce(ctx, contract, chainName, client)
+			if ctx.Err() != nil {
+				return
+			} else if err != nil {
+				log.Error(ctx, "Monitoring contract failed (will retry)", err)
```
corverroos, line 59: We try to use WARN for stuff that retries, and only use ERROR for stuff that can't retry and require operator intervention

monitor/contract/monitor.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// monitorContractOnce monitors contract for the given chain.
+func monitorContractOnce(
+	ctx context.Context,
+	contract contracts.WithFundThreshold,
+	chainName string,
+	client ethclient.Client,
+) error {
+	balance, err := client.BalanceAt(ctx, contract.Address, nil)
```
corverroos, line 74: can use `EtherBalanceAt` that returns `float64` in ether units

corverroos (approved):

monitor/contract/monitor.go
kevinhalliday: we do use big.Int balance for threshold cmp tho so i'll keep it

Issue #1787, opened 2024-08-29
Align vote window and trim lag


corverroos:
#### Problem to Solve
We currently have a `+64 to -64` vote window. This allows votes to be add on-chain 64 attest offsets before and after the latest approved attestation per stream. This prevents spamming the attest module state with far future or far historical votes.

We currently have a `trimLag` of 1, which means that attestations become eligible for deletion 1 block after being approved (note some other conditions also apply).

This means that attestations could be approved, then deleted, then re-created. Recreated attestations will never be approved, since we only approved the next offset after latest approved offset, and recreated attestations will always be before latest approved offset. So it should not have a negative effect on the system, it merely introduces noise.

But this still isn't great, since it allows for strange/unexpected on-chain behaviour, and it makes duplicate vote detection hard and it makes voter reward and penalty calculation also hard. 

Ideally, duplicate vote detection and voter reward and penalty calculation will be performed for some specific attestation once before deletion, and after deletion, it shouldn't be a concern anymore. 

#### Proposed Solution
- Add negative vote window as a deletion logic condition.
- So attestation shouldn't become eligible for deletion until they are out of the vote window (-64 offsets behind latest approved offset). 
- Consider increasing the negative vote window form -64 to -16 or -8. This primary depends on how much we want to reward (or not) late votes (after quorum).

> Note this will require a `network-upgrade`

PR #1949, opened 2024-09-23
feat(halo/attest): only delete after vote window

corverroos:
Wait for attestations to exit vote window before deleting. 

issue: #1787


halo/attest/keeper/keeper.go
```diff
@@ -1134,16 +1132,6 @@ func windowCompare(voteWindow uint64, mid uint64, x uint64) int {
 	return 0
 }
 
-// uintSub returns a - b if a > b, else 0.
-// Subtracting uints can result in underflow, so we need to check for that.
-func uintSub(a, b uint64) uint64 {
```
corverroos, line 1139: use `umath` package instead

Issue #1951, opened 2024-09-23
FeeOracleV2


kevinhalliday:
*Problem to Solve*

Our current fee oracles do effectively keep us "in the green" (charging more in fees than our relayer spends on gas). They could, however, use some improvements.

**Data / non-evm pricing** 

FeeOracleV1 uses a "postsTo" mechanism to deteremine the data gas price for an xmsg. It uses the configured gas price of the "postsTo" chain to charge data fees for an xcall. This means that calls to L2s use the execution gas price of L1. Most L2s now post tx data as blobs, and blobs have an independent fee market. We need to update the fee oracles and xfeemngr to configure and charge xcall fees based on blob gas prices.

Additionally, our pricing is pretty evm specific. We should consider what it would look like to include a non-evm chain. We may well be able to "translate" non-evm chains into our exisiting evm-esque parameters. But we should consider what this would look like.

**Overheads**

Sigma prime suggested some "overheads" - for data costs / gas prices.

**Requires active management**

The current fee orocle contracts and monitor/xfeemngr do a lot of on chain transactions - can we reduce that cost? We want to remain "in the green". We also do not want to overcharge fees too much.  

*Proposed Solution*

FeeOracleV2 was intrduced, but is unused. This version uses explicit `dataGasPrice`, rather than `postsTo` mechanism.

This may be all we need. Though it may not. The owner of this task should read the sigma prime report, as well as our responses to their suggestions. They should also look  as mainnet / omega metrics.

Then, design, propose and implement both the 

- new fee oracle contracts
-  new xfeemngr
- a process for upgrading live networks (omega & mainnet) to the new oracles

Zodomo:
I was able to significantly decrease gas consumption in our bulkSetFeeParams() function in FeeOracleV2 by ~45%. This will make it substantially cheaper for us to operate it. However, I am reopening this issue as there might still need to be some discussion about whether the change to targeting the appropriate L1 fee market (be it blobs or calldata) is sufficient, or if we have any alternative strategies worth pursuing.

PR #1952, opened 2024-09-23
feat(contracts/core): add fee oracle v2

kevinhalliday:
Add FeeOracleV2 contract.

- fee oracle v2 uses explicit data & exec gas prices per chain
- add version() view on fee oracles

issue: #1951


contracts/core/src/xchain/FeeOracleV2.sol
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * @title FeeOracleV2
+ * @notice A simple fee oracle with a fixed fee, controlled by an admin account
+ *         Used by OmniPortal to calculate xmsg fees
+ */
+contract FeeOracleV2 is IFeeOracle, IFeeOracleV2, OwnableUpgradeable {
+    /**
+     * @notice Base gas limit for each xmsg.
+     */
+    uint64 public baseGasLimit;
```
ttarsi, line 17: maybe a bit more descriptive name - I actually don't know what this is referring to. Is it the minimum gas limit a developer can set?

ttarsi (approved):

contracts/core/src/xchain/FeeOracleV2.sol
kevinhalliday: No that's `xmsgMinGasLimit` in the portal .

`baseGasLimit` is the base gas limit we charge for each xmsg, to account for overhead.

Issue #1920, opened 2024-09-19
Implement remaining contract admin actions


fabtreb:
Prioritize and make sub task for the following

Needs Implementing

- [x] pause/unpause-xcall [HIGH]
- [x] pause/unpause-xcall to chain [HIGH]
- [x] pause/unpause-xsubmit  [HIGH]
- [x] pause/unpause-xsubmit from chain [HIGH]
- [x] pause / unpause bridge [HIGH]
- [ ] set XBlock / XMsg offset on a portal [MED]
- [ ] collect-fees from portal [MED]
- [ ] configure portals (all config xreceiptMaxErrorSize , xsubValsetCutoff , etc) [MED]
- [ ] configure fee oracle (protocolFee, baseGasLimit ) [LOW]
- [ ] enable/disable staking allow list [LOW]
- [x]  remaining upgrade actions (FeeOracleV1, Staking, Slashing, Upgrades , OmniGasPump , OmniGasStation , PortalRegistry …) [MED/HIGH - contract dependent]

kc1116:
All admin commands that we need are implemented , closing.

PR #1954, opened 2024-09-24
feat(e2e): add contract upgrade actions

kevinhalliday:
Add e2e upgrade actions for contracts.

Contracts:

- Slashing
- Staking
- OmniBridgeNative
- OmniBridgeL1
- OmniGasStation
- OmniGasPump
- FeeOracleV1

issue: #1920


contracts/core/script/admin/Admin.s.sol
```diff
@@ -137,4 +139,154 @@ contract Admin is Script {
...
+     */
+    function _upgradeProxy(address admin, address proxy, address impl, bytes calldata data) internal {
+        _upgradeProxy(admin, proxy, impl, data, true);
+    }
+
+    function _upgradeProxy(address admin, address proxy, address impl, bytes calldata data, bool initializable)
+        internal
+    {
+        vm.startBroadcast(admin);
+        address proxyAdmin = EIP1967Helper.getAdmin(proxy);
```
ttarsi, line 284: can this be outside of the broadcast?

kevinhalliday: yes good point. in this case it does not matter but better to be careful about what's in broadcast

Issue #1958, opened 2024-09-24
Staging network version is random per-runtime


kevinhalliday:
*Problem To Solve*

Staging's `lib/netconf` network version is a randomly generated uuid per runtime. This network version is in create3 salts. This means that we cannot use `lib/contracts` addresses to interact with staging. 

*Proposed Solution*

Derive staging network version from staging genesis.

PR #1959, opened 2024-09-24
feat(lib/netconf): deterministic staging network version

kevinhalliday:
Derive staging network version from genesis. This let's use lib/contract addresses across run times.

Additional:

- refactor e2e to setup before deployments
- download genesis in e2e for non-deployments

issue: #1958


kevinhalliday:
Closing in favor of https://github.com/omni-network/omni/pull/1963

PR #1963, opened 2024-09-25
feat(lib/contracts): consistent staging addrs

kevinhalliday:
Replace dynamic staging network version / contract salts with consistent staging salts.
This way we can use lib/contract addresses in runtimes and still get the same addrs.

issue:  #1958


lib/contracts/address.go
```diff
@@ -1,88 +1,223 @@
...
+	L1Bridge   string
+	Portal     string
+	Token      string
+}
+
+var (
+	// cached addrs.
+	addrs = map[netconf.ID]Addresses{}
+	// cached salts.
+	salts = map[netconf.ID]Salts{}
```
corverroos, line 101: Are these caches populated?

e2e/app/run.go
```diff
@@ -79,6 +81,14 @@ func Deploy(ctx context.Context, def Definition, cfg DeployConfig) (*pingpong.XD
 		return nil, err
 	}
 
+	if def.Testnet.Network == netconf.Staging {
+		if len(def.Testnet.OmniEVMs) == 0 {
+			return nil, errors.New("no omni evms")
+		}
+
+		contracts.UseStagingOmniRPC(def.Testnet.OmniEVMs[0].ExternalRPC)
```
corverroos, line 89: rather use `def.Testnet.BroadcastOmniEVM()`

lib/contracts/address.go
kevinhalliday: Oops

PR #1971, opened 2024-09-25
feat(e2e): add --broadcast flag to admin commands

kevinhalliday:
Add --broadcast flag to admin commands, to
broadcast transactions. Default to no broadcast.

issue: #1920

### DIFF
--- a/e2e/app/admin/config.go
+++ b/e2e/app/admin/config.go
@@ -1,10 +1,13 @@
 package admin
 
 type Config struct {
+	// Broadcast  determines whether or not transactions.
+	Broadcast bool
+
 	// Chain is the name of chain to run on. Leave empty to run on all applicable chains.
 	Chain string
 }
 
 func DefaultConfig() Config {
-	return Config{Chain: ""}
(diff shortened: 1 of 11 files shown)


e2e/app/admin/config.go
```diff
@@ -1,10 +1,38 @@
...
+	return ChainConfig{Chain: ""}
+}
+
+func DefaultBaseConfig() BaseConfig {
+	return BaseConfig{Broadcast: false}
+}
+
+type configKey struct{}
+
+func CtxWithBaseConfig(ctx context.Context, cfg *BaseConfig) context.Context {
```
corverroos, line 27: putting critical info in the context is considered bad practice. Only stuff that you don't actually need for logic (nice-to-haves) should be included in ctx.

kevinhalliday: yeah I was gonna ask you about this live.  it's nice not to have to wire it though all the function calls. and I thought requiring something be explicitly "positive" might make this okay? 

but yes, noted on base practices

corverroos (approved):

Issue #1962, opened 2024-09-24
Lack of visibility into xchain protocol PnL


kevinhalliday:
*Problem To Solve*

We currently have poor visibility into xchain protocol PnL.

The relayer and monitor both spend ETH / OMNI. The portals collect fees. Are only metrics for this are balance based - relayer balance, monitor balance, portal balance. These balance metrics will "reset" on funding actions, or fee colleciton. We need explicit spend / revenue metrics.

*Proposes Solution*

- add explicit relayer / monitor spend metrics
- add explicit portal revenue metrics
- update dashboard to clearly show pnl

PR #2011, opened 2024-09-30
chore(monitor): use gwei in fee total metrics

kevinhalliday:
Use gwei in fee metrics, add unit to metric name.

issue: #1962

### DIFF
--- a/monitor/xmonitor/indexer/metrics.go
+++ b/monitor/xmonitor/indexer/metrics.go
@@ -4,6 +4,10 @@ import (
 	"math/big"
 	"time"
 
+	"github.com/omni-network/omni/lib/umath"
+
+	"github.com/ethereum/go-ethereum/params"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 )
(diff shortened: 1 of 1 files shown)


monitor/xmonitor/indexer/metrics.go
```diff
@@ -46,10 +50,10 @@ var (
 		Buckets:   prometheus.ExponentialBucketsRange(1, 1e6, 10),
 	}, []string{"stream", "xdapp"})
 
-	feesCollectedTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+	feesGweiTotal = promauto.NewCounterVec(prometheus.CounterOpts{
 		Namespace: "monitor",
 		Subsystem: "indexer",
-		Name:      "fees_collected_total",
+		Name:      "fees_gwei_total",
```
chmllr, line 56: Nit: did you maybe consider `total_fees_gwei` as it's common to use the unit in a suffix?

kevinhalliday: we have a linter that enforces `_total` as suffix for counter metrics

corverroos: yeah, counters are usually suffixed with `total`

chmllr: Oh I see, those are metrics too 👍

corverroos (approved):

PR #2012, opened 2024-09-30
feat(e2e): add ensure-bridge-spec command

kevinhalliday:
Add command to ensure live bridge specs match local config.

Allows pausing of l1 and native bridge.

issue: #1920


ttarsi (approved):
This looks good overall, but a couple comments:
- I like the declarative spec because it's a single source of truth for the state of a network
- But I don't like the following
  -  there are now 2 sources of truth – `bridgespec.go` and `portalspec.go`
  - the actual configurations are deep within the files and in between typedefs/functiondefs, the bridge configuration starts on line 71
  - the files are in deep subdirs `e2e/app/admin/...`
- 2 and 3 mean this is hard to find. And it's very important that this is easy to find in emergencies.

I see a couple options that could solve this:
- define the portaltypes and bridgetypes in their own files. but then have 1 spec.go, which holds the config for each live network, for both bridge and portal. have 1 source of truth. also note clearly in the readme where the current config lives (should do this anyway)
- define the config in ops – like in the `env` dir. could even turn those into yamls. and then when a config changes, submit a PR to ops to change the config there, and then run a GH action there

These are just my recommendations though, up to you what you want to run with.

kevinhalliday:
I like having them next to their corresponding command. I can move them to top of file.  

As for separate sources of truth - they are indeed separate. One for bridge. One for portals. So we can keep them separate. 

I'll update readme

Issue #1682, opened 2024-08-09
Add ready healthcheck endpoint


corverroos:
*Problem to Solve*
Omni node operators don't have a easy way to monitor the "readiness" of a halo node.

The best option at this point is the `halo_comet_synced` prometheus metric or the cometBFT `SyncInfo. catching_up` field. SyncInfo requires a special grpc clients and both require parsing of response.

A simple `curl halo:25560/ready` endpoint would provide great value if it includes a few simple "health checks":
 - CometBFT is synced 
 - CometBFT has peers
 - Geth connected
 - Geth is syned
 - Geth has peers

A prometheus metric that tracks this would also be great.

*Proposed Solution*
- Run `halo/app/monitor.go` loops every 10s. 
- Store results in a shared `ready` struct
- Query shared `ready` struct on `/ready` endpoint.
- Note that we should disable the default cometBFT prometheus server and serve our own on that port including above `/ready` and default prometheus `/metrics` endpoints.

chmllr:
@corverroos two questions:

1. How about serving a simple json response like
```json
{
"CometBFT": "synced",
"CometBFT": "peers",
"Geth": {
     "connected": true,
     "synced": true,
     "peers": 10,
   }
}
```
so that it's both human-readable and parsable if needed later at some point?



2. Wdyt about calling the endpoint and the struct `/health` and `health` instead?

corverroos:
`ready` follows the [k8s](https://kubernetes.io/docs/reference/using-api/health-checks/) style
![image](https://github.com/user-attachments/assets/cccefd21-46ee-4dfe-a2d5-363bbb2fdc6c)

corverroos:
wrt returned struct, sure, we can do that in addition to status codes. I'd suggest to keep it simple and avoid nesting and to stick to general consensus/execution rather than implementations which may change (especially geth):
```
type ReadyResponse struct {
 ConsensusSynced bool `json:"comet_bft_synced"` // Return 503 is false
 ConsensusP2PPeers int  `json...` // Returns 503 if 0
 ExecutionConnected bool `json:..` // Returns 503 if false 
 ExecutionSynced bool `json...` // Returns 503 if false 
ExecutionP2PPeers int `json...` // Returns 503 if 0 
}
```

chmllr:
Closed with https://github.com/omni-network/omni/pull/2035

PR #2035, opened 2024-10-02
feat(halo/app): add ready endpoint

chmllr:
- Disables the default prometheus server of CometBFT,
- Registers two endpoints,
  - `/` returns the prometheus metrics as before (note, it was not served on `/metrics`),
  - `/ready` returns the readiness status (with `503` as a response code if not ready yet).

Note, we use a shared global variable holding the flags since we can only register the http endpoints once, yet the tests are calling the `Start` routine more than once from the same process.

In a follow up PR, we'll add a grafana metrics and docker endpoints using the newly introduced endpoint.

issue: https://github.com/omni-network/omni/issues/1682

### DIFF
--- a/e2e/cmd/cmd.go
+++ b/e2e/cmd/cmd.go
@@ -103,7 +103,12 @@ func newDeployCmd(def *app.Definition) *cobra.Command {
 		Use:   "deploy",
 		Short: "Deploys the e2e network",
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			_, err := app.Deploy(cmd.Context(), *def, cfg)
+			ctx := cmd.Context()
+			_, err := app.Deploy(ctx, *def, cfg)
+			if err != nil {
+				log.Error(ctx, "Deployment failed", err)
+			}
+
 			return err
 		},
 	}
--- a/e2e/cmd/cmd.go
+++ b/e2e/cmd/cmd.go
@@ -103,12 +103,7 @@ func newDeployCmd(def *app.Definition) *cobra.Command {
 		Use:   "deploy",
 		Short: "Deploys the e2e network",
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			ctx := cmd.Context()
-			_, err := app.Deploy(ctx, *def, cfg)
-			if err != nil {
-				log.Error(ctx, "Deployment failed", err)
-			}
-
+			_, err := app.Deploy(cmd.Context(), *def, cfg)
 			return err
 		},
 	}
--- a/halo/app/monitor.go
+++ b/halo/app/monitor.go
@@ -59,12 +59,12 @@ func monitorCometForever(
 
 // monitorCometOnce monitors the cometBFT peers, and sync status.
 func monitorCometOnce(ctx context.Context, rpcClient rpcclient.Client, isSyncing func() bool, lastHeight int64, readiness *ReadyResponse) (int64, error) {
-	if netInfo, err := rpcClient.NetInfo(ctx); err != nil {
+	netInfo, err := rpcClient.NetInfo(ctx)
+	readiness.SetConsensusP2PPeers(netInfo.NPeers)
+	if err != nil {
 		return 0, errors.Wrap(err, "net info")
 	} else if netInfo.NPeers == 0 {
 		log.Error(ctx, "Halo has 0 consensus p2p peers", nil)
-	} else {
-		readiness.SetConsensusP2PPeers(netInfo.NPeers)
 	}
 
 	synced := !isSyncing()
--- a/halo/app/monitor.go
+++ b/halo/app/monitor.go
@@ -28,7 +28,7 @@ func monitorCometForever(
 		return // Simnet doesn't need to monitor cometBFT, since no p2p.
 	}
 
-	ticker := time.NewTicker(time.Second * 30)
+	ticker := time.NewTicker(time.Second * 10)
 	defer ticker.Stop()
 
 	var lastHeight int64
@@ -88,7 +88,7 @@ func monitorEVMForever(ctx context.Context, cfg Config, ethCl ethclient.Client, 
 		return // Simnet doesn't have an EVM tp monitor.
 	}
 
-	ticker := time.NewTicker(time.Second * 30)
+	ticker := time.NewTicker(time.Second * 10)
 	defer ticker.Stop()
 
 	// Geth Auth API (EngineClient) doesn't enable net module, so we can't monitor peer count with it.
--- a/halo/app/monitor.go
+++ b/halo/app/monitor.go
@@ -143,6 +143,7 @@ func monitorEVMOnce(ctx context.Context, ethCl ethclient.Client, status *readine
 		// SyncProgress returns nil of not syncing.
 		evmSynced.Set(0)
 		log.Warn(ctx, "Attached omni evm is syncing", nil, "highest_block", syncing.HighestBlock, "current_block", syncing.CurrentBlock, "tx_indexing", syncing.TxIndexRemainingBlocks)
+		status.setExecutionSynced(false)
 	} else {
 		evmSynced.Set(1)
 		status.setExecutionSynced(true)
@@ -190,8 +191,9 @@ func startMonitoringAPI(
 	status *readinessStatus) func(context.Context) error {
 	mux := http.NewServeMux()
 
+	// We align both with industry standards (endpoint /metrics)
+	// and with cometBFT (endpoint /).
 	mux.Handle("/metrics", promhttp.Handler())
-	// We also export the metrics on the root path as this is how the default prometheus server works.
 	mux.Handle("/", promhttp.Handler())
 
 	// On the `/ready` endpoint, we serve the readiness of the halo node.
--- a/halo/app/monitor.go
+++ b/halo/app/monitor.go
@@ -2,7 +2,6 @@ package app
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -14,7 +13,6 @@ import (
 	"github.com/omni-network/omni/lib/netconf"
 
 	cmtcfg "github.com/cometbft/cometbft/config"
-	"github.com/cometbft/cometbft/node"
 	rpcclient "github.com/cometbft/cometbft/rpc/client"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -82,7 +80,6 @@ func monitorCometOnce(ctx context.Context, rpcClient rpcclient.Client, isSyncing
 		return 0, errors.Wrap(err, "abci info")
 	} else if !isSyncing() && lastHeight > 0 && abciInfo.Response.LastBlockHeight <= lastHeight {
 		log.Warn(ctx, "Halo height is not increasing, evm syncing?", nil, "height", abciInfo.Response.LastBlockHeight)
-		status.setConsensusSynced(false)
 	}
 
 	return abciInfo.Response.LastBlockHeight, nil
@@ -186,8 +183,15 @@ func dirSize(path string) (int64, error) {
 }
 
 // startMonitoringAPI registers metrics and health endpoints.
-func startMonitoringAPI(ctx context.Context, cfg *cmtcfg.Config, cmtNode *node.Node, syncAbort chan error, status *readinessStatus) {
+// Return the function shutting down the HTTP server.
+func startMonitoringAPI(
+	cfg *cmtcfg.Config,
+	asyncAbort chan<- error,
+	status *readinessStatus) func(context.Context) error {
 	mux := http.NewServeMux()
+
+	mux.Handle("/metrics", promhttp.Handler())
+	// We also export the metrics on the root path as this is how the default prometheus server works.
 	mux.Handle("/", promhttp.Handler())
 
 	// On the `/ready` endpoint, we serve the readiness of the halo node.
@@ -212,16 +216,9 @@ func startMonitoringAPI(ctx context.Context, cfg *cmtcfg.Config, cmtNode *node.N
 
 	go func() {
 		if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
-			errorMsg := fmt.Sprintf("HTTP server failed to start: %v", err)
-			syncAbort <- errors.New(errorMsg)
+			asyncAbort <- errors.Wrap(err, "http server failed to start")
 		}
 	}()
 
-	go func() {
-		cmtNode.Wait()
-
-		if err := server.Shutdown(ctx); err != nil {
-			syncAbort <- err
-		}
-	}()
+	return server.Shutdown
 }
--- a/halo/app/monitor.go
+++ b/halo/app/monitor.go
@@ -22,7 +22,7 @@ func monitorCometForever(
 	rpcClient rpcclient.Client,
 	isSyncing func() bool,
 	dbDir string,
-	readiness *ReadyResponse,
+	status *readinessStatus,
 ) {
 	if network == netconf.Simnet {
 		return // Simnet doesn't need to monitor cometBFT, since no p2p.
@@ -38,7 +38,7 @@ func monitorCometForever(
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			height, err := monitorCometOnce(ctx, rpcClient, isSyncing, lastHeight, readiness)
+			height, err := monitorCometOnce(ctx, rpcClient, isSyncing, lastHeight, status)
 			if err != nil {
 				log.Warn(ctx, "Failed monitoring cometBFT (will retry)", err)
 				// Don't reset lastHeight to zero.
@@ -58,9 +58,9 @@ func monitorCometForever(
 }
 
 // monitorCometOnce monitors the cometBFT peers, and sync status.
-func monitorCometOnce(ctx context.Context, rpcClient rpcclient.Client, isSyncing func() bool, lastHeight int64, readiness *ReadyResponse) (int64, error) {
+func monitorCometOnce(ctx context.Context, rpcClient rpcclient.Client, isSyncing func() bool, lastHeight int64, status *readinessStatus) (int64, error) {
 	netInfo, err := rpcClient.NetInfo(ctx)
-	readiness.SetConsensusP2PPeers(netInfo.NPeers)
+	status.setConsensusP2PPeers(netInfo.NPeers)
 	if err != nil {
 		return 0, errors.Wrap(err, "net info")
 	} else if netInfo.NPeers == 0 {
@@ -69,7 +69,7 @@ func monitorCometOnce(ctx context.Context, rpcClient rpcclient.Client, isSyncing
 
 	synced := !isSyncing()
 	setConstantGauge(cometSynced, synced)
-	readiness.SetConsensusSynced(synced)
+	status.setConsensusSynced(synced)
 
 	abciInfo, err := rpcClient.ABCIInfo(ctx)
 	if err != nil {
@@ -83,7 +83,7 @@ func monitorCometOnce(ctx context.Context, rpcClient rpcclient.Client, isSyncing
 
 // monitorEVMForever blocks until the contract is canceled.
 // It periodically calls monitorEVMOnce.
-func monitorEVMForever(ctx context.Context, cfg Config, ethCl ethclient.Client, readiness *ReadyResponse) {
+func monitorEVMForever(ctx context.Context, cfg Config, ethCl ethclient.Client, status *readinessStatus) {
 	if cfg.Network == netconf.Simnet {
 		return // Simnet doesn't have an EVM tp monitor.
 	}
@@ -100,7 +100,6 @@ func monitorEVMForever(ctx context.Context, cfg Config, ethCl ethclient.Client, 
 		if err == nil {
 			ethCl = newEthCl
 			log.Info(ctx, "Using rpc endpoint to monitor attached omni evm", "rpc", omniEVMRPC)
-			readiness.SetExecutionConnected(true)
 		}
 	}
 
@@ -109,7 +108,7 @@ func monitorEVMForever(ctx context.Context, cfg Config, ethCl ethclient.Client, 
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			err := monitorEVMOnce(ctx, ethCl, readiness)
+			err := monitorEVMOnce(ctx, ethCl, status)
 			if err != nil {
 				log.Warn(ctx, "Failed monitoring attached omni evm (will retry)", err, "addr", ethCl.Address())
 			}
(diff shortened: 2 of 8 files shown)


halo/app/start.go
```diff
@@ -49,6 +52,45 @@ import (
 	grpc1 "github.com/cosmos/gogoproto/grpc"
 )
 
+type ReadyResponse struct {
```
corverroos, line 55: lets move this to its own file: `ready.go`

halo/app/monitor.go
```diff
@@ -57,14 +58,18 @@ func monitorCometForever(
...
 
 // monitorCometOnce monitors the cometBFT peers, and sync status.
-func monitorCometOnce(ctx context.Context, rpcClient rpcclient.Client, isSyncing func() bool, lastHeight int64) (int64, error) {
+func monitorCometOnce(ctx context.Context, rpcClient rpcclient.Client, isSyncing func() bool, lastHeight int64, readiness *ReadyResponse) (int64, error) {
 	if netInfo, err := rpcClient.NetInfo(ctx); err != nil {
 		return 0, errors.Wrap(err, "net info")
 	} else if netInfo.NPeers == 0 {
 		log.Error(ctx, "Halo has 0 consensus p2p peers", nil)
+	} else {
+		readiness.SetConsensusP2PPeers(netInfo.NPeers)
```
corverroos, line 67: you should also set this if `netInfo.NPeers == 0`

halo/app/start.go
corverroos: also, don't export types/functions if not used externally #privatebydefault

corverroos: also, please add comments to all types/functions

halo/app/start.go
```diff
@@ -204,8 +246,28 @@ func Start(ctx context.Context, cfg Config) (<-chan error, func(context.Context)
 		return nil, nil, err
 	}
 
-	go monitorCometForever(ctx, cfg.Network, rpcClient, cmtNode.ConsensusReactor().WaitSync, cfg.DataDir())
-	go monitorEVMForever(ctx, cfg, engineCl)
+	readiness := &ReadyResponse{}
+	server := &http.Server{
```
corverroos, line 250: lets move this to a function, simplify the main start func

halo/app/start.go
```diff
@@ -204,8 +246,28 @@ func Start(ctx context.Context, cfg Config) (<-chan error, func(context.Context)
...
+		Addr:              ":25560",
```
corverroos, line 251: don't hardcode this, use the existing cometBFT config to read the value

halo/app/start.go
```diff
@@ -204,8 +246,28 @@ func Start(ctx context.Context, cfg Config) (<-chan error, func(context.Context)
...
+		http.HandleFunc("/ready", func(w http.ResponseWriter, _ *http.Request) {
+			w.Header().Set("Content-Type", "application/json")
+			if err := json.NewEncoder(w).Encode(readiness); err != nil {
+				http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+				return
+			}
+		})
+
+		if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			log.Error(ctx, "HTTP server failed to start", err)
```
corverroos, line 265: this should crash the app it is doesn't work

halo/app/start.go
```diff
@@ -218,10 +280,16 @@ func Start(ctx context.Context, cfg Config) (<-chan error, func(context.Context)
 		}
 		cmtNode.Wait()
 
+		if err := server.Shutdown(ctx); err != nil {
+			log.Error(ctx, "Server shutdown failed:%+v", err)
```
corverroos, line 284: Never include fmt sprint type verbs in logs or errors

corverroos: always return errors

halo/app/start.go
corverroos: prefer to always return errors

halo/app/start.go
```diff
@@ -218,10 +280,16 @@ func Start(ctx context.Context, cfg Config) (<-chan error, func(context.Context)
...
+		}
+
 		// Note that cometBFT doesn't shut down cleanly. It leaves a bunch of goroutines running...
 
 		if err := stopTracer(ctx); err != nil {
 			return errors.Wrap(err, "stop tracer")
+		} else {
+			log.Info(ctx, "HTTP server stopped")
```
corverroos, line 292: avoid unnecessary logs, this doesn't add much value

halo/app/start.go
corverroos: also, don't this clash with cometBFT internal prometheus port like mentioned in the issue?

chmllr: Btw, the CometBFT Prometheus port is `26660` so there be no conflicts with the port suggested in the issue (`25560`) but the real question what is the reason for suggesting disabling the default prometheus server? We can have more than one server, each serving its own thing, right?

chmllr: I think I figured it out.

halo/app/start.go
chmllr: Yep, I added this when I was debugging the failing test.

e2e/cmd/cmd.go
```diff
@@ -103,7 +103,12 @@ func newDeployCmd(def *app.Definition) *cobra.Command {
 		Use:   "deploy",
 		Short: "Deploys the e2e network",
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			_, err := app.Deploy(cmd.Context(), *def, cfg)
+			ctx := cmd.Context()
+			_, err := app.Deploy(ctx, *def, cfg)
+			if err != nil {
+				log.Error(ctx, "Deployment failed", err)
```
chmllr, line 109: Without this log, a misconfigured compose template would lead to a deployment failure without any error message in the console.

corverroos: As per go guidelines, don't "log AND return errors". We return the error below, so don't log it. The caller of this function will log if required.

corverroos: The caller being `libcmd.Main`

chmllr: Thanks a lot, this is very helpful! Will fix.

halo/app/ready.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"encoding/json"
+	"io"
+	"sync"
+
+	"github.com/omni-network/omni/lib/errors"
+)
+
+// Consist of all flags describing the current readiness status of a halo node.
+type ReadyResponse struct {
```
corverroos, line 12: don't export types/funcs by default, rather make it unexported. Same as the fields below, since they are accessed via methods, making them private is better since it makes it clear that callers/users shoudln't read/write fields directly.

corverroos: mmm, but this will result in it not seriazling the json, so maybe fields can be exported. Just add a godoc that fields shouldn't be written/read directly. If you want to be purist, you could have a `readyResponseJSON` that is only used for serialization.

halo/app/ready.go
```diff
@@ -0,0 +1,75 @@
...
+	mu                 sync.RWMutex
+	ConsensusSynced    bool   `json:"consensus_synced"`      // Unhealthy if false
+	ConsensusP2PPeers  int    `json:"consensus_p_2_p_peers"` // Unhealthy if 0
+	ExecutionConnected bool   `json:"execution_connected"`   // Unhealthy if false
+	ExecutionSynced    bool   `json:"execution_synced"`      // Unhealthy if false
+	ExecutionP2PPeers  uint64 `json:"execution_p_2_p_peers"` // Unhealthy if 0
+}
+
+// Indicates if the current status is considered as healthy.
+func (r *ReadyResponse) Healthy() bool {
```
corverroos, line 22: nit: `Ready`

halo/app/monitor.go
```diff
@@ -112,7 +118,7 @@ func monitorEVMForever(ctx context.Context, cfg Config, ethCl ethclient.Client)
 }
 
 // monitorEVMOnce monitors the attached omni_evm height, peers, and sync status.
-func monitorEVMOnce(ctx context.Context, ethCl ethclient.Client) error {
+func monitorEVMOnce(ctx context.Context, ethCl ethclient.Client, readiness *ReadyResponse) error {
 	// Best effort monitoring of peer count, since method not available in auth API.
 	peers, err := ethCl.PeerCount(ctx)
 	if ethclient.IsErrMethodNotAvailable(err) { //nolint:revive // Empty block skips error handling below.
```
corverroos, line 124: most halo instances will not have EVM RPC defined, so `ethclient.IsErrMethodNotAvailable` will be true, which means you will never set geth peers, which will mean

halo/app/monitor.go
```diff
@@ -78,12 +83,12 @@ func monitorCometOnce(ctx context.Context, rpcClient rpcclient.Client, isSyncing
...
+func monitorEVMForever(ctx context.Context, cfg Config, ethCl ethclient.Client, readiness *ReadyResponse) {
 	if cfg.Network == netconf.Simnet {
 		return // Simnet doesn't have an EVM tp monitor.
 	}
 
-	ticker := time.NewTicker(time.Second * 30)
+	ticker := time.NewTicker(time.Second * 10)
 	defer ticker.Stop()
 
 	// Geth Auth API (EngineClient) doesn't enable net module, so we can't monitor peer count with it.
```
corverroos, line 94: see here. Most halo instances only have engine API client, so cannot monitor geth peers. In this case, skip gethpeers==0 checks.

halo/app/monitor.go
```diff
@@ -95,6 +100,7 @@ func monitorEVMForever(ctx context.Context, cfg Config, ethCl ethclient.Client)
 		if err == nil {
 			ethCl = newEthCl
 			log.Info(ctx, "Using rpc endpoint to monitor attached omni evm", "rpc", omniEVMRPC)
+			readiness.SetExecutionConnected(true)
```
corverroos, line 103: engineclient also counts as connected

halo/app/monitor.go
```diff
@@ -134,6 +141,7 @@ func monitorEVMOnce(ctx context.Context, ethCl ethclient.Client) error {
 		log.Warn(ctx, "Attached omni evm is syncing", nil, "highest_block", syncing.HighestBlock, "current_block", syncing.CurrentBlock, "tx_indexing", syncing.TxIndexRemainingBlocks)
 	} else {
 		evmSynced.Set(1)
+		readiness.SetConsensusSynced(true)
```
corverroos, line 144: this is ExecutionSynced, not  ConsensusSynced.

corverroos (commented):
Also, how are we going to:
 - test in dev
 - run the code in prod (regularly)
 - monitor in prod

halo/app/ready.go
chmllr: My plan was also to import the type in the test to unmarshall the endpoint output. I think this will not work for unexported type, right? Would you prefer me to just redefine the struct in the test instead?

corverroos: yeah, don't export just for tests, rather make the test `_internal` or duplicate the struct in the tests

chmllr:
> Also, how are we going to:
> 
> * test in dev
> * run the code in prod (regularly)
> * monitor in prod

Discussed this in a call. 

- Test in dev should be covered by the test added already,
- Monitoring in prod will be addressed in a new PR where I'd add a docker supported endpoint.

halo/app/start.go
```diff
@@ -65,6 +75,34 @@ func (c Config) BackendType() dbm.BackendType {
 	return dbm.BackendType(c.Config.BackendType)
 }
 
+// Register metrics and health endpoints.
+func registerEndpoints(cfg *cmtcfg.Config) {
```
corverroos, line 79: `startMonitoringAPI`

halo/app/start.go
```diff
@@ -65,6 +75,34 @@ func (c Config) BackendType() dbm.BackendType {
...
+		}
+	})
+
+	server := &http.Server{
+		Addr:              cfg.Instrumentation.PrometheusListenAddr,
+		ReadHeaderTimeout: 3 * time.Second,
+	}
+
+	if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+		panic(fmt.Sprintf("HTTP server failed to start: %v", err))
```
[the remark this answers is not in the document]
corverroos: can use `asyncAbort` to return error

halo/app/start.go
```diff
@@ -65,6 +75,34 @@ func (c Config) BackendType() dbm.BackendType {
 	return dbm.BackendType(c.Config.BackendType)
 }
 
+// Register metrics and health endpoints.
+func registerEndpoints(cfg *cmtcfg.Config) {
+	http.Handle("/", promhttp.Handler())
```
corverroos, line 80: lets avoid using global http mux

halo/app/start.go
```diff
@@ -47,8 +50,15 @@ import (
...
 	"github.com/cosmos/cosmos-sdk/types/mempool"
 	grpc1 "github.com/cosmos/gogoproto/grpc"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+// Global variable referenced from the http handler, returning it.
+var status readinessStatus
+
+// Ensures we register the http endpoints exactly once (otherwise, tests will panic).
+var registerOnce sync.Once
```
corverroos, line 60: this shouldn't be required

halo/app/start.go
corverroos: lets move this to `monitor.go`

halo/app/ready.go
```diff
@@ -0,0 +1,96 @@
...
+	mu           sync.RWMutex
+	status       flags
+}
+
+type flags struct {
+	ConsensusSynced    bool   `json:"consensus_synced"`      // Unhealthy if false
+	ConsensusP2PPeers  int    `json:"consensus_p_2_p_peers"` // Unhealthy if 0
+	ExecutionConnected bool   `json:"execution_connected"`   // Unhealthy if false
+	ExecutionSynced    bool   `json:"execution_synced"`      // Unhealthy if false
+	ExecutionP2PPeers  uint64 `json:"execution_p_2_p_peers"` // Unhealthy if 0
```
corverroos, line 24: nit: `execution_p2p_peers`

halo/app/ready.go
```diff
@@ -0,0 +1,96 @@
...
+	// Indicates if the current halo instance uses an EVM RPC endpoint instead of running an execution client.
+	peersExposed bool
+	mu           sync.RWMutex
+	status       flags
+}
+
+type flags struct {
+	ConsensusSynced    bool   `json:"consensus_synced"`      // Unhealthy if false
+	ConsensusP2PPeers  int    `json:"consensus_p_2_p_peers"` // Unhealthy if 0
+	ExecutionConnected bool   `json:"execution_connected"`   // Unhealthy if false
```
corverroos, line 22: we should probably update `ExecutionConnected` if we loose connection?

halo/app/ready.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"encoding/json"
+	"io"
+	"sync"
+
+	"github.com/omni-network/omni/lib/errors"
+)
+
+// Consist of all flags describing the current readiness status of a halo node.
```
corverroos, line 11: nit: godoc start with type name:
```
readinessStatus contains all state related to halo readiness (health).
```

halo/app/ready.go
```diff
@@ -0,0 +1,96 @@
...
+type readinessStatus struct {
+	// Indicates if the current halo instance uses an EVM RPC endpoint instead of running an execution client.
```
corverroos, line 13: nit:
```
// executionPeerAvailable defines whether a optional connection to geth HTTP RPC is available to query peers. It isn't available via the required Auth RPC (EngineAPI)
executionPeerAvailable bool
```

halo/app/ready.go
```diff
@@ -0,0 +1,96 @@
...
+	peersExposed bool
+	mu           sync.RWMutex
+	status       flags
```
corverroos, line 16: not sure about different name for field and type, can't it be the same?

halo/app/start.go
chmllr: Thanks a lot, I didn't know one can specify a local multiplexer! (Why does a global one even exist?) This was the reason why I needed the global variable and `Once.Do`, because test call `Start` within the same process.

halo/app/ready.go
chmllr: I think you had the old version in the tab, I removed this flag because having a pointer gives me all the information I need and I don't need to maintain an additional flag.

halo/app/ready.go
chmllr: Yes, changed!

halo/app/ready.go
chmllr: It looks like the only way how the disconnect manifests itself in the app is when we cannot read peer count or the block number. So I can set the disconnection status upon getting any of these two errors. Do you see a better way doing this?

corverroos: yeah, if either of those error, set connected=false

halo/app/monitor.go
```diff
@@ -57,33 +64,38 @@ func monitorCometForever(
...
+	synced := !isSyncing()
+	setConstantGauge(cometSynced, synced)
+	status.setConsensusSynced(synced)
 
 	abciInfo, err := rpcClient.ABCIInfo(ctx)
 	if err != nil {
 		return 0, errors.Wrap(err, "abci info")
 	} else if !isSyncing() && lastHeight > 0 && abciInfo.Response.LastBlockHeight <= lastHeight {
 		log.Warn(ctx, "Halo height is not increasing, evm syncing?", nil, "height", abciInfo.Response.LastBlockHeight)
+		status.setConsensusSynced(false)
```
corverroos, line 85: readiness ConsensusSynced should be aligned with `cometSynced` metric. Lets remove this for now rather.

halo/app/monitor.go
```diff
@@ -167,3 +184,44 @@ func dirSize(path string) (int64, error) {
 
 	return size, nil
 }
+
+// startMonitoringAPI registers metrics and health endpoints.
+func startMonitoringAPI(ctx context.Context, cfg *cmtcfg.Config, cmtNode *node.Node, syncAbort chan error, status *readinessStatus) {
```
corverroos, line 189: lets stick to existing naming, rename to `asyncAbort`

halo/app/monitor.go
```diff
@@ -167,3 +184,44 @@ func dirSize(path string) (int64, error) {
...
+
+	server := &http.Server{
+		Addr:              cfg.Instrumentation.PrometheusListenAddr,
+		ReadHeaderTimeout: 3 * time.Second,
+		Handler:           mux,
+	}
+
+	go func() {
+		if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			errorMsg := fmt.Sprintf("HTTP server failed to start: %v", err)
```
corverroos, line 215: As mentioned before, **_never_** use fmt verbs like "%w" or "%v" or "%s" for logs or errors. We use structured errors, not unstructured errors. Use `errors.Wrap`, search of uses of it and see how we use it.

halo/app/monitor.go
```diff
@@ -167,3 +184,44 @@ func dirSize(path string) (int64, error) {
...
+			syncAbort <- errors.New(errorMsg)
+		}
+	}()
+
+	go func() {
```
corverroos, line 220: rather return a shutdown function and call it in main `Run` shutdown func.

In run.go#Start
```
stopMonitor := startMonitoringAPI(...)

return asyncAbort, func(ctx context.Context) error {
  if err := stopMonitor() {
    return err
  }
  
  ...
}

```

halo/app/ready.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+// readinessStatus contains all state related to halo readiness (health).
+type readinessStatus struct {
+	mu                 sync.RWMutex
+	ConsensusSynced    bool    `json:"consensus_synced"`              // Unhealthy if false
+	ConsensusP2PPeers  int     `json:"consensus_p2p_peers"`           // Unhealthy if 0
+	ExecutionConnected bool    `json:"execution_connected"`           // Unhealthy if false
+	ExecutionSynced    bool    `json:"execution_synced"`              // Unhealthy if false
+	ExecutionP2PPeers  *uint64 `json:"execution_p2p_peers,omitempty"` // Unhealthy if 0
```
corverroos, line 18: does this omit zero peers or not?

halo/app/ready.go
```diff
@@ -0,0 +1,84 @@
...
+	return ready
+}
+
+// serialize marshals the struct to the specified writer.
+func (r *readinessStatus) serialize(w io.Writer) error {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+
+	if err := json.NewEncoder(w).Encode(r); err != nil {
+		return errors.Wrap(err, "serialization failed")
```
corverroos, line 45: see go style guide, no need to add "failed" when wrapping

halo/app/start.go
```diff
@@ -204,8 +204,12 @@ func Start(ctx context.Context, cfg Config) (<-chan error, func(context.Context)
 		return nil, nil, errors.Wrap(err, "start comet node")
 	}
 
-	go monitorCometForever(ctx, cfg.Network, rpcClient, cmtNode.ConsensusReactor().WaitSync, cfg.DataDir())
-	go monitorEVMForever(ctx, cfg, engineCl)
+	status := &readinessStatus{}
```
corverroos, line 207: nit: `state := new(redinessSTatus)`

halo/app/monitor.go
```diff
@@ -167,3 +184,44 @@ func dirSize(path string) (int64, error) {
 
 	return size, nil
 }
+
+// startMonitoringAPI registers metrics and health endpoints.
+func startMonitoringAPI(ctx context.Context, cfg *cmtcfg.Config, cmtNode *node.Node, syncAbort chan error, status *readinessStatus) {
+	mux := http.NewServeMux()
+	mux.Handle("/", promhttp.Handler())
```
corverroos, line 191: maybe just add a comment that we align with cometBFT so not registering this at `/metrics` which is the industry standard.

halo/app/monitor.go
chmllr: Yep, sync is a type. Will change the type 👍

halo/app/monitor.go
chmllr: Great idea!

halo/app/monitor.go
chmllr: Done, I also added `/metrics` too, just in case (the default server exports metrics on any path)

halo/app/monitor.go
chmllr: Done

halo/app/monitor.go
chmllr: Ah I misunderstood the previous comment then, I thought it's related to verbs not being supported when errors are logged (they aren't), but you were talking about structuring logs via wrapping. Makes sense now!

halo/app/start.go
chmllr: Done!

halo/app/monitor.go
```diff
@@ -134,8 +145,11 @@ func monitorEVMOnce(ctx context.Context, ethCl ethclient.Client) error {
 		log.Warn(ctx, "Attached omni evm is syncing", nil, "highest_block", syncing.HighestBlock, "current_block", syncing.CurrentBlock, "tx_indexing", syncing.TxIndexRemainingBlocks)
```
corverroos, line 145: We should set `status.setExecutionSynced(false)` here I think

halo/app/monitor.go
corverroos: nit: update comment, industry standard is `/metrics`, but cometBFT is `/`.

halo/app/start.go
```diff
@@ -204,8 +204,12 @@ func Start(ctx context.Context, cfg Config) (<-chan error, func(context.Context)
 		return nil, nil, errors.Wrap(err, "start comet node")
 	}
 
-	go monitorCometForever(ctx, cfg.Network, rpcClient, cmtNode.ConsensusReactor().WaitSync, cfg.DataDir())
-	go monitorEVMForever(ctx, cfg, engineCl)
+	status := new(readinessStatus)
+
+	httpShutDownFn := startMonitoringAPI(&cfg.Comet, asyncAbort, status)
```
corverroos, line 209: nit, aligning with `stopTracer`, I'd suggest: `stopMonitoringAPI`

halo/app/start_test.go
```diff
@@ -130,6 +132,27 @@ func testAPI(t *testing.T, cfg haloapp.Config) {
 	}
 }
 
+//nolint:noctx // We don't care about best practices here.
+func testHealthAndMetricsEndpoints(t *testing.T, cfg haloapp.Config) {
```
corverroos, line 136: nit: `testReadyEndpoint`

lib/cmd/cmd.go
```diff
@@ -35,6 +35,7 @@ func Main(cmd *cobra.Command) {
 
 	if err != nil {
 		const errExitCode = 1
+		log.Error(ctx, "Command execution failed", err)
```
corverroos, line 38: this has been fixed in https://github.com/omni-network/omni/pull/2088, please remove this.

corverroos (approved):

Issue #2044, opened 2024-10-02
searchOffsetInHistory() latency function is not executed


sideninja:
In `searchOffsetInHistory()`, the function returned by latency(endpoint) is not executed. This will prevent the latency from being logged.

Original issue: https://cantina.xyz/code/953c0c22-8e30-47d0-8567-07a3dadbef77/findings/28

PR #2046, opened 2024-10-02
fix(lib): execute latency metric in offset history

sideninja:
Make sure the latency function is actually called and metrics are reported.

issue: https://github.com/omni-network/omni/issues/2044


corverroos (approved):

sideninja:
Closed in favour of https://github.com/omni-network/omni/pull/2058

Issue with missconfigured commit signatures

PR #2050, opened 2024-10-02
fix(lib/tokens): sanitize coingecko response

kevinhalliday:
Sanitize coingecko response.

- check all requested tokens are in response
- check requrested currency is in response
- check prices are non zero and non negative

issue: none

### DIFF
--- a/lib/tokens/coingecko/coingecko_test.go
+++ b/lib/tokens/coingecko/coingecko_test.go
@@ -15,33 +15,127 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type testCase struct {
+	name         string
+	invalid      bool         // invalid response
+	empty        bool         // empty response
+	omitToken    tokens.Token // omit a requested token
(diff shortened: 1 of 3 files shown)


lib/tokens/coingecko/coingecko_test.go
```diff
@@ -15,47 +15,134 @@ import (
...
 		ids := strings.Split(q.Get("ids"), ",")
 		currencies := strings.Split(q.Get("vs_currencies"), ",")
 
 		resp := make(map[string]map[string]float64)
+
+		if test.empty {
+			bz, _ := json.Marshal(resp)
+			_, _ = w.Write(bz)
+
+			return
```
chmllr, line 101: Nit: if you happen to change the code before merging, consider moving this part up and put under handling `test.invalid` to avoid the unnecessary query parsing.

chmllr (approved):

corverroos (approved):

PR #2071, opened 2024-10-03
fix(lib): check length of xBlock.Msgs before indexing

kc1116:
Add an else if block to check length of Msgs slice before indexing in next else if block, this prevents potential index out of bounds errors. 
 
issue: none

### DIFF
--- a/lib/xchain/provider/fetch.go
+++ b/lib/xchain/provider/fetch.go
@@ -444,6 +444,8 @@ func getConsXBlock(ctx context.Context, ref xchain.EmitRef, cprov cchain.Provide
 		return xchain.Block{}, errors.Wrap(err, "fetch consensus xblock")
 	} else if !ok {
 		return xchain.Block{}, errors.New("no consensus xblocks [BUG]")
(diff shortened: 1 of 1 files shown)


kc1116:
Cantina - https://cantina.xyz/code/953c0c22-8e30-47d0-8567-07a3dadbef77/findings/29

lib/xchain/provider/fetch.go
```diff
@@ -444,6 +444,8 @@ func getConsXBlock(ctx context.Context, ref xchain.EmitRef, cprov cchain.Provide
 		return xchain.Block{}, errors.Wrap(err, "fetch consensus xblock")
 	} else if !ok {
 		return xchain.Block{}, errors.New("no consensus xblocks [BUG]")
+	} else if len(xblock.Msgs) == 0 {
+		return xchain.Block{}, errors.New("no messages [BUG]")
```
corverroos, line 448: nit: `empty consensus xblock [BUG]`

"no messages" is just a bit vauge, what has no messages?

corverroos (approved):

PR #2080, opened 2024-10-04
fix(*): multiple fixes from the reports

sideninja:
Fixes for multiple reported issues in Cantina.

issue: none

### DIFF
--- a/halo/attest/keeper/cpayload.go
+++ b/halo/attest/keeper/cpayload.go
@@ -31,7 +31,7 @@ var _ evmenginetypes.VoteExtensionProvider = (*Keeper)(nil)
 func (k *Keeper) PrepareVotes(ctx context.Context, commit abci.ExtendedCommitInfo) ([]sdk.Msg, error) {
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
 	if err := baseapp.ValidateVoteExtensions(sdkCtx, k.skeeper, sdkCtx.BlockHeight(), sdkCtx.ChainID(), commit); err != nil {
-		return nil, errors.Wrap(err, "validate extensions")
+		return nil, errors.Wrap(err, "validate extensions [BUG]")
 	}
 
 	// Adapt portal registry to the supportedChainFunc signature.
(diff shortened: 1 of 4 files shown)


halo/attest/keeper/cpayload.go
```diff
@@ -33,7 +33,7 @@ func (k *Keeper) PrepareVotes(ctx context.Context, commit abci.ExtendedCommitInf
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
 	if err := baseapp.ValidateVoteExtensions(sdkCtx, k.skeeper, sdkCtx.BlockHeight(), sdkCtx.ChainID(), commit); err != nil {
 		log.Error(ctx, "Cannot include invalid vote extensions in payload", err, "height", sdkCtx.BlockHeight())
```
corverroos, line 35: Can remove logging so we don't log-and-return errors.

lib/xchain/provider/fetch.go
```diff
@@ -444,6 +444,8 @@ func getConsXBlock(ctx context.Context, ref xchain.EmitRef, cprov cchain.Provide
 		return xchain.Block{}, errors.Wrap(err, "fetch consensus xblock")
 	} else if !ok {
 		return xchain.Block{}, errors.New("no consensus xblocks [BUG]")
+	} else if len(xblock.Msgs) == 0 {
+		return xchain.Block{}, errors.New("no xblock messages [BUG]")
```
corverroos, line 448: nit: `empty consensus xblock`

corverroos (approved):

halo/attest/keeper/cpayload.go
```diff
@@ -33,7 +33,7 @@ func (k *Keeper) PrepareVotes(ctx context.Context, commit abci.ExtendedCommitInf
...
-		return nil, nil
+		return nil, errors.Wrap(err, "validate extensions")
```
corverroos, line 36: this is probably an invariant, so we should add `[BUG]`

PR #2086, opened 2024-10-07
chore(halo): addresses multiple info and low sec findings

chmllr:
Addresses multiple informational and low security findings.

issue: none


octane/evmengine/keeper/evmmsgs.go
```diff
@@ -36,9 +36,7 @@ func (k *Keeper) evmEvents(ctx context.Context, blockHash common.Hash) ([]*types
 			return cmp < 0
 		}
 
-		topicI := slices.Concat(events[i].Topics...)
-		topicJ := slices.Concat(events[j].Topics...)
-		if cmp := bytes.Compare(topicI, topicJ); cmp != 0 {
+		if cmp := slices.CompareFunc(events[i].Topics, events[j].Topics, bytes.Compare); cmp != 0 {
```
corverroos, line 39: can we add a test for this please. It isn't obvious that this is the same logic.

corverroos (approved):

octane/evmengine/keeper/evmmsgs.go
corverroos: If this results in different ordering that before, we will get a app hash mismatch panic and the chain will stall. So lets ensure the ordering remains unchanged.

corverroos: In general, be vary careful when changing on-chain logic, since is must never change.

sideninja (approved):

octane/evmengine/keeper/evmmsgs.go
chmllr: Test added. Inside the test, we're comparing the results of the new implementation to the results of the old one.

PR #2094, opened 2024-10-08
chore(*): prevent uses of common BytesTo

corverroos:
Prevent usage of `common.BytesTo(Address/Hash)` which silently pads or trim to the correct length. Rather use `cast` package for explicit length checks.

issue: none

### DIFF
--- a/halo/attest/keeper/keeper.go
+++ b/halo/attest/keeper/keeper.go
@@ -314,8 +314,10 @@ func (k *Keeper) Approve(ctx context.Context, valset ValSet) error {
 			approvedOffset.WithLabelValues(chainVerName).Set(float64(att.GetAttestOffset()))
 		}
 
-		toDelete, ok := isApproved(sigs, valset)
-		if !ok {
+		toDelete, ok, err := isApproved(sigs, valset)
+		if err != nil {
+			return err
+		} else if !ok {
 			// Check if there is a finalized attestation that overrides this one.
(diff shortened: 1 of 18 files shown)


halo/attest/keeper/keeper.go
```diff
@@ -789,7 +796,7 @@ func (k *Keeper) VerifyVoteExtension(ctx sdk.Context, req *abci.RequestVerifyVot
 		duplicate[vote.AttestHeader.ToXChain()] = true
 
 		// Ensure the votes are from the requesting validator itself.
-		if common.BytesToAddress(vote.Signature.ValidatorAddress) != ethAddr {
+		if !bytes.Equal(vote.Signature.ValidatorAddress, ethAddr[:]) {
```
sideninja, line 799: Why not `ethAddr.Cmp(vote.Signature.ValidatorAddress) != 0`

sideninja, line 799: nit: `ethAddr.Cmp(vote.Signature.ValidatorAddress) != 0`

sideninja (approved):

halo/attest/keeper/keeper.go
corverroos: I want to avoid casting `vote.Signature.ValidatorAddress` to a common.Address since it requires error checking. So I rather convert ethAddr to slice and compare that, so no error checks required.

PR #2096, opened 2024-10-08
chore(xfeemngr): Add info logs to gas price and native token rate syncing

kc1116:
Add info logs to gas price and native token rate syncing. This will allow us to debug Grafana alert firing further. 
issue: none

### DIFF
--- a/monitor/xfeemngr/contract/bound.go
+++ b/monitor/xfeemngr/contract/bound.go
@@ -10,6 +10,7 @@ import (
 	"github.com/omni-network/omni/lib/ethclient"
 	"github.com/omni-network/omni/lib/ethclient/ethbackend"
 	"github.com/omni-network/omni/lib/evmchain"
+	"github.com/omni-network/omni/lib/log"
 	"github.com/omni-network/omni/lib/netconf"
 	"github.com/omni-network/omni/lib/umath"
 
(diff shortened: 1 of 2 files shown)


monitor/xfeemngr/contract/bound.go
```diff
@@ -87,6 +88,7 @@ func (c BoundFeeOracleV1) SetGasPriceOn(ctx context.Context, destChainID uint64,
 		return errors.Wrap(err, "tx opts")
 	}
 
+	log.Info(ctx, "Setting gas price on chain", "destination_chain_id", destChainID, "rate", gasPrice.Int64())
```
corverroos, line 91: Make it easy for humans to read logs: best use proper names instead of chainIDs (see `netconf.Network.ChainName`). Also just use `dest_chain` for consistency with rest of code base. Also maybe convert wei into gwei.

monitor/xfeemngr/oracle.go
```diff
@@ -120,6 +120,7 @@ func (o feeOracle) syncGasPrice(ctx context.Context, dest evmchain.Metadata) err
 
 	shielded := withGasPriceShield(buffered)
 
+	log.Info(ctx, "Syncing gas price", "buffered", buffered, "buffered_w_shield", shielded, "on_chain", onChain.Uint64())
```
corverroos, line 123: stick to `chain` vs `src_chain` vs `dest_chain`

corverroos (approved):

PR #2099, opened 2024-10-08
chore(contracts/allocs): lock omega allocs to omega genesis

kevinhalliday:
Lock omega allocs to those used at deployment.

issue: none

### DIFF
--- a/contracts/allocs/scripts/genallocs.go
+++ b/contracts/allocs/scripts/genallocs.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 
+	"github.com/omni-network/omni/contracts/allocs"
 	"github.com/omni-network/omni/contracts/bindings"
 	"github.com/omni-network/omni/e2e/app/eoa"
 	"github.com/omni-network/omni/halo/genutil/evm"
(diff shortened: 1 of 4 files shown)


corverroos (approved):

contracts/allocs/scripts/genallocs.go
```diff
@@ -32,12 +32,21 @@ func main() {
 	fmt.Println("genallocs succeeded")
 }
 
+// After a persistent network has been deployed, lock it's allocs.
+var locked = map[netconf.ID]bool{
```
kc1116, line 36:
```suggestion
// locked map of network configuration IDs for persistent networks.
// Once a persistent network has been generated and deployed, its allocations
// should remain unchanged to ensure consistency and stability across the system.
var locked = map[netconf.ID]bool{
```

kc1116, line 36: Should we move this map to the `allocs` package so it lives with the actual persistent allocs?

kc1116 (approved):

contracts/allocs/scripts/genallocs.go
kc1116: And export a func like this 

`func IsLocked(id netconf.ID) bool `

sideninja (approved):

contracts/allocs/scripts/genallocs.go
kevinhalliday: yeah I like that

PR #2117, opened 2024-10-10
fix(octane/keeper): retry fecthing evm event logs

corverroos:
Fixes a possible chain halting issue when a networking error causes fetching evm event logs to fail.

Wrap this like all other network calls inside keeper logic inside a `retryForever`.

issue: none

### DIFF
--- a/octane/evmengine/keeper/evmmsgs.go
+++ b/octane/evmengine/keeper/evmmsgs.go
@@ -22,12 +22,12 @@ func (k *Keeper) evmEvents(ctx context.Context, blockHash common.Hash) ([]*types
 			ll, err := proc.Prepare(ctx, blockHash)
 			if err != nil {
 				log.Warn(ctx, "Failed fetching evm events (will retry)", err, "proc", proc.Name())
-				return false, nil
+				return false, nil // Retry
 			}
 
 			events = append(events, ll...)
 
-			return true, nil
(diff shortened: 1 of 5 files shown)


octane/evmengine/keeper/evmmsgs.go
```diff
@@ -16,17 +17,25 @@ import (
...
+		err := retryForever(ctx, func(ctx context.Context) (bool, error) {
+			ll, err := proc.Prepare(ctx, blockHash)
+			if err != nil {
+				log.Warn(ctx, "Failed fetching evm logs (will retry)", err, "proc", proc)
+				return false, nil
 			}
-		}
+			// Verify all logs
+			for _, l := range ll {
+				if err := l.Verify(); err != nil {
```
chmllr, line 28: It looks like this `err` is not a transient but a deterministic one, so retrying seems to be futile. Or could we have topics without an `err` in line `21` that are not valid when we fetch them first, but are valid on a retry fetch?

corverroos: returning an error doesn't retry. but sure, I'll move it down for clarity

corverroos: I added some more comments to make this clear

chmllr (approved):

PR #2120, opened 2024-10-10
fix(provider): potential nil pointer dereference

kc1116:
This PR addresses a potential panic issue in the node's handling of the msgs slice in the xblock response. If any message in the msgs slice is nil, attempting to access msg.ShardID will lead to a panic. In a Byzantine environment, it’s crucial to consider that a malicious client could return a compromised block response, which may include nil messages. Implementing this check helps safeguard the node against such scenarios, enhancing its resilience against potential attacks.

issue: none


kc1116:
resolves sigma-prime audit OM2-01

lib/cchain/provider/xblock.go
```diff
@@ -40,6 +40,12 @@ func (p Provider) XBlock(ctx context.Context, height uint64, latest bool) (xchai
 
 	var msgs []xchain.Msg
 	for _, msg := range block.Msgs {
+		// msg will only be nil if receive a malicious response from a client
+		// nil check ensures we won't panic in that case.
+		if msg == nil {
+			return xchain.Block{}, false, errors.New("unexpected nil msg in block response msgs slice possible malicious response [BUG]")
```
corverroos, line 46: nit: please be less verbose, suggest: `invalid nil msg [BUG]`

corverroos: #lessismore

corverroos: you can also add a check for empty msgs:
```
block, ok, err := p.portalBlock(ctx, height, latest)
	if err != nil {
		return xchain.Block{}, false, err
	} else if !ok {
		return xchain.Block{}, false, nil
	} else if !latest && block.Id != height {
		return xchain.Block{}, false, errors.New("unexpected block height [BUG]")
	} else if len(block.Msgs) == 0 {
   		return xchain.Block{}, false, errors.New("unexpected empty block [BUG]")
	}
```

corverroos (approved):

sideninja (approved):

chmllr (approved):

PR #2121, opened 2024-10-10
fix(cchain): check length of publickey after unmarshal

kc1116:
This PR adds a check for the size of the public key ensuring it is the expected secp256k1.PubKeySize (33) after unmarshaling the public key. This prevents unexpected panics. 

issue: none

### DIFF
--- a/lib/cchain/types.go
+++ b/lib/cchain/types.go
@@ -8,6 +8,7 @@ import (
 	"github.com/omni-network/omni/lib/umath"
 
 	cmtcrypto "github.com/cometbft/cometbft/crypto"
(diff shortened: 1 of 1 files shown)


lib/cchain/types.go
```diff
@@ -101,6 +102,10 @@ func (v SDKValidator) ConsensusCmtAddr() (cmtcrypto.Address, error) {
 		return nil, errors.Wrap(err, "unmarshal consensus pubkey")
 	}
 
+	if len(pk.Bytes()) != k1.PubKeySize {
```
corverroos, line 105: also add it to `ConsensusPublicKey`

corverroos (approved):

PR #2122, opened 2024-10-10
refactor(e2e/eoa): Setup mainnet roles and eoa accounts with Fireblocks 

sideninja:
Setup codebase to work with new Fireblock setup for mainnet as well as updated staging vaults. Omega setup stays the same although we introduce few new roles they map to the same vault on Fireblocks.

New vaults roles introduced:
`RoleManager` (previously `RoleAdmin` but split into two: manager and upgrader), used to pause and unpause contracts
`RoleUpgrader` used to upgrade contracts and registering new portals
`RoleFuner` is a new role extracted from before hard-coded single value for all networks

issue: https://github.com/omni-network/ops/issues/483

### DIFF
--- a/cli/cmd/register_test.go
+++ b/cli/cmd/register_test.go
@@ -141,7 +141,7 @@ func testDeployCfg(t *testing.T) deployConfig {
 
 	return deployConfig{
 		deployer:         eoa.MustAddress(netconf.Devnet, eoa.RoleDeployer),
-		owner:            eoa.MustAddress(netconf.Devnet, eoa.RoleOwner),
+		owner:            eoa.MustAddress(netconf.Devnet, eoa.RoleManager),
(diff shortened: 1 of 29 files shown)


kevinhalliday (commented):
For stuff in `lib/contracts`, can you keep 

```
Owner: eoa.MustAddress(network, role.Manger)
ProxyAdminOwner: eoa.MustAddress(network, role.Upgrader)
```

The names "owner" and "proxy admin owner" have meaning in the contract context. Leaving the names makes what we are doing

assigning RoleManager to be the contract owner. And assigning RoleUpgrader to proxy admin owner.

(ProxyAdmin is a type of contract, and the owner of it is allowed to upgrade)

sideninja:
@kevinhalliday will update, sorry had to open another PR https://github.com/omni-network/omni/pull/2123

PR #2123, opened 2024-10-10
refactor(e2e): setup new fireblocks vaults for mainnet

sideninja:
Setup codebase to work with new Fireblock setup for mainnet as well as updated staging vaults. Omega setup stays the same although we introduce few new roles they map to the same vault on Fireblocks.

New vaults roles introduced:
`RoleManager` (previously `RoleAdmin` but split into two: manager and upgrader), used to pause and unpause contracts
`RoleUpgrader` used to upgrade contracts and registering new portals
`RoleFuner` is a new role extracted from before hard-coded single value for all networks

issue: none


cli/cmd/register_test.go
```diff
@@ -141,7 +141,7 @@ func testDeployCfg(t *testing.T) deployConfig {
 
 	return deployConfig{
 		deployer:         eoa.MustAddress(netconf.Devnet, eoa.RoleDeployer),
-		owner:            eoa.MustAddress(netconf.Devnet, eoa.RoleAdmin),
+		manager:          eoa.MustAddress(netconf.Devnet, eoa.RoleManager),
```
corverroos, line 144: doesn't seem like we configure the `upgrader` in the contracts we deploy? So I don't think it will be able to upgrade anything....

corverroos (changes requested):

cli/cmd/register_test.go
sideninja: you are right, this should be the upgrader because it gets bound to the proxy contract, fixed: 72555a2241dfd86df0aab1f6e5439c71dcc46d37

contracts/allocs/scripts/genallocs.go
```diff
@@ -105,7 +105,7 @@ func allocConfig(network netconf.ID, prefunds types.GenesisAlloc) (bindings.Allo
 	}
 
 	return bindings.AllocPredeploysConfig{
-		Admin:                  eoa.MustAddress(network, eoa.RoleAdmin),
+		Admin:                  eoa.MustAddress(network, eoa.RoleManager),
```
corverroos, line 108: this sets the manager as owner of all predeployed I think, while we want to some to be owned by the upgrader. Like octane predeployes. Maybe even all predeployes, I'm not sure about OmniNativeBridge.

kevinhalliday: Yes we need to have a split.  Which means updating forge script. I can handle that on add to this PR

sideninja: I'm doing it no worries

sideninja: I will just need you to likely correct me after @kevinhalliday :D

kevinhalliday (approved):

PR #2130, opened 2024-10-10
refactor(proto): non-nullable proto arrays

corverroos:
Refactors all cosmos module query protos array fields to be non-nullable. This changes the generated types from pointers to non-points. This avoid the need for nil checks.

Nice thing is that the wire serialization remain identical, only runtime types are different, which means this is backwards compatible!

issue: none


lib/cchain/provider/testdata/TestXBlock.golden
```diff
@@ -5,12 +5,12 @@
  "Msgs": [
   {
    "SourceChainID": 77,
-   "DestChainID": 1894687307556873581,
```
corverroos, line 8: this change is just due to fuzzing that follows a different path and doesn't break things on the wire.

octane/evmengine/types/tx.proto
```diff
@@ -25,7 +26,7 @@ message MsgExecutionPayload {
   option (cosmos.msg.v1.signer) = "authority";
   string            authority           = 1;
   bytes             execution_payload   = 2;
-  repeated EVMEvent prev_payload_events = 3;
+  repeated EVMEvent prev_payload_events = 3 [(gogoproto.nullable) = false];
```
sideninja, line 29: 💸

sideninja (approved):
very nice

PR #2131, opened 2024-10-10
chose(ci): update pnpm and add make install tools

sideninja:
This PR adds make action to install PNPM version that matches the version in CI so no lockfile conflicts are caused.
It also makes installing tools easier.

issue: none


kevinhalliday (changes requested):
There are a couple other places in github workflows we specify pnpm version - please update there. (would be nice of we didn't have to change a bunch of places but alas)

Also in each package.json we specify a `packageManager` - please update those as well.

Also I think we'll need to remake all of our lockfiles. You can probably do that with a `pnpm i` in each place we have one

PR #2137, opened 2024-10-10
fix(xchain): add signature validation 

kc1116:
Add signature validation to SubmissionToBinding and SubmissionFromBinding. 

issue: none


kc1116:
resolves sigma-prime audit OM2-10

lib/tutil/fixtures.go
```diff
@@ -0,0 +1,20 @@
...
+	"crypto/ecdsa"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+
+	"github.com/stretchr/testify/require"
+)
+
+func PrivateKeyFixture(t *testing.T) (*ecdsa.PrivateKey, common.Address) {
```
corverroos, line 13: add a godoc comment please

lib/xchain/abi.go
```diff
@@ -147,14 +154,20 @@ func SubmissionFromBinding(sub bindings.XSubmission, destChainID uint64) (Submis
 }
 
 // SubmissionToBinding converts a go xchain submission to a solidity binding submission.
-func SubmissionToBinding(sub Submission) bindings.XSubmission {
+func SubmissionToBinding(sub Submission) (bindings.XSubmission, error) {
```
corverroos, line 157: the best place to verify data is when you get it over the wire (or any IO).

a submission is a in-memory calculated type. Rather very the attestation we get from cprovider in `Worker::newCallback` via something like: `att.Verify()`.

kc1116: https://github.com/omni-network/omni/pull/2137/commits/2a5e8b3dc89eee068f432c54e1de4f66ed932cb2

lib/tutil/fixtures.go
kc1116: https://github.com/omni-network/omni/pull/2137/commits/a1a4c44b698dcf6981ec2bf9bdd9a6de513e077c

lib/tutil/fixtures.go
```diff
@@ -0,0 +1,21 @@
...
+	"crypto/ecdsa"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+
+	"github.com/stretchr/testify/require"
+)
+
+// PrivateKeyFixture generates an ethereum ecds private key.
```
corverroos, line 13: nit: `PrivateKeyFixture returns a new random ethereum ecdsa private key and associated address`

lib/xchain/types.go
```diff
@@ -248,6 +249,24 @@ func (a Attestation) AttestationRoot() ([32]byte, error) {
...
 
+// Verify verifies all the signatures in the attestation against the sigtuple validator address.
+func (a Attestation) Verify() (bool, error) {
+	attRoot, err := a.AttestationRoot()
+	if err != nil {
+		return false, err
+	}
+
+	for _, sig := range a.Signatures {
+		if ok, err := k1util.Verify(sig.ValidatorAddress, attRoot, sig.Signature); !ok {
```
corverroos, line 260: **_always_** check `err` first.

lib/xchain/types.go
```diff
@@ -248,6 +249,24 @@ func (a Attestation) AttestationRoot() ([32]byte, error) {
...
 }
 
+// Verify verifies all the signatures in the attestation against the sigtuple validator address.
+func (a Attestation) Verify() (bool, error) {
+	attRoot, err := a.AttestationRoot()
+	if err != nil {
+		return false, err
+	}
+
+	for _, sig := range a.Signatures {
```
corverroos, line 259: also ensure it doesn't contain duplicate signatures

lib/xchain/types.go
```diff
@@ -248,6 +249,24 @@ func (a Attestation) AttestationRoot() ([32]byte, error) {
 	return AttestationRoot(a.AttestHeader, a.BlockHeader, a.MsgRoot)
 }
 
+// Verify verifies all the signatures in the attestation against the sigtuple validator address.
+func (a Attestation) Verify() (bool, error) {
```
corverroos, line 253: just return an error, no need for the bool

lib/xchain/types.go
```diff
@@ -248,6 +249,24 @@ func (a Attestation) AttestationRoot() ([32]byte, error) {
 	return AttestationRoot(a.AttestHeader, a.BlockHeader, a.MsgRoot)
 }
 
+// Verify verifies all the signatures in the attestation against the sigtuple validator address.
```
corverroos, line 252: nit: don't repeat the logic of the function of the comments. It will just get out of date. Rather focus on the return types.

e.g. `Verify returns an error if the attestation is invalid.`

relayer/app/worker.go
```diff
@@ -191,6 +191,12 @@ func (w *Worker) newCallback(
 	msgStreamMapper msgStreamMapper,
 ) cchain.ProviderCallback {
 	return func(ctx context.Context, att xchain.Attestation) error {
+		if ok, err := att.Verify(); !ok {
```
corverroos, line 194: I had a look at the cprovider where it receives the data over the wire in `cchain/provider/abc.go`. We already do this in `atypes.AttestationFromProto` which is called as soon as we get the data over the wire. I think that is actually sufficient. As that is the point where we get the data over the wire. So I don't think we need to perform duplicate verification.

corverroos: We should probably add "duplicate" signature verification to `halo/attest/types/tx.go::Attestation.Verify`

corverroos (changes requested):

relayer/app/worker.go
kc1116: https://github.com/omni-network/omni/pull/2137/commits/78b35136c5ff9dc7b8c2efcdbd6a8a24f3b87431

corverroos (approved):

PR #2140, opened 2024-10-10
fix(lib/contracts): add mutex to caches

kevinhalliday:
Add mutex to caches.

issue: none

### DIFF
--- a/lib/contracts/address.go
+++ b/lib/contracts/address.go
@@ -3,6 +3,7 @@ package contracts
 import (
 	"context"
 	"math/big"
+	"sync"
 
 	"github.com/omni-network/omni/e2e/app/eoa"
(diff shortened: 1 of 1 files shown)


lib/contracts/address.go
```diff
@@ -94,17 +95,31 @@ type Salts struct {
...
+type cache[T any] struct {
+	mu    sync.Mutex
+	cache T
+}
+
 var (
 	// cached addresses by network.
-	addrsCache = map[netconf.ID]Addresses{}
+	addrsCache = cache[map[netconf.ID]Addresses]{
+		mu:    sync.Mutex{},
```
sideninja, line 106: no need to initialize it, empty value is fine
```suggestion
```

PR #2147, opened 2024-10-11
chore(admin): set omega uluwatu upgrade height

corverroos:
Sets the omega `uluwatu` network-upgrade height to 3_073_000 which is roughly 9am Mon 14 Oct EST.

issue: https://github.com/omni-network/ops/issues/499


corverroos:
image

fabtreb (approved):

PR #2148, opened 2024-10-11
chore(lib): validate inputs before locking

chmllr:
The PR https://github.com/omni-network/omni/pull/2140 introduced a locking around a cache. Some parts of the critical section seem unrelated to the locking and can be done before without changing the read semantics (because this check happened after the cache read before).

issue: none

### DIFF
--- a/lib/contracts/address.go
+++ b/lib/contracts/address.go
@@ -114,17 +114,17 @@ var (
 
 // GetAddresses returns the contract addresses for the given network.
 func GetAddresses(ctx context.Context, network netconf.ID) (Addresses, error) {
+	ver, err := version(ctx, network)
+	if err != nil {
+		return Addresses{}, err
+	}
+
 	addrsCache.mu.Lock()
 	defer addrsCache.mu.Unlock()
 
 	addrs, ok := addrsCache.cache[network]
 	if ok {
 		return addrs, nil
-	}
-
-	ver, err := version(ctx, network)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


chmllr:
@kevinhalliday I have another question to the original PR: how critical are lock contentions on `GetAddresses`? The original PR uses a simple mutex, while we could also use an RW-lock and separate cache reads and cache updates so that at least concurrent reads (which will be the most frequent accesses) are not blocking each other.

corverroos (approved):

corverroos:
@chmllr RWMutex should only be used for performance optimisation, I would not worry about that in that case. Performance is a non-issue. Lets keep things simple.

chmllr:
> @chmllr RWMutex should only be used for performance optimisation, I would not worry about that in that case. Performance is a non-issue. Lets keep things simple.
> 
> I would even argue that time spent on this PR by multiple people isn't worth the effort. The value it adds is insignificant and it feels like we are making the code harder to read.
> 
> A simple method with lock-defer-unlock as first line is common and standard and doesn't need much thinking. As soon as you move the lock around, readers of the code need to think why is this done here and not there, etc.
> 
> I wouldn't worry about these type of optimisations.

Thanks Corver! Fully agree with optimizing for simplicity first, that's why I asked how critical this contention is. Because the fact that we _have_ a cache means we do have a lot of reads. So if they all wait for each other now, might not be what we want but here I fully trust your intuition 👍

PR #2150, opened 2024-10-11
test(contracts/core): fix predeploy num alloc tests

kevinhalliday:
Num alloc test had correct number but wrong breakdown.

- We have 5 predeploy impls (Staking, Slashing, Upgrades, PortalReg, BridgeNative)
- 15 preinstall contracts
- 1 account we bump nonce off

issue: none


sideninja (approved):

chmllr:
Would it make sense to actually declare static arrays with strings referring to things we count and then use array lengths? I hope the Solidity compiler is smart enough to optimize it to a constant...

chmllr (approved):

kevinhalliday:
It's pretty long - it would be like a list of 2000 "proxy admin" and 2000 "transparent proxies"

PR #2155, opened 2024-10-11
feat(*): non zero system addresses

kevinhalliday:
Use non-zero system addresses.

Marking as draft because more thought would need to go into supporting omega upgrade. This is just for consideration. 

Reasoning:

- these two should be distinct values (so we can differentiate by value, not just by context)
- zero addr is already pretty overloaded
- we can tag these explicitly in our explorer / other things
- no need to juggle "zero on purpose" vs "zero cause empty" (explicit  > implicit)

issue: none


kevinhalliday:
Closing - will revisit later

PR #2156, opened 2024-10-11
chore(*): confirm xmsgs ordered by log index

kevinhalliday:
XMsgs are ordered by log index, which implicitly orders them by 

Update contract test utils to reflect that.  
Enforce this order in lib/xchain/provider.

issue: none


contracts/core/test/xchain/ts/merkle.ts
```diff
@@ -45,13 +45,11 @@ export class XMsgMerkleTree {
     return tree[0]
   }
 
-  // Order msgs by destination chain id, then offset
+  // Order msgs by offset. XMsgs within an xblock are ordered by log index, which implicity orders
+  // by offset (ascending).
```
corverroos, line 49: Note this is only correct for blocks with msgs for a single stream. It isn't correct for blocks with msgs for multiple streams.

PR #2170, opened 2024-10-14
chore(lib/cmd): redact rpc url tokens

corverroos:
Best effort redaction of tokens in URL paths.

issue: https://github.com/omni-network/ops/issues/170

### DIFF
--- a/lib/cmd/flags.go
+++ b/lib/cmd/flags.go
@@ -2,7 +2,10 @@ package cmd
 
 import (
 	"context"
+	"encoding/hex"
+	"fmt"
 	"net/url"
+	"path/filepath"
 	"strings"
 
 	"github.com/omni-network/omni/lib/log"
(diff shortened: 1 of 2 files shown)


lib/cmd/flags.go
```diff
@@ -88,3 +92,16 @@ func redact(flag, val string) string {
 
 	return val
 }
+
+// maybeRedactToken redacts the last element of the given path
+// if it is a hex token of more than 16 chars.
```
chmllr, line 99: Technically 15...

chmllr: Should we add more context like "to remove secrets"? I'd never guess why we do this.

chmllr (approved):

lib/cmd/flags.go
corverroos: see:
```
// redact returns a redacted version of the given flag value. It currently supports redacting
// passwords in valid URLs as well as flags that contains words like "token", "password", "secret", "db" or "key".
func redact(flag, val string) string {
```

Issue #2078, opened 2024-10-04
Add self-delegate omni cli command


corverroos:
#### Problem to Solve
We need to increase operator power for omega. But we don't have a cli command to do it.

#### Proposed Solution
- Following the example of `omni operator create-validator`.
- Add a `omni operator self-delegate` command.

PR #2174, opened 2024-10-14
feat(cli): added self-delegate CLI command

sideninja:
The PR adds a new self-delegate CLI command that can be used like so:
```
omni operator delegate --network "" --amount "" --private-key-file ""
```

The command will increase the delegation of the validator by the amount provided as the flag.

The PR also introduces new CLI e2e test suite that was used to test this command. The test currently creates a new validator using the create validator CLI command and then delegates more stake using the delegate command.

The output of the test:
```
=== RUN   TestValidatorCommands/network
=== PAUSE TestValidatorCommands/network
=== CONT  TestValidatorCommands/network
24-10-15 00:23:29.258 DEBU Txmgr initialized nonce                  dest_chain=omni_evm from_addr=0x976EA740 req_id=bf52d7c chain=omni_evm nonce=0
24-10-15 00:23:29.265 DEBU Crafted tx                               dest_chain=omni_evm from_addr=0x976EA740 req_id=bf52d7c chain=omni_evm sender=0x976EA74026E726554dB657fA54763abd0C3a0aa9 nonce=0 tx=0xb4976a6227d9e7a60e5c05261c33930da13a1e9b837c306127a8b3774f49582b
24-10-15 00:23:30.621 DEBU Backend sent tx                          dest_chain=omni_evm from_addr=0x976EA740 req_id=bf52d7c chain=omni_evm nonce=0 gas_used=32234 status=1 height=121 tx=0xb4976a6227d9e7a60e5c05261c33930da13a1e9b837c306127a8b3774f49582b
24-10-15 00:23:30.625 INFO 🎉 Create-validator transaction sent and included on-chain link=https://devnet.omniscan.network/tx/0xb4976a6227d9e7a60e5c05261c33930da13a1e9b837c306127a8b3774f49582b block=121
24-10-15 00:23:31.670 DEBU Txmgr initialized nonce                  dest_chain=omni_evm from_addr=0x976EA740 req_id=2216d14 chain=omni_evm nonce=1
24-10-15 00:23:31.676 DEBU Crafted tx                               dest_chain=omni_evm from_addr=0x976EA740 req_id=2216d14 chain=omni_evm sender=0x976EA74026E726554dB657fA54763abd0C3a0aa9 nonce=1 tx=0x3f6632feb03d0cad34c537317f416c6bd1edff8db3f0c4326c747aab2b892a42
24-10-15 00:23:32.356 DEBU Backend sent tx                          dest_chain=omni_evm from_addr=0x976EA740 req_id=2216d14 chain=omni_evm nonce=1 gas_used=28624 status=1 height=123 tx=0x3f6632feb03d0cad34c537317f416c6bd1edff8db3f0c4326c747aab2b892a42
24-10-15 00:23:32.360 INFO 🎉 Delegate increase transaction sent and included on-chain link=https://devnet.omniscan.network/tx/0x3f6632feb03d0cad34c537317f416c6bd1edff8db3f0c4326c747aab2b892a42 block=123
--- PASS: TestValidatorCommands (0.02s)
    --- PASS: TestValidatorCommands/network (4.67s)
```

issue: #2078


halo/genutil/testdata/TestMakeGenesis.golden
```diff
@@ -42,7 +42,7 @@
      "coins": [
```
sideninja, line 42: this changed because we increased minimum power from 100 to 1000

cli/cmd/staking.go
```diff
@@ -245,6 +359,8 @@ func unjailValidator(ctx context.Context, cfg unjailConfig) error {
 	}
 	opAddr := crypto.PubkeyToAddress(opPrivKey.PublicKey)
 
+	// todo reuse setupClient
```
sideninja, line 362: will refactor in another PR

e2e/test/cli_test.go
```diff
@@ -0,0 +1,138 @@
...
+		validatorAddr := ethcrypto.PubkeyToAddress(validatorPriv.PublicKey)
+		tmpDir := t.TempDir()
+		pkeyFile := filepath.Join(tmpDir, "pkey")
+		require.NoError(
+			t,
+			ethcrypto.SaveECDSA(pkeyFile, validatorPriv),
+			"failed to save new validator private key to temp file",
+		)
+
+		cl, err := http.New(testnet.Network.Static().ConsensusRPC(), "/websocket")
```
corverroos, line 77: rather panic in `Static().ConsensusRPC` for devnet

e2e/test/cli_test.go
```diff
@@ -0,0 +1,138 @@
...
+		require.True(t, ok)
+		executionRPC, err := endpoints.ByNameOrID(e.Name, e.ID)
+		require.NoError(t, err)
+
+		// use an existing test anvil account for new validator and write it's pkey to temp file
+		validatorPriv := anvil.DevPrivateKey6()
+		validatorPub := ethcrypto.CompressPubkey(&validatorPriv.PublicKey)
+		validatorAddr := ethcrypto.PubkeyToAddress(validatorPriv.PublicKey)
+		tmpDir := t.TempDir()
+		pkeyFile := filepath.Join(tmpDir, "pkey")
```
corverroos, line 70: `pkey` is ambigious, rather do `privkey` or `pubkey`

e2e/test/cli_test.go
```diff
@@ -0,0 +1,138 @@
...
+		testnet, _, _, _ := loadEnv(t)
+
+		e, ok := network.OmniEVMChain()
+		require.True(t, ok)
+		executionRPC, err := endpoints.ByNameOrID(e.Name, e.ID)
+		require.NoError(t, err)
+
+		// use an existing test anvil account for new validator and write it's pkey to temp file
+		validatorPriv := anvil.DevPrivateKey6()
+		validatorPub := ethcrypto.CompressPubkey(&validatorPriv.PublicKey)
```
corverroos, line 67: rather indicate if something is bytes by adding a `bz` suffix (cometBFT style)

e2e/test/cli_test.go
```diff
@@ -0,0 +1,138 @@
...
+		require.NoError(
+			t,
+			ethcrypto.SaveECDSA(pkeyFile, validatorPriv),
+			"failed to save new validator private key to temp file",
+		)
+
+		cl, err := http.New(testnet.Network.Static().ConsensusRPC(), "/websocket")
+		require.NoError(t, err)
+
+		cprov := provider.NewABCIProvider(cl, network.ID, netconf.ChainVersionNamer(network.ID))
```
corverroos, line 80: rather create things when they are needed, so move this down

e2e/test/cli_test.go
```diff
@@ -0,0 +1,138 @@
...
+			// we use minimum stake so the new validator doesn't affect the network too much
+			"--self-delegation", fmt.Sprintf("%d", delegation),
+			"--evm-rpc", executionRPC,
+		)
+		require.NoError(t, err)
+		require.Empty(t, res)
+
+		// wait for validator to be created
+		require.Eventuallyf(t, func() bool {
+			_, ok, _ := cprov.SDKValidator(context.Background(), validatorAddr)
```
corverroos, line 98: rather use common `ctx`, define one at the top of the test

e2e/test/cli_test.go
```diff
@@ -0,0 +1,138 @@
...
+		require.Equal(t, delegation, power)
+
+		// operator delegate test
+
+		// delegate more stake for the validator, since we are using an anvil account
+		// it is already sufficiently funded
+		res, err = exeCLI(
+			"operator",
+			"delegate",
+			"--network",
```
corverroos, line 117: maybe put flags and values in same line for easy readability? `"--network", "devnet",`

e2e/test/cli_test.go
```diff
@@ -0,0 +1,138 @@
...
+			"devnet",
+			"--private-key-file",
+			pkeyFile,
+			"--amount",
+			"5",
```
corverroos, line 122: do minimum rather

e2e/test/cli_test.go
```diff
@@ -0,0 +1,138 @@
...
+			"--evm-rpc", executionRPC,
+		)
+		require.NoError(t, err)
+		require.Empty(t, res)
+
+		// make sure the validator power is actually increased
+		require.Eventuallyf(t, func() bool {
+			val, ok, _ := cprov.SDKValidator(context.Background(), validatorAddr)
+			require.True(t, ok)
```
corverroos, line 131: don't use require inside eventual block, since the test will fail immediately, rather just return true or false

e2e/test/cli_test.go
```diff
@@ -0,0 +1,138 @@
...
+			newPower, err := val.Power()
+			require.NoError(t, err)
+
+			return newPower > power
+		}, 5*time.Second, 500*time.Millisecond, "failed to create validator")
```
corverroos, line 136: 5s isn't much, I'd increase to avoid flapping tests

e2e/test/cli_test.go
```diff
@@ -0,0 +1,138 @@
...
+// TestValidatorCommands tests multiple CLI operator commands in sequence.
+// The test runs the following commands:
+// - operator create-validator creates a new validator and makes sure the validator is added to the consensus chain
+// - operator delegate increases the newly created validator stake and makes sure its power is increased
+//
+// Since they rely first on validator being created it must be run as a unit.
+func TestValidatorCommands(t *testing.T) {
+	t.Parallel()
+
+	testNetwork(t, func(t *testing.T, network netconf.Network, endpoints xchain.RPCEndpoints) {
```
corverroos, line 56: why not have a nice `testCLI` with signature: `func(t *testing.T, omniEVMCl ethclient.Client, rpcClient *comethttp.HTTP)`?

You can do queries and stuff to both omni evm and omni conensus using those clients, construct xproviders or cproviders if required, and they both expose internal addresses that you should pass to the cli command.

cli/cmd/flags.go
```diff
@@ -64,19 +66,38 @@ func bindRPCURL(cmd *cobra.Command, rpcURL *string) {
...
+	_ = cmd.MarkFlagRequired(flagNetwork)
+}
+
+func bindDelegateValConfig(cmd *cobra.Command, cfg *delegateValConfig) {
+	netconf.BindFlag(cmd.Flags(), &cfg.network)
+	bindPrivateKeyFile(cmd, &cfg.privateKeyFile)
+
+	cmd.Flags().Uint64Var(&cfg.amount, flagDelegationAmount, cfg.amount, "increase delegation amount in OMNI (minimum 1 OMNI)")
+
+	bindExecutionRPC(cmd, &cfg.evmRPC)
```
corverroos, line 78: lets create a `bindEOAConfig` function that is shared between all commands that embed eoaConfig.

cli/cmd/flags.go
```diff
@@ -64,19 +66,38 @@ func bindRPCURL(cmd *cobra.Command, rpcURL *string) {
...
 
 	_ = cmd.MarkFlagRequired(flagConsPubKeyHex)
 	_ = cmd.MarkFlagRequired(flagSelfDelegation)
-	_ = cmd.MarkFlagRequired("network")
+	_ = cmd.MarkFlagRequired(flagNetwork)
+}
+
+func bindExecutionRPC(cmd *cobra.Command, evmRPC *string) {
+	cmd.Flags().StringVar(evmRPC, "evm-rpc", "", "Omni EVM execution RPC API endpoint")
```
corverroos, line 100: `--execution-rpc` and `--consensus-rpc`. Please add a comment that that this is optional. And something like `consensus.<network>.omni.network` will be used if empty.

corverroos (changes requested):

cli/cmd/staking.go
```diff
@@ -311,3 +424,41 @@ func unjailValidator(ctx context.Context, cfg unjailConfig) error {
...
+	chainMeta, ok := evmchain.MetadataByID(chainID)
+	if !ok {
+		return nil, nil, nil, errors.New("chain metadata not found")
+	}
+
+	if evmRPC == "" {
+		evmRPC = network.Static().ExecutionRPC()
+	}
+
+	cprov, err := provider.Dial(network)
```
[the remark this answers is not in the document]
sideninja: That is actually more problematic than it seems. The problem is it's being used in a lot of places and if we panic it fails a lot of tests. The other thing is that after investigating the consensus rpc default is correct for devnet, and if a test wants to specify it differently it can by creating the client manually. I did that first but then reverted this changes because I see all the issue it causes in other tests as well as figured out the default is the same as I was setting. So it didn't make sense to change this.

e2e/test/cli_test.go
sideninja: see the above comment, we shouldn't

e2e/test/cli_test.go
sideninja: what is bz?

e2e/test/cli_test.go
sideninja: but the test should fail, because if the validator doesn't exist at this point, we shouldn't wait for it

e2e/test/cli_test.go
sideninja: we can, I was trying to wrap things up and we can do that when we test more commands and know what will be used by other commands as well, I can change this here if you think it's better

cli/cmd/staking.go
```diff
@@ -81,85 +112,170 @@ func (c createValConfig) ConsensusPubKey() (*ecdsa.PublicKey, error) {
...
+		return errors.Wrap(err, "wait mined")
+	}
+
+	link := fmt.Sprintf("https://%s.omniscan.network/tx/%s", cfg.network, tx.Hash().Hex())
+	log.Info(ctx, "🎉 Create-validator transaction sent and included on-chain", "link", link, "block", rec.BlockNumber.Uint64())
+
+	return nil
+}
+
+type delegateValConfig struct {
```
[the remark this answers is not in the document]
sideninja: I was thinking we would introduce this once we have that option. I can add them now.

cli/cmd/flags.go
```diff
@@ -64,19 +66,37 @@ func bindRPCURL(cmd *cobra.Command, rpcURL *string) {
...
 	bindPrivateKeyFile(cmd, &cfg.PrivateKeyFile)
-	_ = cmd.MarkFlagRequired("network")
+	_ = cmd.MarkFlagRequired(flagNetwork)
+}
+
+func bindEOAConfig(cmd *cobra.Command, cfg *eoaConfig) {
+	bindPrivateKeyFile(cmd, &cfg.privateKeyFile)
+	netconf.BindFlag(cmd.Flags(), &cfg.network)
+
+	cmd.Flags().StringVar(&cfg.evmRPC, "evm-rpc", "", "Optional Omni EVM execution RPC API endpoint. Defaults to consensus.<network>.omni.network")
```
corverroos, line 76: lets rename this to `execution-rpc`, and lets add `consensus-rpc`. Users might want to specify their own halo node.

cli/cmd/flags.go
```diff
@@ -64,19 +66,37 @@ func bindRPCURL(cmd *cobra.Command, rpcURL *string) {
...
+
+	_ = cmd.MarkFlagRequired(flagNetwork)
+}
+
+func bindDelegateConfig(cmd *cobra.Command, cfg *delegateConfig) {
+	bindEOAConfig(cmd, &cfg.eoaConfig)
+	const flagSelf = "self"
+	cmd.Flags().Uint64Var(&cfg.amount, flagDelegationAmount, cfg.amount, "Increase delegation amount in OMNI (minimum 1 OMNI)")
+	cmd.Flags().BoolVar(&cfg.self, flagSelf, true, "Mark self-delegation")
```
corverroos, line 85: lets do default false, since this will be the default in the future, so then we don't need to change defaults later.

cli/cmd/flags.go
```diff
@@ -64,19 +66,37 @@ func bindRPCURL(cmd *cobra.Command, rpcURL *string) {
...
+
+	cmd.Flags().StringVar(&cfg.evmRPC, "evm-rpc", "", "Optional Omni EVM execution RPC API endpoint. Defaults to consensus.<network>.omni.network")
+
+	_ = cmd.MarkFlagRequired(flagNetwork)
+}
+
+func bindDelegateConfig(cmd *cobra.Command, cfg *delegateConfig) {
+	bindEOAConfig(cmd, &cfg.eoaConfig)
+	const flagSelf = "self"
+	cmd.Flags().Uint64Var(&cfg.amount, flagDelegationAmount, cfg.amount, "Increase delegation amount in OMNI (minimum 1 OMNI)")
```
corverroos, line 84: lets remove the word "increase"

cli/cmd/flags.go
corverroos: also maybe change the comment: `Enables self-delegation setting target validator address to provided private key`

cli/cmd/staking.go
```diff
@@ -28,6 +30,7 @@ import (
 )
 
 const minSelfDelegation = uint64(100)
+const minDelegationIncrease = uint64(1)
```
corverroos, line 33: lets remove the term `Increase`

cli/cmd/staking.go
```diff
@@ -56,19 +59,56 @@ func newCreateValCmd() *cobra.Command {
...
+	if _, err := v.privateKey(); err != nil {
+		return errors.Wrap(err, "verify --private-key-file flag")
+	}
+
+	if err := v.network.Verify(); err != nil {
+		return errors.Wrap(err, "verify --network flag")
+	}
+
+	if v.evmRPC != "" {
+		if _, err := url.ParseRequestURI(v.evmRPC); err != nil {
```
corverroos, line 92: why not use simple `url.Parse`?

cli/cmd/staking.go
```diff
@@ -81,85 +121,176 @@ func (c createValConfig) ConsensusPubKey() (*ecdsa.PublicKey, error) {
...
+		Long:  `Delegate an amount of Omni tokens to a validator from your wallet. Only self-delegation by validators supported at the moment.`,
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			if err := cfg.validate(); err != nil {
+				return errors.Wrap(err, "verify flags")
+			}
+
+			err := delegate(cmd.Context(), cfg)
+			if err != nil {
+				return errors.Wrap(err, "create-validator")
```
corverroos, line 251: nit: `delegate`

cli/cmd/staking.go
```diff
@@ -81,85 +121,176 @@ func (c createValConfig) ConsensusPubKey() (*ecdsa.PublicKey, error) {
...
+
+	return cmd
+}
+
+func delegate(ctx context.Context, cfg delegateConfig) error {
+	operatorPriv, err := cfg.privateKey()
+	if err != nil {
+		return err
+	}
+	opAddr := crypto.PubkeyToAddress(operatorPriv.PublicKey)
```
corverroos, line 268: this is `delegatorAddr`

cli/cmd/staking.go
```diff
@@ -81,85 +121,176 @@ func (c createValConfig) ConsensusPubKey() (*ecdsa.PublicKey, error) {
...
+}
+
+type delegateConfig struct {
+	eoaConfig
+	amount uint64
+	self   bool
+}
+
+func (d delegateConfig) validate() error {
+	if d.amount < minDelegationIncrease {
```
corverroos, line 229: should probably add `if !d.self return "only self-delegation supported, please provide --self `

cli/cmd/staking.go
```diff
@@ -81,85 +121,176 @@ func (c createValConfig) ConsensusPubKey() (*ecdsa.PublicKey, error) {
...
+		},
+	}
+
+	bindDelegateConfig(cmd, &cfg)
+
+	return cmd
+}
+
+func delegate(ctx context.Context, cfg delegateConfig) error {
+	operatorPriv, err := cfg.privateKey()
```
corverroos, line 264: technically, this is `delegatorPriv`, and `if cfg.Self`, then `opAddr = address(operatorPriv)`

cli/cmd/staking.go
```diff
@@ -81,85 +121,176 @@ func (c createValConfig) ConsensusPubKey() (*ecdsa.PublicKey, error) {
...
 		return err
-	} else if bal <= float64(cfg.SelfDelegation) {
+	}
+
+	bal, err := eth.EtherBalanceAt(ctx, opAddr)
+	if err != nil {
+		return err
+	} else if bal <= float64(cfg.amount) {
 		return &CliError{
 			Msg:     fmt.Sprintf("Operator address has insufficient balance=%.2f OMNI, address=%s", bal, opAddr),
```
corverroos, line 295: Delegator address has ...

cli/cmd/staking.go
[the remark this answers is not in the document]
corverroos: I still think we should allow setting a `--consensus-rpc` since users might want to use their own halo nodes.

e2e/test/cli_test.go
corverroos: `bz` means `bytes`. It is cometBFT and cosmosSDK style

e2e/test/cli_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+		// operator delegate test
+
+		// delegate more stake for the validator, since we are using an anvil account
+		// it is already sufficiently funded
+		res, err = execCLI(
+			ctx, "operator", "delegate",
+			"--network", "devnet",
+			"--private-key-file", privKeyFile,
+			"--amount", "1",
```
corverroos, line 117: above you use a constant `const delegation = uint64(100)` here you use a inline string "1". Lets choose the same style for consistency

e2e/test/cli_test.go
corverroos: ah right, indeed

halo/genutil/genutil.go
```diff
@@ -48,7 +48,7 @@ const slashingBlocksWindow = 1000
 
 // validatorPower is the default power assigned to genesis validators.
 // It is also the minimum stake enforced by the omni staking contract.
```
corverroos, line 50: this comment is now incorrect, since the minimum self delegation is still 100.

lib/netconf/static.go
```diff
@@ -118,7 +118,7 @@ func (s Static) ExecutionSeeds() []string {
 
 func (s Static) ExecutionRPC() string {
 	if s.Network == Devnet {
-		return "http://localhost:8001"
+		return "http://127.0.0.1:8000"
```
corverroos, line 121: if we don't want to panic, at least return invalid URL so that it is clear that this should not be used and will error. I suggest: `no well known ephemeral rpc`

cli/cmd/staking.go
corverroos: note the error wrap message here is copypaste mistake

cli/cmd/staking.go
```diff
@@ -56,19 +61,63 @@ func newCreateValCmd() *cobra.Command {
...
+
 type createValConfig struct {
-	Network            netconf.ID
-	PrivateKeyFile     string
-	ConsensusPubKeyHex string
-	SelfDelegation     uint64
+	eoaConfig
+	consensusPubKeyHex string
+	selfDelegation     uint64
 }
```
corverroos, line 113: I know most people say that ALL fields of private types must also be private. But I like to design and communiate the API of a type, even if it is a private type. I therefore use exported fields to indicate that these fields can be read/written by any function. And that they are not private to the type. Private fields should never be accessed by another outside the type. They should be accessed via getters/setters. 

Also, simple structs with exported fields indicate simple structs. I'd suggest leaving the fields as public to make that explicit.

cli/cmd/staking.go
```diff
@@ -187,12 +222,112 @@ func createValidator(ctx context.Context, cfg createValConfig) error {
...
+		return err
+	}
+
+	bal, err := eth.EtherBalanceAt(ctx, delegatorAddr)
+	if err != nil {
+		return err
+	} else if bal <= float64(cfg.amount) {
+		return &CliError{
+			Msg:     fmt.Sprintf("Delegator address has insufficient balance=%.2f OMNI, address=%s", bal, delegatorAddr),
+			Suggest: "Fund the operator address with sufficient OMNI for self-delegation and gas",
```
corverroos, line 305: Fund the delegator address

corverroos (approved):

PR #2179, opened 2024-10-15
DRAFT

chmllr:
Body goes here...

issue: none or #12345


halo/healthcheck.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+response=$(wget -S -O /dev/null http://localhost:26660 2>&1 | awk '/^  HTTP/{print $2}' | tail -1)
```
corverroos, line 3: for consistency, we define bash variables in `UPPERCASE`

halo/healthcheck.sh
```diff
@@ -0,0 +1,9 @@
...
+
+if [ "$response" -ge 300 ]; then
```
corverroos, line 5: nit: lets be thorough, use `"${FOO}"` syntax

halo/healthcheck.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
corverroos, line 1: not sure `scratch` has `sh` installed?

chmllr: Yep it definitely doesn't, but I am not even there yet: I'm struggling to even copy the scripts into the right contianer.

halo/cmd/ready.go
```diff
@@ -0,0 +1,61 @@
...
+			return nil
+		},
+	}
+
+	return cmd
+}
+
+// assertReady calls halo's /ready endpoint and returns nil if the status is ready
+// or an error otherwise.
+func assertReady(ctx context.Context) error {
```
corverroos, line 37: follow `status` command example, use `readyConfig` struct with a field: `MonitoringAddr` which defaults to `cmtconfig.DefaultInstrumentationConfig().PrometheusListenAddr`.

halo/cmd/ready.go
```diff
@@ -0,0 +1,61 @@
...
+	cfg := cmtcfg.DefaultConfig()
+	url := fmt.Sprintf("http://0.0.0.0%v/ready", cfg.Instrumentation.PrometheusListenAddr)
+
```
corverroos, line 40: suggest rather:
```
host, port, err := net.SplitHostPort(readyCfg.MonitoringAddr)
if err ...
else if host == "" {
  host = "locahost"
}

addr := net.JoinHostPort(host, port)
```

halo/cmd/ready.go
```diff
@@ -0,0 +1,61 @@
...
+		return errors.Wrap(err, "http request creation")
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return errors.Wrap(err, "http request")
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode < 400 {
```
corverroos, line 55: rather use http standard `if resp.StatusCode / 100 != 2 `

corverroos: also use "golden path" pattern, which final return being the happy path.

halo/cmd/ready.go
```diff
@@ -0,0 +1,61 @@
...
+	return cmd
+}
+
+// assertReady calls halo's /ready endpoint and returns nil if the status is ready
+// or an error otherwise.
+func assertReady(ctx context.Context) error {
+	cfg := cmtcfg.DefaultConfig()
+	url := fmt.Sprintf("http://0.0.0.0%v/ready", cfg.Instrumentation.PrometheusListenAddr)
+
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
```
corverroos, line 41: why add this timeout, callers can add their own timeout of they want

halo/cmd/ready.go
```diff
@@ -0,0 +1,61 @@
...
+
+	cmtcfg "github.com/cometbft/cometbft/config"
+
+	"github.com/spf13/cobra"
+)
+
+func newReadyCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "ready",
+		Short: "Assert the readiness of the halo node",
```
corverroos, line 20: follow `status` command description

halo/cmd/ready.go
```diff
@@ -0,0 +1,61 @@
...
+		return errors.Wrap(err, "http request")
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode < 400 {
+		log.Info(ctx, "The node is ready")
+		return nil
+	}
+
+	return errors.New("the node is not ready yet")
```
corverroos, line 60: rather: `node not ready`

halo/cmd/ready.go
corverroos: can also just simply this by setting default `monitoringAddr="http://localhost:26660"`

halo/cmd/ready.go
chmllr: SGTM, no strong feelings 👍

halo/cmd/ready.go
chmllr: Yes, but I'd still take the port from `cfg.Instrumentation.PrometheusListenAddr`

corverroos: It is a tradeoff of readability vs correctness. I'd stick with simple string literal. We can always add a test that ensures the string literal matches `cfg.Instrumentation.PrometheusListenAddr`

chmllr: Sounds good!

halo/Dockerfile
```diff
@@ -8,6 +8,8 @@ COPY --from=alpine:latest /tmp /tmp
 
 COPY halo /app
 
+HEALTHCHECK --interval=30s --timeout=5s --start-period=10s CMD ["/app", "ready"]
```
corverroos, line 11: I'd just stick to default values, why do we need to change timeout or start period?

halo/app/monitor.go
```diff
@@ -160,6 +160,25 @@ func monitorEVMOnce(ctx context.Context, ethCl ethclient.Client, status *readine
 	return nil
 }
 
+// exportReadiness exports the node readiness to prometheus.
+func exportReadiness(ctx context.Context, status *readinessStatus) {
```
corverroos, line 164: nit: `instrumentReadiness`

halo/cmd/ready.go
```diff
@@ -0,0 +1,64 @@
...
+		MonitoringAddr: "http://localhost:26660",
+	}
+}
+
+func newReadyCmd() *cobra.Command {
+	cfg := defaultReadyConfig()
+
+	cmd := &cobra.Command{
+		Use:   "ready",
+		Short: "Query node for readiness",
```
corverroos, line 28: `Query remote node for readiness`

halo/cmd/ready.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func defaultReadyConfig() readyConfig {
+	return readyConfig{
+		MonitoringAddr: "http://localhost:26660",
+	}
+}
+
+func newReadyCmd() *cobra.Command {
+	cfg := defaultReadyConfig()
```
corverroos, line 24: follow `status` command example, bind a monitoring address flag.

halo/cmd/ready.go
```diff
@@ -0,0 +1,64 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "ready",
+		Short: "Query node for readiness",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			err := queryReady(cmd.Context(), cfg)
+			if err != nil {
+				return errors.Wrap(err, "ready failed")
```
corverroos, line 33: see go style guide, no need to specify "failed" when wrapping

halo/app/monitor.go
```diff
@@ -160,6 +160,25 @@ func monitorEVMOnce(ctx context.Context, ethCl ethclient.Client, status *readine
...
+	ticker := time.NewTicker(time.Second * 10)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			health := 0.0
```
corverroos, line 173: rather initialise zero variables by type: `var ready float64`

halo/app/metrics.go
```diff
@@ -47,6 +47,13 @@ var (
 		Name:      "size_bytes",
 		Help:      "Current size of the database directory in bytes.",
 	})
+
+	nodeReadiness = promauto.NewGauge(prometheus.GaugeOpts{
+		Namespace: "halo",
+		Subsystem: "health",
+		Name:      "ready",
+		Help:      "Node readiness",
```
corverroos, line 55: see `cometValidator` above for how to comment "constant gauges"

corverroos (approved):

halo/cmd/ready.go
chmllr: But is it "remote" if run the command on my halo node that queries localhost?

halo/Dockerfile
chmllr: True, it helped me debugging it, but the default ones seem to be good enough for production, epscially since this is not really used anywhere yet.

halo/cmd/ready.go
chmllr: Discussed in Slack

PR #2216, opened 2024-10-18
feat(lib/cchain): support grpc cprovider

### LINKED ISSUE #2002: Add gRPC CProvider impl
#### Problem to Solve
CProvider currently uses the CometBFT RPC to query CosmosSDK application state (attest/portal/valsync modules). 

CometBFT however implements a sequential concurrency model. So queries and block processing compete for time resources. See https://github.com/cosmos/cosmos-sdk/issues/10859.

Queries are therefore slower, and high query load can negatively affect block building latency. See #1865 .
 
Local simnet testing showed that gRPC is 2x~3x faster than CometBFT.

#### Proposed Solution
- Add a second gRPC based CProvider implementation. 
- Note we'll need to remove the CometClient method (only used by xconnector). 
- We don't expose gRPC API publicly on staging or omega yet, so all external use still need to use the CometBFT impl.
- First usage of gRPC could be validators querying themselves.
- Second could be relayer querying archive nodes (ensure firewall is open)
- Third, think about and discuss exposing the gRPC API publicly (and create subsequent ticket).

corverroos:
Implement a gRPC cprovider that connects directly to CosmosSDK gprc server bypassing cometBFT. This reduces negative impact of CometBFT ABCI queries on chain performance and speeds up queries.

issue: fixes #2002


lib/cchain/provider/abci.go
```diff
@@ -709,7 +732,7 @@ func queryLatestAttestation(ctx context.Context, cl atypes.QueryClient, chainVer
 			ChainId:   chainVer.ID,
 			ConfLevel: uint32(chainVer.ConfLevel),
 		})
-	if errors.Is(err, sdkerrors.ErrKeyNotFound) {
+	if errors.Is(err, sdkerrors.ErrKeyNotFound) || status.Code(err) == codes.NotFound {
 		return xchain.Attestation{}, false, nil
 	} else if err != nil {
 		return xchain.Attestation{}, false, err
```
sideninja, line 738: the function bellow: `attsFromAtHeight` should that also check for `status.Code(err) == codes.NotFound` and return nil instead of an error?

sideninja (approved):

lib/cchain/provider/abci.go
corverroos: No, I don't think so, since the function returns an empty list of nothing is found.

kc1116 (approved):

kevinhalliday (approved):

PR #2225, opened 2024-10-18
fix(halo/app): fix ready command

chmllr:
Fixes the `ready` command that is using the wrong URL for readiness check.

issue: none

### DIFF
--- a/halo/cmd/ready.go
+++ b/halo/cmd/ready.go
@@ -45,7 +45,8 @@ func newReadyCmd() *cobra.Command {
 // queryReady calls halo's /ready endpoint and returns nil if the status is ready
 // or an error otherwise.
 func queryReady(ctx context.Context, cfg readyConfig) error {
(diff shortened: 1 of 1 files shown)


halo/cmd/ready.go
```diff
@@ -16,7 +16,7 @@ type readyConfig struct {
 
 func defaultReadyConfig() readyConfig {
 	return readyConfig{
-		MonitoringURL: "http://localhost:26660",
+		MonitoringURL: "http://localhost:26660/ready",
```
corverroos, line 19: I'd argue that `MonitoringURL` should only be the base host and port of the monitoring API to query. The `ready` command should add the path it wants to query to in that base URL

corverroos (approved):

halo/cmd/ready.go
chmllr: Agreed, I had the same first intuition, but then remembered the "remote node" discussion and was not sure anymore. Will make it a non-configurable part of the URL.

Issue #2207, opened 2024-10-17
Some metrics are not exported on staging anymore


chmllr:
#### Problem to Solve

Some metrics are not exported on staging anymore. This is likely related to https://github.com/omni-network/omni/pull/2035 which disables the CometBFT metrics.

#### Proposed Solution

Find a way how to export the metrics again.

chmllr:
Closing as it was addressed in https://github.com/omni-network/omni/pull/2232

PR #2232, opened 2024-10-20
fix(halo): fix missing comet prometheus metrics

corverroos:
Fix missing cometBFT prometheus metrics

issue:  #2207

### DIFF
--- a/halo/app/monitor.go
+++ b/halo/app/monitor.go
@@ -15,6 +15,7 @@ import (
 	cmtcfg "github.com/cometbft/cometbft/config"
 	rpcclient "github.com/cometbft/cometbft/rpc/client"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
@@ -64,12 +65,12 @@ func monitorCometForever(
 // monitorCometOnce monitors the cometBFT peers, and sync status.
 func monitorCometOnce(ctx context.Context, rpcClient rpcclient.Client, isSyncing func() bool, lastHeight int64, status *readinessStatus) (int64, error) {
 	netInfo, err := rpcClient.NetInfo(ctx)
-	status.setConsensusP2PPeers(netInfo.NPeers)
 	if err != nil {
 		return 0, errors.Wrap(err, "net info")
 	} else if netInfo.NPeers == 0 {
 		log.Error(ctx, "Halo has 0 consensus p2p peers", nil)
 	}
+	status.setConsensusP2PPeers(netInfo.NPeers)
 
 	synced := !isSyncing()
 	setConstantGauge(cometSynced, synced)
@@ -78,9 +79,16 @@ func monitorCometOnce(ctx context.Context, rpcClient rpcclient.Client, isSyncing
 	abciInfo, err := rpcClient.ABCIInfo(ctx)
 	if err != nil {
 		return 0, errors.Wrap(err, "abci info")
-	} else if !isSyncing() && lastHeight > 0 && abciInfo.Response.LastBlockHeight <= lastHeight {
-		log.Warn(ctx, "Halo height is not increasing, evm syncing?", nil, "height", abciInfo.Response.LastBlockHeight)
+	} else if lastHeight > 0 && abciInfo.Response.LastBlockHeight <= lastHeight {
+		if !isSyncing() {
+			log.Warn(ctx, "Halo consensus height is not increasing", nil, "height", abciInfo.Response.LastBlockHeight)
+		}
+		status.setConsensusRunning(false)
+	} else {
+		status.setConsensusRunning(true)
 	}
+
+	status.setConsensusHeight(abciInfo.Response.LastBlockHeight)
 
 	return abciInfo.Response.LastBlockHeight, nil
 }
@@ -156,6 +164,7 @@ func monitorEVMOnce(ctx context.Context, ethCl ethclient.Client, status *readine
 		return errors.Wrap(err, "block number")
 	}
 	evmHeight.Set(float64(latest))
+	status.setExecutionHeight(latest)
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


halo/app/monitor.go
```diff
@@ -202,30 +211,32 @@ func dirSize(path string) (int64, error) {
...
+) func(context.Context) error {
 	mux := http.NewServeMux()
 
-	// We align both with industry standards (endpoint /metrics)
-	// and with cometBFT (endpoint /).
-	mux.Handle("/metrics", promhttp.Handler())
-	mux.Handle("/", promhttp.Handler())
+	promHandler := promhttp.HandlerFor(
+		prometheus.DefaultGatherer,
+		promhttp.HandlerOpts{MaxRequestsInFlight: cfg.Instrumentation.MaxOpenConnections},
```
corverroos, line 226: wire `cfg.Instrumentation.MaxOpenConnections` which was also not used anymore

halo/app/monitor.go
```diff
@@ -202,30 +211,32 @@ func dirSize(path string) (int64, error) {
...
+	)
+	mux.Handle("/metrics", promHandler) // Industry standard `/metrics` path
+	mux.Handle("/", promHandler)        // CometBFT standard `/` path
 
-	// On the `/ready` endpoint, we serve the readiness of the halo node.
-	// Additionally, in case when the node is not ready, the response status code is set to 503.
+	// Serve readiness status json at `/ready`, returning 503 if not ready.
```
corverroos, line 231: Hey, I adjusted the comment to make it more concise and direct. I also removed "we" to keep it neutral and focused on the action.

halo/app/monitor.go
```diff
@@ -202,30 +211,32 @@ func dirSize(path string) (int64, error) {
...
 	mux.HandleFunc("/ready", func(w http.ResponseWriter, _ *http.Request) {
 		w.Header().Set("Content-Type", "application/json")
 
-		if !status.ready() {
-			w.WriteHeader(http.StatusServiceUnavailable)
-		}
-		if err := status.serialize(w); err != nil {
```
corverroos, line 226: there is actualy a race condition here. We read ready and then we serialiaze. Technically, the status could change in between these two steps resulting in inconsistent output. This isn't critical code, but I thought I would clean it up in any case

halo/app/monitor.go
```diff
@@ -236,8 +247,8 @@ func startMonitoringAPI(
 	}
 
 	go func() {
-		if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
```
corverroos, line 239: see `server.ListenAndServe` docs, it always returns a non-nil error, so no need to check for nil

halo/app/ready.go
```diff
@@ -13,39 +13,40 @@ type readinessStatus struct {
 	mu                 sync.RWMutex
 	ConsensusSynced    bool    `json:"consensus_synced"`              // Unhealthy if false
 	ConsensusP2PPeers  int     `json:"consensus_p2p_peers"`           // Unhealthy if 0
+	ConsensusHeight    int64   `json:"consensus_height"`              // Unhealthy if 0
+	ConsensusRunning   bool    `json:"consensus_running"`             // Unhealthy if false
 	ExecutionConnected bool    `json:"execution_connected"`           // Unhealthy if false
 	ExecutionSynced    bool    `json:"execution_synced"`              // Unhealthy if false
-	ExecutionP2PPeers  *uint64 `json:"execution_p2p_peers,omitempty"` // Unhealthy if 0
+	ExecutionP2PPeers  *uint64 `json:"execution_p2p_peers,omitempty"` // Unhealthy if 0 (ignored if not available)
+	ExecutionHeight    uint64  `json:"execution_height"`              // Unhealthy if 0
```
corverroos, line 21: added Heights, just as a nice to have to users querying the endpoint and looking at the json.

halo/app/ready.go
```diff
@@ -13,39 +13,40 @@ type readinessStatus struct {
 	mu                 sync.RWMutex
 	ConsensusSynced    bool    `json:"consensus_synced"`              // Unhealthy if false
 	ConsensusP2PPeers  int     `json:"consensus_p2p_peers"`           // Unhealthy if 0
+	ConsensusHeight    int64   `json:"consensus_height"`              // Unhealthy if 0
+	ConsensusRunning   bool    `json:"consensus_running"`             // Unhealthy if false
```
corverroos, line 17: aded `consensus_running` so the app becomes unhealthy if consensus stalls like on staging. The nodes were healthy even though they stalled

halo/app/start.go
```diff
@@ -314,7 +314,12 @@ func newCometNode(ctx context.Context, cfg *cmtcfg.Config, app *App, privVal cmt
 		},
 	)
 
-	// Don't instantiate the default prometheus server.
+	// Configure CometBFT prometheus metrics as per provided config
+	metrics := node.DefaultMetricsProvider(&cmtcfg.InstrumentationConfig{
+		Prometheus: cfg.Instrumentation.Prometheus,
+		Namespace:  cfg.Instrumentation.Namespace,
```
corverroos, line 320: I think is that you were missing in your PR. That resulted in incorrect metric names so it seemed like it wasn't renabled.

chmllr: Yes that must be it because I saw there is no prefix `cometbft_` 👍

Thanks a lot Corver!

halo/app/ready.go
```diff
@@ -13,39 +13,40 @@ type readinessStatus struct {
...
 	r.mu.RLock()
 	defer r.mu.RUnlock()
 
 	if err := json.NewEncoder(w).Encode(r); err != nil {
-		return errors.Wrap(err, "readiness status serialization")
+		return false, errors.Wrap(err, "readiness status serialization")
 	}
 
-	return nil
+	return r.ready(), nil
```
chmllr, line 49: Wouldn't this cause a dead lock? `ready()` will try to lock while we're holding the lock?

chmllr: Ah looks like read locks are reentrant in Go by default. This explains why no test is failing.

chmllr (approved):

PR #524, opened 2024-03-06
chore(contracts): add goerli avs deployment script

kevinhalliday:
Add goerli avs deployment script.

- add proxy admin deployment script
- add goerli avs deployment script
- add utils for deploying & testing again a local goerli fork
- update tests to handle stateful fork tests

task: https://app.asana.com/0/1206208509925075/1206648863973026

PR #2240, opened 2024-10-21
fix(e2e): update mainnet funding thresholds

sideninja:
Update thresholds for funding.

issue: #524


e2e/app/eoa/fund.go
```diff
@@ -39,7 +39,7 @@ var (
 	}
 
 	defaultThresholdsByRole = map[Role]FundThresholds{
-		RoleRelayer:         thresholdLarge,  // Relayer needs a ton of balance.
+		RoleRelayer:         thresholdMedium, // Relayer needs a ton of balance.
```
corverroos, line 42: Update comment

halo/genutil/evm/evm.go
```diff
@@ -185,13 +184,13 @@ func omegaPrefundAlloc() types.GenesisAlloc {
...
 func mainnetPrefundAllocs() types.GenesisAlloc {
 	return types.GenesisAlloc{
-		eoa.MustAddress(netconf.Mainnet, eoa.RoleCreate3Deployer): {Balance: eth0_1}, // deploys one contract
-		eoa.MustAddress(netconf.Mainnet, eoa.RoleDeployer):        {Balance: eth0_1}, // deploys a couple contracts
-		eoa.MustAddress(netconf.Mainnet, eoa.RoleManager):         {Balance: eth10},  // rarely used
-		eoa.MustAddress(netconf.Mainnet, eoa.RoleUpgrader):        {Balance: eth10},  // rarely used
-		eoa.MustAddress(netconf.Mainnet, eoa.RoleRelayer):         {Balance: eth100}, // enough to relay for a while
-		eoa.MustAddress(netconf.Mainnet, eoa.RoleMonitor):         {Balance: eth100}, // enough to monitor for a while
-		eoa.MustAddress(netconf.Mainnet, eoa.RoleFunder):          {Balance: eth500}, // enough for a few funding actions
+		eoa.MustAddress(netconf.Mainnet, eoa.RoleCreate3Deployer): {Balance: eth0_01}, // deploys one contract
```
corverroos, line 187: don't hardcode this, pull it from eoa package.

corverroos (approved):

halo/genutil/evm/evm.go
corverroos: can add a `eoa.MustTargetBalance`

e2e/app/eoa/fund.go
```diff
@@ -39,7 +39,7 @@ var (
 	}
 
 	defaultThresholdsByRole = map[Role]FundThresholds{
-		RoleRelayer:         thresholdLarge,  // Relayer needs a ton of balance.
+		RoleRelayer:         thresholdMedium, // Relayer needs sufficient balance to operator for 2 weeks
```
ttarsi, line 42:
```suggestion
		RoleRelayer:         thresholdMedium, // Relayer needs sufficient balance to operate for 2 weeks
```

kevinhalliday (changes requested):
Some things I think we need to consider

- we don't want to fund `RoleTester` in mainnet genesis. which is set to `ZeroXDead` address.
- Right now we have same thresholds for OMNI and ETH chains. Which I don't think is sufficient. For ex a target balance of 2 OMNI on Omn's EVM is not sufficient to run for 2 weeks. (for ex in Omege we spend 2 OMNI in a couple hours). We haven't really had to deal with our thresholds being too low on Omni because we prefund all of our addresses with a lot, so we've never hit those thresholds. But now is the time

To solve:

- let's add we have different thresholds for OMNI and ETH. we can scale up ETH thresholds by ~20x
- make sure we do not fund RoleTester on mainnet. let's add guiderails to error if we try to

halo/genutil/evm/evm.go
```diff
@@ -184,15 +181,26 @@ func omegaPrefundAlloc() types.GenesisAlloc {
...
+	fundRoles := []eoa.Role{
+		eoa.RoleRelayer, eoa.RoleMonitor,
+		eoa.RoleCreate3Deployer, eoa.RoleManager,
+		eoa.RoleUpgrader, eoa.RoleDeployer,
+		eoa.RoleFunder,
 	}
+
+	for _, role := range fundRoles {
+		fund, ok := eoa.GetFundThresholds(tokens.OMNI, netconf.Mainnet, role)
+		if !ok {
```
kevinhalliday, line 195: This will skip the `eoa.Funder` role, because we do not set thresholds for it. Either choose a sensible value, or derive it from the target balances of other roles

e2e/app/eoa/fund.go
```diff
@@ -81,10 +109,22 @@ func (t FundThresholds) MinBalance() *big.Int {
...
 }
+
+func convert(threshold FundThresholds, token tokens.Token) (FundThresholds, error) {
+	conversion, ok := tokenConversion[token]
+	if !ok {
+		return FundThresholds{}, errors.New("fund conversion", "token", token.String())
+	}
+
+	return FundThresholds{
+		minETH:    threshold.minETH * conversion,
```
sideninja, line 127: I don't think this is nicest thing to do because we are overwritting ETH with Omni, but the alternative is to implement Balance that can either be OMNI and ETH and then have conversions between those, which for this point where we only have ETH and OMNI was a bit overkill, I didn't want to generalise it to just `minBalance` because it's not clear then in which currency we define it above in the funding thresholds

contracts/allocs/mainnet.json
```diff
@@ -1347,7 +1347,7 @@
     }
   },
   "0x121e240000000000000000000000000000000002": {
```
sideninja, line 1349: @kevinhalliday this is likely just the fact we sum two floats 0.01 + 0.01 != 0.02 so we maybe should change from floats

kevinhalliday: yep that's what i did

sideninja: shouldn't the above value be updated then?

Issue #1399, opened 2024-07-08
test(contracts): test individual preinstall functions


kevinhalliday:
*Problem To Solve*

We predeploy a number of "preinstall" contracts in genesis (see [Preinstalls.sol](https://github.com/omni-network/omni/blob/main/)). These are all well known contracts deployed on many chains. We insert their bytecode directly, but currently have no tests on their functionality.

*Proposed Solution*

We should have some "smoke" tests that each preinstall works. Basically, try to call a couple functions on it and make sure it does what it should and doesn't revert.

PR #2243, opened 2024-10-22
test(contracts/core): added basic smoke tests for preinstalls

Zodomo:
I added a handful of basic functionality tests purely intended to ensure each preinstall is functioning and that their bytecode can execute. Most of the tests involve view function call checks, however some also handle writes as well. At the time of PR, `BeaconBlockRoots` is the only preinstall remaining that is untested. 

issue: #1399


Zodomo:
@kevinhalliday I applied your and @ttarsi's feedback regarding comments, unifying code patterns, and cleaning up the test contract itself. Please give it another round of review when able.

contracts/core/foundry.toml
```diff
@@ -26,6 +26,7 @@ ffi = true
 
 # TODO: upgrade solc
 solc_version = "0.8.24"
+evm_version = "shanghai"
```
kevinhalliday, line 29: Perhaps a comment explaining why we added this

contracts/core/test/octane/Preinstalls.t.sol
```diff
@@ -2,18 +2,55 @@
...
 
 import { Preinstalls } from "src/octane/Preinstalls.sol";
-import { Test, Vm } from "forge-std/Test.sol";
-
-interface IEIP712 {
-    function DOMAIN_SEPARATOR() external view returns (bytes32);
-}
+import "./PreinstallInterfaces.sol";
+import { AllocPredeploys } from "script/genesis/AllocPredeploys.s.sol";
+import { Test, Vm, console2 } from "forge-std/Test.sol";
```
kevinhalliday, line 7: think we can remove console2

we have linting for unused imports but we ignore lints on tests files because was noisy. wish solhint gave us more config options , rather than just ignoring directories entirely. you have any other sol linters you like?

kevinhalliday (approved):
Nice work. Just restore pnpm lock changes before merge

contracts/core/foundry.toml
Zodomo: We originally added this because we needed to set the evm_version in order to interact with the BeaconBlockRoots contract locally. However, once applied, we found that the bytecode would change as the EVM was upgraded from the default version Paris, which is likely due to PUSH0 being activated in solc 0.8.24. After some deliberation, we decided to remove the BeaconBlockRoots test as we did not want the bytecode to change, and eliminated the evm_version config as well. External testing of this contract on Omega has shown it works, so we will leave it at that instead of committing to new bytecode.

contracts/core/test/octane/Preinstalls.t.sol
Zodomo: Removed console2. Unfortunately I have not used any solidity linters other than `forge fmt`. I know there's solidhunter from Osmium, but its just a faster (not completely featured) solhint.

PR #2277, opened 2024-10-24
feat(relayer): cache attestation retrievals

chmllr:
WIP for https://www.notion.so/omni-network/Dedup-relayer-streams-12769a99643a80c7b78cea02dbea35ba

issue: none

### DIFF
--- a/lib/cchain/provider/cache.go
+++ b/lib/cchain/provider/cache.go
@@ -1,6 +1,7 @@
 package provider
 
 import (
+	"math"
 	"sync"
 
 	"github.com/omni-network/omni/lib/xchain"
@@ -61,20 +62,22 @@ func (ac *attestationCache) update(chainVer xchain.ChainVersion, attestastions [
 
 	// If no cache exist yet, initialize it.
 	if _, present := ac.caches[chainVer]; !present {
-		ac.caches[chainVer] = &chainVersionCache{map[uint64]xchain.Attestation{}, 0}
+		ac.caches[chainVer] = &chainVersionCache{map[uint64]xchain.Attestation{}, math.MaxUint64}
 	}
 
 	chainVerCache := ac.caches[chainVer]
 
 	// Insert each attestation above the watermark into the cache.
-	// Note, it could overwrite previous values, but this is idempotent.
 	for _, att := range attestastions {
-		if att.BlockHeader.BlockHeight < chainVerCache.watermark {
+		attHeight := att.AttestOffset
+		if attHeight < chainVerCache.watermark {
 			continue
 		}
-		attHeight := att.BlockHeader.BlockHeight
+		if _, exists := chainVerCache.cache[attHeight]; exists {
+			break
+		}
 		chainVerCache.cache[attHeight] = att
-		chainVerCache.watermark = max(chainVerCache.watermark, attHeight)
+		chainVerCache.watermark = min(chainVerCache.watermark, attHeight)
 	}
 
 	// Raise the watermark until we hit the configured cache size and delete all
--- a/lib/cchain/provider/cache.go
+++ b/lib/cchain/provider/cache.go
@@ -1,0 +1,86 @@
+package provider
+
+import (
+	"sync"
+
+	"github.com/omni-network/omni/lib/xchain"
+)
+
+// chainVersionCache contains approved attestations of a chain version with all heights above the watermark.
+type chainVersionCache struct {
+	cache     map[uint64]xchain.Attestation
+	watermark uint64
+}
+
+// attestationCache if a thread-safe collection of caches containing attestations per chain version.
+// All cache sizes will be kept under `cacheSize` elements.
+type attestationCache struct {
+	cacheSizes int
+	caches     map[xchain.ChainVersion]*chainVersionCache
+	mu         sync.RWMutex
+}
+
+// newAttestationCache constructs a default cache collection with size 100 per cache.
+func newAttestationCache() *attestationCache {
+	return &attestationCache{
+		cacheSizes: 100,
+		caches:     make(map[xchain.ChainVersion]*chainVersionCache),
+	}
+}
+
+// get returns all cached consequent attestations, starting from the specified height.
+func (ac *attestationCache) get(chainVer xchain.ChainVersion, height uint64) []xchain.Attestation {
+	var attestations []xchain.Attestation
+
+	ac.mu.RLock()
+	defer ac.mu.RUnlock()
+	chainVerCache := ac.caches[chainVer]
+
+	if chainVerCache == nil || height < chainVerCache.watermark {
+		return attestations
+	}
+
+	for {
+		value, exists := chainVerCache.cache[height]
+		if !exists {
+			break
+		}
+		attestations = append(attestations, value)
+		height++
+	}
+
+	return attestations
+}
+
+// update extends the cache of chain version with all new attestations that are above the stored watermark, and then
+// removes the oldest ones to keep the total cache size at `cachesSize`. Note we do not cache attestations below the
+// watermark because they are only expected to be occasionally fetched by retrying workers.
+func (ac *attestationCache) update(chainVer xchain.ChainVersion, attestastions []xchain.Attestation) {
+	ac.mu.Lock()
+	defer ac.mu.Unlock()
+
+	// If no cache exist yet, initialize it.
+	if _, present := ac.caches[chainVer]; !present {
+		ac.caches[chainVer] = &chainVersionCache{map[uint64]xchain.Attestation{}, 0}
+	}
+
+	chainVerCache := ac.caches[chainVer]
+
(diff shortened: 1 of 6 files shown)


lib/cchain/provider/provider.go
```diff
@@ -181,7 +183,22 @@ func (p Provider) stream(
...
 		FetchBatch: func(ctx context.Context, _ uint64, offset uint64) ([]xchain.Attestation, error) {
-			return p.fetch(ctx, chainVer, offset)
+			// If cached attestations found, return them.
+			cachedAtts := p.cache.get(chainVer, offset)
+			if len(cachedAtts) > 0 {
+				return cachedAtts, nil
+			}
+
+			// If the height is not cached, fetch the attestations.
+			atts, err := p.fetch(ctx, chainVer, offset)
```
corverroos, line 193: here you use attestOffset for lookup, but you are indexing by block offset

lib/cchain/provider/provider.go
```diff
@@ -181,7 +183,22 @@ func (p Provider) stream(
...
+			if err != nil {
+				return []xchain.Attestation{}, err
```
corverroos, line 195: nil is valid slice, so rather `return nil, errro`

lib/cchain/provider/cache.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// Insert each attestation above the watermark into the cache.
+	// Note, it could overwrite previous values, but this is idempotent.
+	for _, att := range attestastions {
+		if att.BlockHeader.BlockHeight < chainVerCache.watermark {
+			continue
+		}
+		attHeight := att.BlockHeader.BlockHeight
+		chainVerCache.cache[attHeight] = att
+		chainVerCache.watermark = max(chainVerCache.watermark, attHeight)
```
corverroos, line 77: shouldn't it be `min`? watermark is lowest element right?

lib/cchain/provider/cache.go
```diff
@@ -0,0 +1,86 @@
...
+type chainVersionCache struct {
+	cache     map[uint64]xchain.Attestation
+	watermark uint64
+}
+
+// attestationCache if a thread-safe collection of caches containing attestations per chain version.
+// All cache sizes will be kept under `cacheSize` elements.
+type attestationCache struct {
+	cacheSizes int
+	caches     map[xchain.ChainVersion]*chainVersionCache
```
corverroos, line 19: i'd personally use a slice for performance reasons, but maybe that is overkill (and adds complexity), might be worth doing a benchmark teststo compare slices vs maps

corverroos, line 19: I'd just using `sync.Map` here and then call `sync.LoadOrStore` below, this avoids having a single lock on all caches, then the lock can moved inside each cache, slightly more performant

lib/cchain/provider/provider.go
```diff
@@ -72,6 +72,7 @@ type Provider struct {
 	backoffFunc func(context.Context) func()
 	chainNamer  func(xchain.ChainVersion) string
 	network     netconf.ID
+	cache       *attestationCache
```
corverroos, line 75: as discussed in the meet yesterday, what about putting this inside the `stream` package rather?

lib/cchain/provider/cache.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"sync"
+
+	"github.com/omni-network/omni/lib/xchain"
+)
+
+// chainVersionCache contains approved attestations of a chain version with all heights above the watermark.
+type chainVersionCache struct {
+	cache     map[uint64]xchain.Attestation
```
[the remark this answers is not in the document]
chmllr: I even thought about it, but decided against to avoid looking up the insertion point, plus handling of slice mutation (I remember slices have a lot of gotchas)

lib/cchain/provider/cache.go
chmllr: Oh great, not sure I knew about `sync.Map` 👍

lib/cchain/provider/cache.go
chmllr: Yes, this is a bug.

lib/cchain/provider/provider.go
chmllr: I definitely have no strong feelings yet, so I'll move there.

lib/cchain/provider/cache.go
chmllr: I remember now why I need the global lock. The `caches` itself is mutable and shared too. Unless we can pre-initialize upfront for all known chain versions.

lib/cchain/provider/provider.go
chmllr: Interesting observation: since `Deps` is generic (supports `any E`), in order to wrap the generic `FetchBatch` with a internal cached version, my cache needs to work for `any E` too, yet I have a requirement for some simple interface that can give me the key of the value. Then we could implement this interface for Attestations to return the `AttestOffset` as key and do the same for whatever else is used with `Stream[E any]`.

This is a bigger change and I can definitely do it, but wanted to make you aware that the footprint of the change will be much larger.

lib/stream/stream.go
```diff
@@ -20,7 +21,7 @@ type Deps[E any] struct {
 
 	// FetchBatch fetches the next batch of elements from the provided height (inclusive).
 	// The elements must be sequential, since the internal height cursors is incremented for each element returned.
-	FetchBatch func(ctx context.Context, chainID uint64, height uint64) ([]E, error)
+	FetchBatch func(ctx context.Context, chainVer xchain.ChainVersion, height uint64) ([]E, error)
```
corverroos, line 24: creating a dependency between streams and xchain isn't good

lib/stream/cache.go
```diff
@@ -0,0 +1,89 @@
...
+	if _, present := ac.caches[chainVer]; !present {
+		ac.caches[chainVer] = &chainVersionCache{map[uint64]xchain.Attestation{}, math.MaxUint64}
+	}
+
+	chainVerCache := ac.caches[chainVer]
+
+	// Insert each attestation above the watermark into the cache.
+	for _, att := range attestastions {
+		attHeight := att.AttestOffset
+		if attHeight < chainVerCache.watermark {
```
corverroos, line 73: one should be able to add elements in the front of the cache. it should only be skipped if the cache is full

lib/stream/cache.go
```diff
@@ -0,0 +1,89 @@
...
+			continue
+		}
+		if _, exists := chainVerCache.cache[attHeight]; exists {
+			break
+		}
+		chainVerCache.cache[attHeight] = att
+		chainVerCache.watermark = min(chainVerCache.watermark, attHeight)
```
corverroos, line 80: this will upgrade watermark is being added. I thought that watermark is lowest height in the cache. Maybe improve variable name, as watermark isn't obvious what it means.

lib/stream/cache.go
```diff
@@ -0,0 +1,89 @@
...
+	cache     map[uint64]xchain.Attestation
+	watermark uint64
+}
+
+// attestationCache if a thread-safe collection of caches containing attestations per chain version.
+// All cache sizes will be kept under `cacheSize` elements.
+type attestationCache struct {
+	cacheSizes int
+	caches     map[xchain.ChainVersion]*chainVersionCache
+	mu         sync.RWMutex
```
corverroos, line 21: a single lock around all caches isn't good

lib/stream/cache.go
```diff
@@ -0,0 +1,89 @@
...
+// get returns all cached consequent attestations, starting from the specified height.
+func (ac *attestationCache) get(chainVer xchain.ChainVersion, height uint64) []xchain.Attestation {
+	var attestations []xchain.Attestation
+
+	ac.mu.RLock()
+	defer ac.mu.RUnlock()
+	chainVerCache := ac.caches[chainVer]
+
+	if chainVerCache == nil || height < chainVerCache.watermark {
+		return attestations
```
corverroos, line 41: rather return nil

lib/stream/cache.go
corverroos: nit: a watermark is usually higher than the water, it is set during floods, in this case it is the lower thing in the cache, so not sure about naming it watermark

lib/stream/cache.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+		chainVerCache.cache[attHeight] = att
+		chainVerCache.watermark = min(chainVerCache.watermark, attHeight)
+	}
+
+	// Raise the watermark until we hit the configured cache size and delete all
+	// attestations below the watermark.
+	for len(chainVerCache.cache) > ac.cacheSizes {
+		delete(chainVerCache.cache, chainVerCache.watermark)
+		chainVerCache.watermark++
```
corverroos, line 87: the new value of watermark may be invalid if there are gaps in the cache

PR #2281, opened 2024-10-25
fix(ci): increase the nightly-join test timeout

chmllr:
By default, a Go test times out after 10m. But the syncing is expected to take much longer, so we disable the time out because the test has a built in timeout of 6h.

issue: none

### DIFF
--- a/.github/workflows/ci-join.yml
+++ b/.github/workflows/ci-join.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Run join test
         run: |
           cd scripts/join
-          go test . -v --integration --logs_file=docker_logs.txt
+          go test . -v --integration --logs_file=docker_logs.txt -timeout 8h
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-join.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Run join test
         run: |
           cd scripts/join
-          go test . -v --integration --logs_file=docker_logs.txt
+          go test . -v --integration --logs_file=docker_logs.txt -timeout 8h
```
corverroos, line 34: I'd just disable timeout since the test handles timeouts itself, so ``--timeout=0`

corverroos (approved):

.github/workflows/ci-join.yml
chmllr: Ah thanks, yeah this is even better.

PR #2283, opened 2024-10-25
fix(halo/app): do not override the http status header

chmllr:
We're currently setting the http status code twice when the server is unhealthy.

issue: none

### DIFF
--- a/halo/app/monitor.go
+++ b/halo/app/monitor.go
@@ -231,24 +231,19 @@ func startMonitoringAPI(
 
 	// Serve readiness status json at `/ready`, returning 503 if not ready.
 	mux.HandleFunc("/ready", func(w http.ResponseWriter, _ *http.Request) {
-		w.Header().Set("Content-Type", "application/json")
-
-		// We do not serialize `status` directly to the http writer, because even if
-		// it can be serialized without errors, we might need to set a different http status
-		// code than the one set by default on a successful write.
-		var buf bytes.Buffer
-		if ready, err := status.serialize(&buf); err != nil {
+		var body bytes.Buffer
+		statusCode := http.StatusOK
+		if ready, err := status.serialize(&body); err != nil {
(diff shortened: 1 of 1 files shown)


halo/app/monitor.go
```diff
@@ -232,11 +233,21 @@ func startMonitoringAPI(
 	mux.HandleFunc("/ready", func(w http.ResponseWriter, _ *http.Request) {
 		w.Header().Set("Content-Type", "application/json")
 
-		if ready, err := status.serialize(w); err != nil {
+		// We do not serialize `status` directly to the http writer, because even if
```
corverroos, line 236: this is good, but another option is to group all writes for clarity:
```
// Serve readiness status json at `/ready`, returning 503 if not ready.
	mux.HandleFunc("/ready", func(w http.ResponseWriter, _ *http.Request) {
		var body bytes.Buffer
		statusCode := http.StatusOK
		if ready, err := status.serialize(&body); err != nil {
			http.Error(w, err.Error(), http.StatusInternalServerError)
			return
		} else if !ready {
			statusCode = http.StatusServiceUnavailable
		}

		// Do writes in correct order
		w.Header().Set("Content-Type", "application/json")
		w.WriteHeader(statusCode)
		_, _ = w.Write(body.Bytes())
	})
```

corverroos (approved):

halo/app/monitor.go
```diff
@@ -232,11 +233,21 @@ func startMonitoringAPI(
...
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		} else if !ready {
 			w.WriteHeader(http.StatusServiceUnavailable)
+			return
+		}
+
+		_, err := w.Write(buf.Bytes())
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
```
corverroos, line 250: we've already called `w.Write` by this point, so not sure doing more writes here will actually work, can just skip this I think,

PR #2285, opened 2024-10-25
chore(docs): upload sigma prime audit final version

kc1116:
Upload sigma prime audit final version.

issue: none


ttarsi (commented):
2 comments
- let's just replace the `sigma-prime-chain-2.pdf` since this is just the finalized version of that report. It's referenced in some public links (i.e. protocol overview for cantina) so replacing directly means we don't have to go and update any references to it.
- maybe we can wait until OM2-12 is resolved, otherwise we'll need to replace it again

ttarsi (approved):

Issue #2154, opened 2024-10-11
Update xsubmission tests to use XSubGen


kevinhalliday:
#### Problem to Solve

Our `OmniPortal.xsubmit` tests use a [ts script](https://github.com/omni-network/omni/tree/a782d51ad534f59ffaa20201f5711ee7ecb47e79/contracts/core/test/xchain/ts/script/genxsubs) to generate xsubmission merkle proofs. We did this so that we could use the [oz's merkle tree js lib](https://github.com/OpenZeppelin/merkle-tree). The current flow is

- define xblocks (by name) in [fixtures](https://github.com/omni-network/omni/blob/a782d51ad534f59ffaa20201f5711ee7ecb47e79/contracts/core/test/xchain/common/Fixtures.sol)
- output xblocks to json with forge script
- generate xsub roots and multi proofs with ts, write them to json
- read json xblocks & xsubs in forge tests

This is clunky. We recently added a new [XSubGen](https://github.com/omni-network/omni/blob/main/contracts/core/test/utils/XSubGen.sol) util contract that uses sigma primes sol [implementation of oz's merkle tree js](https://github.com/sigp/multiproof). This let's use genereate roots and multiproofs synchronously in forge tests.

#### Proposed Solution

Rewrite exisiting tests to use XSubGen

PR #2287, opened 2024-10-25
test(contracts/core): replaced TS helpers with XSubGen

Zodomo:
Replaced TS helpers with XSubGen.sol for XBlock and XSub handling directly via foundry tests

issue: #2154


contracts/core/test/utils/XSubGen.sol
```diff
@@ -59,18 +62,20 @@ contract XSubGen is Test {
...
     function makeXSub(
         uint64 valSetId,
+        uint64 destChainId,
         XTypes.BlockHeader memory xheader,
         XTypes.Msg[] memory msgs,
         bool[] memory msgFlags
     ) public view returns (XTypes.Submission memory) {
         require(msgs.length == msgFlags.length, "msg length must match msgFlags length");
 
+        (msgs, msgFlags) = _filterXMsgs(msgs, msgFlags, destChainId);
```
kevinhalliday, line 78: Now we're introducing two ways of selecting a subset of `msgs` - `msgFlags`and `destChainId`.

I'd prefer to keep surface are of this function sig simpler. If a caller wants filter by destChain, he can use msgFlags. But he may not always want to do that (ex, testing that a submission with mutiple dest chains revert).

Also, by filtering the `msgs` (rather than just msgFlags) we are creating a submission with an attestation root built only from xmsgs with a that dest. The attestation root will change when selecting new dest chains - this is not reflective of how attestation roots are built in the wild. 

I suggest refactoring so something like this

```solidity
makeXSub(valSetId, xheader, msgs, msgFlagsForDest(msgs, destChainId))
```

kevinhalliday (commented):
Great work - one comment worth addressing

contracts/core/test/utils/XSubGen.sol
Zodomo: I've refactored to use this flow instead, and adjusted all relevant tests to use it where needed.

kevinhalliday (approved):
👏

PR #2288, opened 2024-10-25
feat(monitor/app): track and compare quicknode RPC's last height with omni node

chmllr:
This PR sets up monitoring of the public quicknode RPCs and compares the reporeted highest height with what we get from an omni node.

issue: https://github.com/omni-network/ops/issues/494


monitor/app/quicknode.go
```diff
@@ -0,0 +1,70 @@
...
+	"time"
+
+	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/ethclient"
+	"github.com/omni-network/omni/lib/log"
+	"github.com/omni-network/omni/lib/netconf"
+)
+
+const (
+	quicknodePublicRPC = "https://damp-wandering-gadget.omni-omega.quiknode.pro"
```
corverroos, line 14: actually, lets rather make this a flag/config item. Then we can specify different values in ops repo (per network) and not leak this to public

corverroos (approved):

sideninja (approved):

PR #2290, opened 2024-10-25
POC stream cache

sideninja:
POC for https://www.notion.so/omni-network/Dedup-relayer-streams-12769a99643a80c7b78cea02dbea35ba

My take doesn't change the workers it just minimise the overhead of having more workers. I believe having more workers with sufficient caching this PR introduces is a benefit because a blast radius in case something goes wrong or is held up of current workers setup is as minimal as it can get. Beside minimising fetching same data this also minimises the cost of doing a bisections when searching for offests. However this will still be done on a cold cache.

This approach also makes the changes minimal and well isolated to a seperate component which can and should be unit tested in the next step. The added complexity is small, but impact is big, because this should also be used with xprovider and would have even bigger impact (TODO).

issue: none

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -294,6 +294,8 @@ require (
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
+require github.com/maypok86/otter v1.2.3
+
 require (
 	buf.build/gen/go/bufbuild/bufplugin/protocolbuffers/go v1.34.2-20240928190436-5e8abcfd7a7e.2 // indirect
 	buf.build/gen/go/pluginrpc/pluginrpc/protocolbuffers/go v1.34.2-20240828222655-5345c0a56177.2 // indirect
@@ -309,6 +311,8 @@ require (
 	github.com/containerd/platforms v0.2.1 // indirect
 	github.com/containerd/ttrpc v1.2.5 // indirect
 	github.com/containerd/typeurl/v2 v2.2.0 // indirect
+	github.com/dolthub/maphash v0.1.0 // indirect
+	github.com/gammazero/deque v0.2.1 // indirect
 	github.com/go-task/slim-sprig/v3 v3.0.0 // indirect
 	github.com/moby/locker v1.0.1 // indirect
 	github.com/moby/patternmatcher v0.6.0 // indirect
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -294,6 +294,8 @@ require (
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
+require github.com/maypok86/otter v1.2.3
```
sideninja, line 297: this should be tided, weird

lib/stream/cache.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/maypok86/otter"
+)
+
+// Cacher defines the cache type for caching stream items.
+// Stream uses height as stream cursor so our cache implementation
+// relies on the height for calculating the key.
+//
+// The key implementation can vary based on the context and is up
+// to the implementor.
+type Cacher[K any, E any] interface {
```
corverroos, line 18: K not required

lib/stream/cache.go
```diff
@@ -0,0 +1,93 @@
+package stream
+
+import (
+	"strconv"
+	"time"
+
+	"github.com/omni-network/omni/lib/errors"
+
+	"github.com/maypok86/otter"
```
corverroos, line 9: not sure about using a external KV lib that requires string based keys.

lib/stream/cache.go
```diff
@@ -0,0 +1,93 @@
...
+
+	return cache, nil
+}
+
+func (c *Cache[E]) Get(height uint64) ([]E, bool) {
+	// todo report hit/miss metrics
+	return c.cache.Get(c.key(height))
+}
+
+func (c *Cache[E]) Set(height uint64, value []E) bool {
```
corverroos, line 91: This isn't correct. Each element has a height, the height here is just the height of the first element, they should be indexed separately. Your get will only return values if there was a set for that exact number, it will return nothing even f the cache contains that height otherwise

lib/cchain/provider/provider.go
```diff
@@ -179,6 +180,17 @@ func (p Provider) stream(
 	srcChain := p.chainNamer(chainVer)
 	ctx := log.WithCtx(in, "src_chain", srcChain, "worker", workerName)
 
+	// note: see the note from cache, this is poc
+	key := func(height uint64) string {
```
corverroos, line 184: this cache isn't actually shared between stream, so will result in 0 cache hits

lib/stream/cache.go
sideninja: Sure, a for loop is needed to itterate over the items inside. That was an overlook

PR #2292, opened 2024-10-27
feat(cchain/provider): add attest stream cache

corverroos:
Add support for cache in `stream` package. Implement a map based cache and wire it as part of cprovider.

issue: #2321

### DIFF
--- a/lib/stream/cache.go
+++ b/lib/stream/cache.go
@@ -1,0 +1,82 @@
+package stream
+
+import (
+	"math"
+	"sync"
+)
+
+// NewNopCache returns a cache that does nothing.
+func NewNopCache[E any]() Cache[E] {
+	return new(nopCache[E])
+}
+
+type nopCache[E any] struct{}
+
+func (nopCache[E]) Get(uint64) []E  { return nil }
+func (nopCache[E]) Set(uint64, []E) {}
+
+// NewCache returns a new map based cache using the provided limit.
+func NewCache[E any](limit int) Cache[E] {
+	return &mapCache[E]{
+		limit: limit,
+		elems: make(map[uint64]E),
+		floor: math.MaxUint64,
+	}
+}
+
+// mapCache implements Cache storing elements in a map.
+// It will only retain the highest `limit` elements.
+// It assumes elements are roughly sequential and that heights are not sparse.
+type mapCache[E any] struct {
+	limit int // Immutable
+
+	// Mutable fields
+	mu    sync.RWMutex
+	elems map[uint64]E
+	floor uint64 // floor is equaled to or lower than the lowest height in elems. It is MaxUint64 if elems is empty.
+}
+
+// Get returns all strictly sequential elements in the cache from the provided height (inclusive).
+func (c *mapCache[E]) Get(from uint64) []E {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	var resp []E
+	for height := from; ; height++ {
+		elem, ok := c.elems[height]
+		if !ok {
+			break
+		}
+
+		resp = append(resp, elem)
+	}
+
+	return resp
+}
+
+// Set adds the provided elements to the cache starting from the provided height (inclusive).
+// Elements are assumed to be in strictly sequential order.
+func (c *mapCache[E]) Set(from uint64, elems []E) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	for i, elem := range elems {
+		height := from + uint64(i)
+
+		if len(c.elems) >= c.limit && height < c.floor {
+			// Don't add too low elements when cache is full, they will just be deleted below.
+			continue
+		}
+
+		c.elems[height] = elem
+
+		if height < c.floor {
+			c.floor = height // Update floor
+		}
+	}
+
+	for len(c.elems) > c.limit {
+		delete(c.elems, c.floor)
+		c.floor++
+	}
+}
--- a/lib/stream/cache_test.go
+++ b/lib/stream/cache_test.go
@@ -1,0 +1,165 @@
+package stream_test
+
+import (
+	"math/rand/v2"
+	"sync"
+	"testing"
+
+	"github.com/omni-network/omni/lib/stream"
+
+	"github.com/stretchr/testify/require"
+)
+
+//go:generate go test . -run=TestFloor -count=100
+
+func TestFloor(t *testing.T) {
+	t.Parallel()
+
+	const total = 100
+	toAdd := make(map[int]struct{})
+	for i := range total {
+		toAdd[i] = struct{}{}
+	}
+
+	cache := stream.NewCache[int](total)
+	// Add in random order
+	for i := range toAdd {
+		cache.Set(uint64(i), []int{i})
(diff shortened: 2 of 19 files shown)


lib/stream/cache.go
```diff
@@ -0,0 +1,172 @@
...
+	return new(nopCache[E])
+}
+
+type nopCache[E any] struct{}
+
+func (nopCache[E]) Get(uint64) []E  { return nil }
+func (nopCache[E]) Set(uint64, []E) {}
+
+//
+// func NewSliceCache[E any](limit int) Cache[E] {
```
corverroos, line 17: I implemented a slice based cache, but went for map in the end since it is much simpler and performance is acceptable.

corverroos: I'll remove this before finally merging the PR, it is just here for reference

lib/stream/cache.go
```diff
@@ -0,0 +1,196 @@
...
+// mapCache implements Cache storing elements in a map.
+// It will only retain the highest `limit` elements.
+// It assumes elements are roughly sequential and that heights are not sparse.
+type mapCache[E any] struct {
+	limit int // Immutable
+
+	// Mutable fields
+	mu     sync.RWMutex
+	elems  map[uint64]E
+	lowest uint64 // Invalid while elems is empty
```
corverroos, line 137: other option is using a 3rd party ordered map like [this](https://github.com/elliotchance/orderedmap)

kc1116 (changes requested):
1. Why does the cache need to maintain _strictly sequential order_ ? This generic cache doesn't need to know about that functionality since the caller already knows all attestation keys (they are sequential) the caller can request them sequentially. This will simplify this cache and make it completely reusable. 

2. We can use this popular [LRU cache implementation](https://github.com/hashicorp/golang-lru) by hashicorp. It's LRU and has a TTL mechanism which should be all we need. A light wrapper around this that exposes a basic CRUD API should suffice.

lib/stream/cache.go
chmllr: Are there any other places in the code base where we manage the key ordering of a map? If yes, this could be another good issue for beginner to refactor multiple places using one new dependency.

lib/stream/cache.go
```diff
@@ -0,0 +1,198 @@
...
+	for len(c.elems) > c.limit {
+		delete(c.elems, c.lowest)
+
+		// Find the new lowest
+		for {
+			c.lowest++
+			if _, ok := c.elems[c.lowest]; ok {
+				break
+			}
+		}
```
chmllr, line 196: Since you rely on the `lowest` to point to some element in the map eventually anyway, why do we need the second loop advancing to the next existing element? I think `delete` for a non-existing key is a noop and will do all the same checks you're doing manually right now.

lib/stream/cache.go
```diff
@@ -0,0 +1,198 @@
...
+		height := from + uint64(i)
+
+		if len(c.elems) >= c.limit && height < c.lowest {
+			// Don't add too low elements when cache is full, they will just be deleted below.
+			continue
+		}
+
+		if len(c.elems) == 0 {
+			c.lowest = height // Initialize lowest when first element added.
+		}
```
chmllr, line 178: Why not initializing `lowest` with max uint64 instead and remove this?

lib/stream/cache_test.go
```diff
@@ -0,0 +1,145 @@
...
+	var wg sync.WaitGroup
+	added := make(chan []int)
+
+	// Ensure it is empty
+	requireGetLen(t, 0, 0)
+	requireGetLen(t, 1, 0)
+	requireGetLen(t, limit, 0)
+
+	// Concurrently add elements to the cache
+	const writers = 10
```
chmllr, line 39: Should we add one slacking writer, so that we have a gap in the range of values?

lib/stream/cache.go
corverroos: Not that I know of

lib/stream/cache.go
corverroos: Sure, we can do that: `c.lowest` is the lower height in the cache or `math.Maxuint64` while the cache is empty.

lib/stream/cache.go
corverroos: it is about correctness for me. 

My definition: c.lowest is the lowest height in the cache or math.Maxuint64 while the cache is empty.

Are you suggesting: c.lower is lower or equaled to lowest elements in the cache...?

lib/stream/cache_test.go
corverroos: we do test gaps below?

lib/stream/cache.go
chmllr: > c.lower is lower or equaled to lowest elements in the cache...?

Yes! This is how I read the code:

```
while having too many elements
  a. remove lowest
  b. find the next lowest by incrementing the pointer to something that is present
  c. repeat
```

So it feels like the step b. can be simply replaced by an `increment the pointer`.

Because both versions will work if the invariant holds: `c.lowest` is below `limit` elements still present in the cache.

But no strong feels, I'm just saying when reading the code and explaining correctness to myself I "optimize" this loop "away" because I can see how it adds extra asserts for anything.

lib/stream/cache.go
chmllr: Then I'd vote for not using a dependency.

lib/stream/cache.go
```diff
@@ -0,0 +1,198 @@
...
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	var resp []E
+	for height := from; ; height++ {
+		elem, ok := c.elems[height]
+		if !ok {
+			break
+		} else if height < c.lowest {
+			return nil, errors.New("height < lowest [BUG]") // Sanity check
```
sideninja, line 153: why is this a bug? if a cache keeps a limit and the items exceeding the limit gets ejected, then if a relayers want to restart from begining it will request items that have height lower than the lowest height of the current cache state.

lib/stream/cache.go
```diff
@@ -0,0 +1,198 @@
...
+	limit int // Immutable
+
+	// Mutable fields
+	mu     sync.RWMutex
+	elems  map[uint64]E
+	lowest uint64 // Invalid while elems is empty
+}
+
+// Get returns all strictly sequential elements in the cache from the provided height (inclusive).
+func (c *mapCache[E]) Get(from uint64) ([]E, error) {
```
sideninja, line 143: wouldn't this be a problem because it will contain all chain versions (final, latest etc), but the consumer expects only the chain version it wants to get streamed

lib/stream/cache.go
corverroos: the contract of `c.lowest` is: "the lower element in the cache". So if you find something lower, then that invariant doesn't hold right?

lib/stream/cache.go
corverroos: each cache instance is only used by a single chain version

chmllr (approved):

sideninja (approved):

PR #2301, opened 2024-10-28
fix(contracts/core): remove redundant xcall fee check

kevinhalliday:
address(this).balance is inclusive of msg.value. We only need to check the former

Add usage comment.

issue: none


Zodomo (approved):
Agreed. It is on the integrator to determine how the payment is made. As long as the contract has the funds, the payment can technically succeed. Enforcing a msg.value check also prevents a contract from subsidizing the payment for the user, so removing this redundant check gives apps more flexibility in that regard.

PR #2308, opened 2024-10-28
ci(lib/contracts): retry factory deploys

corverroos:
Add retries to all create3 factory deploy.

issue: none


ttarsi (approved):

contracts/bindings/create3_helper.go
```diff
@@ -0,0 +1,32 @@
...
+		if err == nil {
+			return tx, nil
+		} else if i >= retry {
+			return nil, errors.Wrap(err, "factory deploy", "attempt", i)
+		}
+
+		ctx := opts.Context
+		if ctx == nil {
+			ctx = context.Background()
+		}
```
chmllr, line 28: Does it make sense to move out of the loop?

sideninja (approved):

sideninja:
nit: since we started doing retries on different deploy operations we could consider extracting this into a lib function

PR #2316, opened 2024-10-29
refactor(lib/ethclient): make syncprogress safe

corverroos:
Replace `ethclient.SyncProgress` which is prone to panics with a safe version `ethclient.ProgressIfSyncing` which returns `pointer, bool, error`.

issue: none

### DIFF
--- a/lib/ethclient/ethclient.go
+++ b/lib/ethclient/ethclient.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/omni-network/omni/lib/errors"
 
+	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/ethclient"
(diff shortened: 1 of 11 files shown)


sideninja (approved):

lib/ethclient/ethclient.go
```diff
@@ -92,6 +93,25 @@ func (w Wrapper) Address() string {
 	return w.address
 }
 
+// ProgressIfSyncing returns the sync progress (and true) if the node is syncing else false if node is
+// not syncing or an error.
+//
+// This wrap SyncProgress which returns nil-nil if the node is not syncing which results in panics.
+func (w Wrapper) ProgressIfSyncing(ctx context.Context) (*ethereum.SyncProgress, bool, error) {
```
chmllr, line 100: Would it be simpler to assume that every client is _always_ syncing with the only question is how much it is behind. In this case we could always return the sync progress struct, but for nodes that are not behind, the `CurrentBlock == HighestBlock` would always hold, which could be quickly checked with something like `Done()` (or a new function `Synced` meaning we're on the tip, because we are never really "done") 

(In fact this is exactly how I initally understood this API and overlooked the pointer)

chmllr (approved):

lib/ethclient/ethclient.go
corverroos: I feel that would change the API too much, I'd rather stick with standard API, just make it safer.

PR #2317, opened 2024-10-29
fix(monitor/app): public rpc tracking

chmllr:
Fixes the wrong usage of the eth client API and makes it use the known RPC for each network or uses one of the omni nodes otherwise.

issue: none


monitor/app/public_rpc.go
```diff
@@ -52,18 +47,34 @@ func monitorPublicRPCForever(
...
+		netconf.Staging: "staging",
+		netconf.Omega:   "quicknode.omega", // TODO: delete quicknode subdomain
+		netconf.Mainnet: "mainnet",
+	}
+
+	if subdomain, exists := subdomains[network.ID]; exists {
+		return ethclient.Dial(chain.Name, fmt.Sprintf("https://%v.omni.network", subdomain))
+	}
+
+	return ethClients[chain.ID], nil
```
corverroos, line 63: also check this isn't nil

monitor/app/public_rpc.go
```diff
@@ -52,18 +47,34 @@ func monitorPublicRPCForever(
 	}
 }
 
+// publicRPCEndpoint returns the public RPC endpoint for the network and chain specified.
+// If no public RPC is known, return a node of the chain directly.
+func publicRPCEndpoint(network netconf.Network, chain netconf.Chain, ethClients map[uint64]ethclient.Client) (ethclient.Client, error) {
+	subdomains := map[netconf.ID]string{
+		netconf.Staging: "staging",
```
corverroos, line 54: might as well use the full url here, easier to read.

corverroos (approved):

Issue #2314, opened 2024-10-29
join mainnet nightly


corverroos:
#### Problem to Solve
We need a CI test to ensure that joining mainnet works and stays performant.

Omega join nightly is failing since it takes longer than max github runner duration of 6 hours.

#### Proposed Solution
TBD

PR #2318, opened 2024-10-29
ci(github): add mainnet join nightly

chmllr:
- Disable omega nightly join until the sync timeout is understood
- Enable mainnet nightly sync (syncs in 10m on a Macbook)
- Allow to run the action manually both for test and mainnet

issue: #2314

### DIFF
--- a/scripts/join/join_test.go
+++ b/scripts/join/join_test.go
@@ -30,14 +30,15 @@ import (
 
 var (
 	logsFile    = flag.String("logs_file", "join_test.log", "File to write docker logs to")
+	network     = flag.String("network", "omega", "Network to join (default: omega)")
 	integration = flag.Bool("integration", false, "Run integration tests")
 )
 
-// TestJoinOmega starts a local node (using omni operator init-nodes)
+// TestJoinNetwork starts a local node (using omni operator init-nodes)
 // and waits for it to sync.
 //
 //nolint:paralleltest // Parallel tests not supported since we start docker containers.
-func TestJoinOmega(t *testing.T) {
+func TestJoinNetwork(t *testing.T) {
 	if !*integration {
 		t.Skip("skipping integration test")
 	}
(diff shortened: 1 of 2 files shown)


scripts/join/join_test.go
```diff
@@ -30,14 +30,15 @@ import (
 
 var (
 	logsFile    = flag.String("logs_file", "join_test.log", "File to write docker logs to")
+	network     = flag.String("network", "omega", "Network to join (default: omega)")
```
corverroos, line 33: see `netconf.BindFlag` how to this with proper types.

.github/workflows/ci-join-omega.yml
```diff
@@ -1,10 +1,11 @@
 name: join omega nightly
 # Nightly action that tests joining omega as a full node.
 
-on:
-  workflow_dispatch:
-  schedule:
-    - cron: "0 1 * * 1-5" # Weekdays at 1am UTC
+# We temporarily disable this action because it fails and needs to be investigated
+# on:
+#   workflow_dispatch:
```
corverroos, line 6: maybe retain manual triggering

.github/workflows/ci-join-mainnet.yml
```diff
@@ -0,0 +1,42 @@
+name: join mainnet nightly
+# Nightly action that tests joining mainnet as a full node.
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 1 * * 1-5" # Weekdays at 1am UTC
+
+permissions:
```
corverroos, line 9: instead of duplicating, can't we reuse the same job by just using variables?

scripts/join/join_test.go
chmllr: What I saw there is that I'd need to provide a place holder variable to write a value into. But in this test the params need to be declared before `TestJoinNetwork` is called. So I guess I could declare a `var network netconf.ID` and then maybe bind it from the `init` function (I need something running before the test itself), but I remember `init` is discouraged by the linter, so even if I mask it with annotations or use some pre-fixture, I am not sure why this added complexity is better than just reading a string?

corverroos: ok yeah, fair enough, that pattern doesn't really work for these simple global flags, fine to go with this impl

corverroos (approved):

PR #2320, opened 2024-10-29
feat(realyer): cache offset

sideninja:
Body goes here...

issue: none or #12345

### DIFF
--- a/e2e/app/rpc.go
+++ b/e2e/app/rpc.go
@@ -84,7 +84,7 @@ func waitForHeight(ctx context.Context, testnet *e2e.Testnet, height int64) (*ty
 			if len(clients) == 0 {
 				return nil, nil, errors.New("unable to connect to any network nodes")
 			}
-			if time.Since(lastIncrease) >= 20*time.Second {
+			if time.Since(lastIncrease) >= 180*time.Second {
 				if maxResult == nil {
 					return nil, nil, errors.New("chain stalled at unknown height")
 				}
--- a/e2e/manifests/devnet1.toml
+++ b/e2e/manifests/devnet1.toml
@@ -4,6 +4,7 @@ anvil_chains = ["mock_l1", "mock_l2"]
 
 multi_omni_evms = true
 prometheus = true
+pingpong_n = 30 # Increased ping pong to span validator updates
 
 [node.validator01]
 [node.validator02]
--- a/lib/cchain/provider/abci.go
+++ b/lib/cchain/provider/abci.go
@@ -13,7 +13,6 @@ import (
 	"github.com/omni-network/omni/lib/cchain"
 	"github.com/omni-network/omni/lib/errors"
 	"github.com/omni-network/omni/lib/expbackoff"
-	"github.com/omni-network/omni/lib/log"
 	"github.com/omni-network/omni/lib/netconf"
 	"github.com/omni-network/omni/lib/tracer"
 	"github.com/omni-network/omni/lib/umath"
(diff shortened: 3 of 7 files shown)


e2e/app/rpc.go
```diff
@@ -84,7 +84,7 @@ func waitForHeight(ctx context.Context, testnet *e2e.Testnet, height int64) (*ty
 			if len(clients) == 0 {
 				return nil, nil, errors.New("unable to connect to any network nodes")
 			}
-			if time.Since(lastIncrease) >= 20*time.Second {
+			if time.Since(lastIncrease) >= 180*time.Second {
```
sideninja, line 87: reverted after docker with arch amd is stalling otherwise

e2e/manifests/devnet1.toml
```diff
@@ -4,6 +4,7 @@ anvil_chains = ["mock_l1", "mock_l2"]
 
 multi_omni_evms = true
 prometheus = true
+pingpong_n = 30 # Increased ping pong to span validator updates
```
sideninja, line 7: temp test

lib/cchain/provider/abci.go
```diff
@@ -344,54 +343,11 @@ func newABCIAllAttsFunc(cl atypes.QueryClient) allAttsFunc {
 }
 
 func newABCIFetchFunc(attCl atypes.QueryClient, cmtCl cmtservice.ServiceClient, chainNamer func(xchain.ChainVersion) string) fetchFunc {
-	return func(ctx context.Context, chainVer xchain.ChainVersion, fromOffset uint64) ([]xchain.Attestation, error) {
```
sideninja, line 347: moved to attestations type

lib/cchain/provider/attestations.go
```diff
@@ -0,0 +1,350 @@
...
+	"github.com/omni-network/omni/lib/umath"
+	"github.com/omni-network/omni/lib/xchain"
+
+	"github.com/cosmos/cosmos-sdk/client/grpc/cmtservice"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type attestations struct {
```
corverroos, line 19: not a fan of this OO style of coding, why can't we use simple functions?

corverroos: also the type name doesn't really make sense?

lib/cchain/provider/attestations.go
```diff
@@ -0,0 +1,350 @@
...
+	var startHeight, endHeight uint64
+	// if we are performing search for the first time and no previous height was cached do a full search
+	if cachedHeight == 0 {
+		startHeight, endHeight, err = a.lookbackRange(ctx, chainVer, fromOffset, latestHeight)
+		if err != nil {
+			incQueryErr(endpoint)
+			return nil, 0, errors.Wrap(err, "lookback search")
+		}
+	} else {
+		startHeight = cachedHeight + 1 // we move to the next height
```
corverroos, line 82: not sure we can skip cached height, it might still contain the attestation we are looking for, best start from this point (inclusive)

lib/cchain/provider/attestations.go
```diff
@@ -0,0 +1,350 @@
...
+		endHeight = latestHeight
+
+		// try first doing a simple forward search on next height before binary search
```
corverroos, line 85: not sure this is a good optimisation, what is the probability of the next block containing the attestation, I'd say it is very low

sideninja: yeah I'm not sure, I agree we can just use binary search, I was thinking of doing some forward search but don't have a good argument

lib/cchain/provider/attestations.go
sideninja: I guess is a bit of a preference, I just thought there was a lot of passing of clients around the functions, I can go back to functions and test those directly

Issue #1860, opened 2024-09-12
Cache historical height when streaming attestations


corverroos:
#### Problem to Solve
When using cprovider to stream attestations, we do a full historical search when fetching each subsequent batch of attestations. 

But subsequent attestations will never be at an earlier height, so if the previous fetched height is known, the `lookback` step can be skipped.

This should decrease the `historical baseline` metric.

#### Proposed Solution
Cache the height used to fetch the next back of attestations when streaming.
Use this as lookback result when fetching the next batch

sideninja:
Related thread https://omni-chain.slack.com/archives/C02HCJN6KNE/p1729174654091949

PR #2325, opened 2024-10-29
feat(realyer): cache height for offset search

sideninja:
Cache the height found during attestation offset lookback search so it can be reused on the next offset search.

issue: #1860

### DIFF
--- a/lib/cchain/provider/abci.go
+++ b/lib/cchain/provider/abci.go
@@ -13,7 +13,6 @@ import (
 	"github.com/omni-network/omni/lib/cchain"
 	"github.com/omni-network/omni/lib/errors"
 	"github.com/omni-network/omni/lib/expbackoff"
-	"github.com/omni-network/omni/lib/log"
 	"github.com/omni-network/omni/lib/netconf"
 	"github.com/omni-network/omni/lib/tracer"
 	"github.com/omni-network/omni/lib/umath"
@@ -352,7 +351,7 @@ func newABCIAllAttsFunc(cl atypes.QueryClient) allAttsFunc {
 }
 
 func newABCIFetchFunc(attCl atypes.QueryClient, cmtCl cmtservice.ServiceClient, chainNamer func(xchain.ChainVersion) string) fetchFunc {
-	return func(ctx context.Context, chainVer xchain.ChainVersion, fromOffset uint64) ([]xchain.Attestation, error) {
+	return func(ctx context.Context, chainVer xchain.ChainVersion, fromOffset uint64, cursor uint64) ([]xchain.Attestation, uint64, error) {
 		const endpoint = "fetch_attestations"
 		defer latency(endpoint)()
 
@@ -365,40 +364,60 @@ func newABCIFetchFunc(attCl atypes.QueryClient, cmtCl cmtservice.ServiceClient, 
 		atts, ok, err := attsFromAtHeight(ctx, attCl, chainVer, fromOffset, 0)
 		if err != nil {
 			incQueryErr(endpoint)
-			return nil, errors.Wrap(err, "abci query attestations-from")
+			return nil, 0, errors.Wrap(err, "abci query attestations-from")
 		} else if ok {
-			fetchStepsMetrics(chainName, 0, 0)
-			return atts, nil
+			binarySearchStepsMetric(chainName, 0)
+			lookbackStepsMetric(chainName, 0)
+
+			return atts, cursor, nil
 		}
 
 		earliestAttestationAtLatestHeight, ok, err := queryEarliestAttestation(ctx, attCl, chainVer, 0)
 		if err != nil {
 			incQueryErr(endpoint)
-			return nil, errors.Wrap(err, "abci query earliest-attestation-in-state")
+			return nil, 0, errors.Wrap(err, "abci query earliest-attestation-in-state")
 		}
 
 		// Either no attestations have happened yet, or the queried fromOffset is in the "future"
 		// Caller has to wait and retry in both cases
 		if !ok || earliestAttestationAtLatestHeight.AttestOffset < fromOffset {
-			// First attestation hasn't happened yet, return empty
-			return []xchain.Attestation{}, nil
+			// First attestation hasn't happened yet, return empty and set cursor to latest height
+			return []xchain.Attestation{}, earliestAttestationAtLatestHeight.BlockHeight, nil
 		}
 
-		offsetHeight, err := searchOffsetInHistory(ctx, cmtCl, attCl, chainVer, chainName, fromOffset)
+		latestBlockResp, err := cmtCl.GetLatestBlock(ctx, &cmtservice.GetLatestBlockRequest{})
+		if err != nil {
+			return []xchain.Attestation{}, 0, errors.Wrap(err, "query latest block")
+		}
+
+		latestHeight := uint64(latestBlockResp.SdkBlock.Header.Height)
+
+		// Binary search range from cached to latest
+		searchStart, searchEnd := cursor, latestHeight
+		if cursor == 0 {
+			// Unless no cached height provided, then perform lookback search
+			searchStart, searchEnd, err = lookbackRange(ctx, attCl, chainVer, chainName, fromOffset, latestHeight)
+			if err != nil {
+				incQueryErr(endpoint)
+				return nil, 0, errors.Wrap(err, "lookback search")
+			}
+		}
+
+		offsetHeight, err := binarySearch(ctx, attCl, chainVer, chainName, fromOffset, searchStart, searchEnd)
 		if err != nil {
 			incQueryErr(endpoint)
-			return nil, errors.Wrap(err, "searching offset in history")
+			return nil, 0, errors.Wrap(err, "binary search")
 		}
 
 		atts, ok, err = attsFromAtHeight(ctx, attCl, chainVer, fromOffset, offsetHeight)
 		if err != nil {
 			incQueryErr(endpoint)
-			return nil, errors.Wrap(err, "abci query attestations-from")
+			return nil, 0, errors.Wrap(err, "abci query attestations-from")
 		} else if !ok {
-			return nil, errors.New("expected to find attestations [BUG]")
+			return nil, 0, errors.New("expected to find attestations [BUG]")
 		}
 
-		return atts, nil
+		return atts, offsetHeight, nil
 	}
 }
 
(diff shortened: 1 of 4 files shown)


lib/cchain/provider/abci.go
```diff
@@ -763,3 +671,134 @@ func attsFromAtHeight(ctx context.Context, cl atypes.QueryClient, chainVer xchai
...
+
+	return 0, errors.New("unexpectedly reach end of search method [BUG]")
+}
+
+// lookbackRange does an exponential lookback from the latest block height provided
+// until attestation found has lower offset than the provided from offset, this
+// guarantees the fromOffset attestation will be found in the range.
+//
+//nolint:nonamedreturns // named returned for clarity
+func lookbackRange(
```
sideninja, line 735: lookback was extracted from search history function but it is unchanged (beside steps count part of for loop and metric report)

lib/cchain/provider/abci.go
```diff
@@ -763,3 +671,134 @@ func attsFromAtHeight(ctx context.Context, cl atypes.QueryClient, chainVer xchai
 
 	return atts, true, nil
 }
+
+// binarySearch uses a binary search between defined start and end consensus block heights to
+// find the attestations with the provided fromOffset.
+func binarySearch(
```
sideninja, line 677: binary search was extracted from search history function but it is unchanged (beside steps count part of for loop and metric report)

lib/cchain/provider/abci.go
```diff
@@ -562,124 +588,6 @@ func spanName(endpoint string) string {
 	return "cprovider/" + endpoint
 }
 
-// searchOffsetInHistory searches the consensus state history and
-// returns a historical consensus block height that contains an approved attestation
-// for the provided chain version and fromOffset.
-func searchOffsetInHistory(ctx context.Context, cmtCl cmtservice.ServiceClient, attCl atypes.QueryClient, chainVer xchain.ChainVersion, chainName string, fromOffset uint64) (uint64, error) {
```
sideninja, line 568: exploded into two functions lookback and binary search bellow

lib/cchain/provider/abci.go
```diff
@@ -357,40 +356,67 @@ func newABCIFetchFunc(attCl atypes.QueryClient, cmtCl cmtservice.ServiceClient,
...
+			startHeight, endHeight, err = lookbackRange(ctx, attCl, chainVer, chainName, fromOffset, latestHeight)
+			if err != nil {
+				incQueryErr(endpoint)
+
+				return nil, 0, errors.Wrap(err, "lookback search")
+			}
+		} else {
+			// otherwise use cached height from previous search as the next start height
+			// but increased by one since we are searching the next attestation offset
+			startHeight = cachedHeight + 1
```
corverroos, line 399: as mentioned before, there is no guarantee that this offset isn't the same height as previous fetch. You can't assume this. Remove the +1.

lib/cchain/provider/abci.go
corverroos: please improve comment, focus on the what values are returned what what they mean.

lib/cchain/provider/abci.go
```diff
@@ -763,3 +671,134 @@ func attsFromAtHeight(ctx context.Context, cl atypes.QueryClient, chainVer xchai
...
+//nolint:nonamedreturns // named returned for clarity
+func lookbackRange(
+	ctx context.Context,
+	attCl atypes.QueryClient,
+	chainVer xchain.ChainVersion,
+	chainName string,
+	fromOffset uint64,
+	latestHeight uint64,
+) (start uint64, end uint64, err error) {
+	const endpoint = "search_offset"
```
corverroos, line 743: copy paste, lets improve this label, also add latency metrics for this.

lib/cchain/provider/abci.go
```diff
@@ -763,3 +671,134 @@ func attsFromAtHeight(ctx context.Context, cl atypes.QueryClient, chainVer xchai
...
+func binarySearch(
+	ctx context.Context,
+	attCl atypes.QueryClient,
+	chainVer xchain.ChainVersion,
+	chainName string,
+	fromOffset uint64,
+	startHeight uint64,
+	endHeight uint64,
+) (uint64, error) {
+	const endpoint = "search_offset"
```
corverroos, line 686: copy paste, lets improve this label, also add latency metrics for this.

lib/cchain/provider/abci.go
```diff
@@ -763,3 +671,134 @@ func attsFromAtHeight(ctx context.Context, cl atypes.QueryClient, chainVer xchai
...
+//
+//nolint:nonamedreturns // named returned for clarity
+func lookbackRange(
+	ctx context.Context,
+	attCl atypes.QueryClient,
+	chainVer xchain.ChainVersion,
+	chainName string,
+	fromOffset uint64,
+	latestHeight uint64,
+) (start uint64, end uint64, err error) {
```
corverroos, line 742: lets use the actual variables: `startHeight, endHeight`

lib/cchain/provider/abci.go
```diff
@@ -763,3 +671,134 @@ func attsFromAtHeight(ctx context.Context, cl atypes.QueryClient, chainVer xchai
...
+	chainVer xchain.ChainVersion,
+	chainName string,
+	fromOffset uint64,
+	startHeight uint64,
+	endHeight uint64,
+) (uint64, error) {
+	const endpoint = "search_offset"
+
+	for steps := 0; startHeight <= endHeight; steps++ {
+		midHeight := startHeight + umath.SubtractOrZero(startHeight, startHeight)/2
```
corverroos, line 689: bug in this math

lib/cchain/provider/abci.go
```diff
@@ -763,3 +671,134 @@ func attsFromAtHeight(ctx context.Context, cl atypes.QueryClient, chainVer xchai
...
+
+		earliestAtt, ok, err := queryEarliestAttestation(ctx, attCl, chainVer, midHeight)
+		if err != nil {
+			incQueryErr(endpoint)
+			return 0, errors.Wrap(err, "abci query earliest-attestation-in-state")
+		}
+		if !ok {
```
corverroos, line 696: nit: I like to combine all validation into a single logical block, so prefer to use `else if`

lib/cchain/provider/abci.go
```diff
@@ -763,3 +671,134 @@ func attsFromAtHeight(ctx context.Context, cl atypes.QueryClient, chainVer xchai
...
+			// If we're so far back that there's no attestation at all, move forward
+			startHeight = midHeight + 1
+			continue
+		}
+
```
corverroos, line 701: can add here:
```
		if fromOffset == earliestAtt.AttestOffset {
			binarySearchStepsMetric(chainName, steps)
			return midHeight, nil
		}
```

lib/cchain/provider/abci.go
corverroos: in function comments, focus on what it returns, from your comment it looks like it returns attestations?

lib/cchain/provider/abci.go
```diff
@@ -357,40 +356,67 @@ func newABCIFetchFunc(attCl atypes.QueryClient, cmtCl cmtservice.ServiceClient,
...
-		offsetHeight, err := searchOffsetInHistory(ctx, cmtCl, attCl, chainVer, chainName, fromOffset)
+		latestBlockResp, err := cmtCl.GetLatestBlock(ctx, &cmtservice.GetLatestBlockRequest{})
+		if err != nil {
+			return []xchain.Attestation{}, 0, errors.Wrap(err, "query latest block")
+		}
+		latestHeight := uint64(latestBlockResp.SdkBlock.Header.Height)
+
+		var startHeight, endHeight uint64
+		// if we are performing search for the first time and no previous height was cached do a full search
+		if cachedHeight == 0 {
```
corverroos, line 389: suggest:
```

		// Binary search range from cached to latest
		searchStart, searchEnd := cachedHeight, latestHeight
		if cachedHeight == 0 {
			// Unless no cached height provided, then perform lookback search
			searchStart, searchEnd, err = lookbackRange(ctx, attCl, chainVer, chainName, fromOffset, latestHeight)
			if err != nil {
				incQueryErr(endpoint)

				return nil, 0, errors.Wrap(err, "lookback search")
			}
		}
```

lib/cchain/provider/abci.go
```diff
@@ -357,40 +356,67 @@ func newABCIFetchFunc(attCl atypes.QueryClient, cmtCl cmtservice.ServiceClient,
...
+			// but increased by one since we are searching the next attestation offset
+			startHeight = cachedHeight + 1
+			endHeight = latestHeight
+		}
+
+		offsetHeight, err := binarySearch(ctx, attCl, chainVer, chainName, fromOffset, startHeight, endHeight)
 		if err != nil {
 			incQueryErr(endpoint)
-			return nil, errors.Wrap(err, "searching offset in history")
+
```
corverroos, line 406: no need to add newlines before return if the function only had two lines

lib/cchain/provider/abci.go
```diff
@@ -344,7 +343,7 @@ func newABCIAllAttsFunc(cl atypes.QueryClient) allAttsFunc {
 }
 
 func newABCIFetchFunc(attCl atypes.QueryClient, cmtCl cmtservice.ServiceClient, chainNamer func(xchain.ChainVersion) string) fetchFunc {
-	return func(ctx context.Context, chainVer xchain.ChainVersion, fromOffset uint64) ([]xchain.Attestation, error) {
+	return func(ctx context.Context, chainVer xchain.ChainVersion, fromOffset uint64, cachedHeight uint64) ([]xchain.Attestation, uint64, error) {
```
corverroos, line 346: we need add comments somewhere about what this function returns

lib/cchain/provider/provider.go
```diff
@@ -179,9 +183,18 @@ func (p Provider) stream(
 	srcChain := p.chainNamer(chainVer)
 	ctx := log.WithCtx(in, "src_chain", srcChain, "worker", workerName)
 
+	// cachedOffsetHeight is used to store the height at which the previous attestation offset was found
```
corverroos, line 186: nit: `// Cache the previous batch fetch height, to prevent lookback searching each time`

lib/cchain/provider/provider.go
```diff
@@ -179,9 +183,18 @@ func (p Provider) stream(
...
+	var cachedOffsetHeight uint64
+
 	deps := stream.Deps[xchain.Attestation]{
 		FetchBatch: func(ctx context.Context, _ uint64, offset uint64) ([]xchain.Attestation, error) {
-			return p.fetch(ctx, chainVer, offset)
+			atts, foundHeight, err := p.fetch(ctx, chainVer, offset, cachedOffsetHeight)
+			if err != nil {
+				return nil, err
+			}
+			cachedOffsetHeight = foundHeight
```
corverroos, line 195: There is a bug here...

corverroos: hint: you need to make the contracts of your functions clearer in comments, specifically edge cases, then callers need to handle those edge cases.

lib/cchain/provider/provider_test.go
```diff
@@ -230,7 +234,7 @@ func (f *testFetcher) Fetch(ctx context.Context, chainVer xchain.ChainVersion, f
 
 	f.fetched += len(resp)
 
-	return resp, nil
+	return resp, cachedHeight, nil
```
corverroos, line 237: I don't understand why we do this, why not just ignore it and return 0 everywhere?

lib/cchain/provider/abci.go
sideninja: I intentionally left them that way to not break existing counters for the same operation. I will change them if you think it's better.

kc1116 (approved):

corverroos (changes requested):
Just blocking this until we have fully fixed and confirmed the dedup streams fix, then we can get this in afterwards

lib/cchain/provider/abci.go
```diff
@@ -763,3 +664,141 @@ func attsFromAtHeight(ctx context.Context, cl atypes.QueryClient, chainVer xchai
...
+		} else {
+			startHeight = midHeight + 1
+		}
+	}
+
+	return 0, errors.New("unexpectedly reach end of search method [BUG]")
+}
+
+// lookbackRange does an exponential lookback from the latest block height provided
+// until attestation found has lower offset than the provided from offset, this
```
corverroos, line 730: lower or equal offset

lib/cchain/provider/provider.go
```diff
@@ -29,7 +29,10 @@ import (
 
 var _ cchain.Provider = Provider{}
 
-type fetchFunc func(ctx context.Context, chainVer xchain.ChainVersion, fromOffset uint64) ([]xchain.Attestation, error)
+// fetchFunc fetches all attestations at offset for provided chain version. It returns found
+// attestations as well as the consensus block height at which attestations were found.
+// If offset is found at the latest block attestations are returned but height returned is 0.
+type fetchFunc func(ctx context.Context, chainVer xchain.ChainVersion, fromOffset uint64, cachedHeight uint64) ([]xchain.Attestation, uint64, error)
```
corverroos, line 35: I'd suggest (nit, rename `cachedHeight` to `searchStartHeight`):
```
fetchFunc returns a slice of strictly-sequential attestations for the provided fromOffset (inclusive) and chain version. It also returns the consensus block height where these attestations were found. This height can be provided in subsequent calls as searchStartHeight to improve performance by skipping the internal lookbackRange step. Providing a zero searchStartHeigt first performs an exponential lookback to identify binary search range.

If the consensus chain doesn't contain the fromOffset yet, the latest height is returned since this is safe to use as searchStartHeight. 
```

sideninja: Added a comment. The variable name is better. One nit feedback tho is that I believe comments explaining interfaces shouldn't contain implementation details. In this case I believe explaining about lookbackRange etc is an implementation detail, and since interfaces are APIs we shouldn't explain how things get implemented. I think the comment could end at: searchStartHeight is used to narrow the height search range.

lib/cchain/provider/abci.go
```diff
@@ -357,40 +356,60 @@ func newABCIFetchFunc(attCl atypes.QueryClient, cmtCl cmtservice.ServiceClient,
...
 			incQueryErr(endpoint)
-			return nil, errors.Wrap(err, "abci query attestations-from")
+			return nil, 0, errors.Wrap(err, "abci query attestations-from")
 		} else if ok {
-			fetchStepsMetrics(chainName, 0, 0)
-			return atts, nil
+			binarySearchStepsMetric(chainName, 0)
+			lookbackStepsMetric(chainName, 0)
+
+			return atts, 0, nil
```
corverroos, line 364: edge case not handled wrt opaque cursor

lib/cchain/provider/abci.go
```diff
@@ -357,40 +356,60 @@ func newABCIFetchFunc(attCl atypes.QueryClient, cmtCl cmtservice.ServiceClient,
...
-			return nil, errors.Wrap(err, "abci query earliest-attestation-in-state")
+			return nil, 0, errors.Wrap(err, "abci query earliest-attestation-in-state")
 		}
 
 		// Either no attestations have happened yet, or the queried fromOffset is in the "future"
 		// Caller has to wait and retry in both cases
 		if !ok || earliestAttestationAtLatestHeight.AttestOffset < fromOffset {
 			// First attestation hasn't happened yet, return empty
-			return []xchain.Attestation{}, nil
+			return []xchain.Attestation{}, 0, nil
```
corverroos, line 377: as mentioned, we can return latest height as opaque cursor here.

lib/cchain/provider/provider.go
```diff
@@ -29,7 +29,12 @@ import (
 
 var _ cchain.Provider = Provider{}
 
-type fetchFunc func(ctx context.Context, chainVer xchain.ChainVersion, fromOffset uint64) ([]xchain.Attestation, error)
+// fetchFunc returns a slice of strictly-sequential attestations for the
+// provided fromOffset (inclusive) and chain version if present.
+//
+// It also accepts (optional) and returns an opaque cursor (when error is nil).
+// When streaming attestation, the returned cursor can be provided in the subsequent call for improved performance.
+type fetchFunc func(ctx context.Context, chainVer xchain.ChainVersion, fromOffset uint64, searchStartHeight uint64) ([]xchain.Attestation, uint64, error)
```
corverroos, line 37: not, rename to `searchStartHeight` to `cursor` to align with comment

lib/cchain/provider/provider.go
```diff
@@ -179,9 +188,19 @@ func (p Provider) stream(
 	srcChain := p.chainNamer(chainVer)
 	ctx := log.WithCtx(in, "src_chain", srcChain, "worker", workerName)
 
+	// Cache the previous batch fetch height, to prevent lookback searching each time
+	var cachedOffsetHeight uint64
```
corverroos, line 192: nit: rename to `fetchCursor` to align with new comments of interface, since this is a user of the interface and should follow its docs.

corverroos (approved):

PR #2326, opened 2024-10-30
chore(docs): add mainnet

fabtreb:
Add mainnet

issue: none

### DIFF
--- a/docs/content/build/5-mainnet.md
+++ b/docs/content/build/5-mainnet.md
@@ -1,0 +1,41 @@
+# Mainnet
+
+**🚧 COMING SOON**
+
+
+### Have Questions?
+
+Join our developer Telegram channel [here](https://t.me/omnidevsupport)!
+
+### Overview
+
+This document lays out relevant developer resources for the Omni mainnet. Please read below for necessary RPC URLs, Explorers, Tools, Libraries and Documentation.
+
+
+### General Information
+
+
+### RPC URLs
+
+| **Chain Name** | **Chain ID** | RPC URL |
+| --- | --- | --- |
+| Omni mainnet | 166 | [See on ChainList](https://chainlist.org/chain/166) |
+| Ethereum | 1 | [See on ChainList](https://chainlist.org/chain/1) |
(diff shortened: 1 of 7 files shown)


docs/content/build/5-mainnet.md
```diff
@@ -0,0 +1,41 @@
+# Mainnet
```
ttarsi, line 1: This page is titled `5-mainnet`, but is 3rd in the sidebar (defined in screenshot below). Let's be consistent – the numbers in the title of the file are for convenience (documentation on how we use this [here](https://github.com/omni-network/omni/tree/main/docs/content), but feel free to update it).

Screenshot 2024-10-30 at 1 29 57 PM

docs/content/build/5-mainnet.md
```diff
@@ -0,0 +1,41 @@
...
+| Ethereum | 1 | [See on ChainList](https://chainlist.org/chain/1) |
+| Arbitrum One | 42161 | [See on ChainList](https://chainlist.org/chain/42161) |
+| Optimism | 10 | [See on ChainList](https://chainlist.org/chain/10) |
+| Base | 8453 | [See on ChainList](https://chainlist.org/chain/8453) |
+
+### Resources
+
+| **Resource** | **Address** |
+| --- | --- |
+| Archive Node(s) RPC | Available upon request |
```
ttarsi, line 32: Let's just point them to quicknode

ttarsi (approved):
Approving with a couple comments

- Nit but I think now that we have `Mainnet`, it would look better to have `Testnet` and `Mainnet` for simplicity (rather than `Omega Testnet`)
Screenshot 2024-10-30 at 1 32 04 PM

docs/content/build/5-mainnet.md
fabtreb: Bothered me too. I started updating the numbers but reverted to make the review easier, without deleted / new files. Suggest another PR to actually remove the numbering on the file names.

fabtreb:
> Approving with a couple comments
> 
> * Nit but I think now that we have `Mainnet`, it would look better to have `Testnet` and `Mainnet` for simplicity (rather than `Omega Testnet`)

Agree, will update in separate PR

PR #2330, opened 2024-10-30
chore(lib/stream): dont count empty fetches as misses

corverroos:
Don't count empty fetches as cache misses, since the cache couldn't contain the data.

issue: none


lib/stream/stream.go
```diff
@@ -103,22 +102,7 @@ func Stream[E any](ctx context.Context, deps Deps[E], srcChainID uint64, startHe
 				continue
 			}
 
-			heightsOK := true
```
corverroos, line 106: we actually check strict ordering below already, can remove this.

chmllr (approved):

PR #2348, opened 2024-10-31
feat(contracts/core): start solve inbox

kevinhalliday:
Start solve inbox prototype.

Introduce single Inbox.request() method.

issue: none


contracts/core/src/solve/Inbox.sol
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+
+    /**
+     * @dev Increment and return _lastId.
+     */
+    function _nextId() internal returns (uint256) {
+        _lastId++;
+        return _lastId;
+    }
```
kevinhalliday, line 113: @Zodomo what was argument for bytes32 w/ assembly increment you used in prototype again?

Zodomo:
```solidity
assembly {
    // Load from storage slot
    guid := sload(nextGuid.slot)
    // Increment and store back
    sstore(nextGuid.slot, add(guid, 1))
}
```

Where `nextGuid` is a storage variable

kevinhalliday: yes - what was the motivation? why better than just `uint256` and plain ol sol

Zodomo: Ah, bytes32 values are smaller and easier to pass around. They visually "look" more like a guid. Just a style choice tbh. We can also opt for truly random guid's that are always 64 characters rather than unwieldly large numbers that can far outpace a bytes32 in text length.

Zodomo: `115792089237316195423570985008687907853269984665640564039457584007913129639935`
`0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`

Purely a style decision imo. The assembly is just to increment it without wasting gas on casting.

kevinhalliday: bytes32 and uint256 are both 32 bytes not?

kevinhalliday: They are more visually pleasing, indeed

contracts/core/src/solve/Solve.sol
```diff
@@ -0,0 +1,69 @@
...
+     */
+    struct Request {
+        uint256 id;
+        uint96 openedAt;
+        address from;
+        bool accepted;
+        bool rejected;
+        bool cancelled;
+        bool fulfilled;
+        bool paid;
```
kevinhalliday, line 27: Going to switch these to a uint8 status. doesn't change number of slots packed

contracts/core/src/solve/Inbox.sol
kevinhalliday: I'll switch to bytes32 and then we can have a look at gas savings if we do assmbly bit

contracts/core/src/solve/Solve.sol
Zodomo: Could use an enum too, it's a uint8 under the hood

contracts/core/src/solve/Inbox.sol
```diff
@@ -0,0 +1,114 @@
...
+     */
+    function _openRequest(address from, Solve.Call calldata call, Solve.TokenDeposit[] calldata deposits)
+        internal
+        returns (Solve.Request storage req)
+    {
+        bytes32 id = _nextId();
+
+        req = _requests[id];
+        req.id = id;
+        req.updatedAt = uint96(block.timestamp);
```
Zodomo, line 87: Use uint40 for timestamps. They support dates as far as the year 3593.

contracts/core/src/solve/Inbox.sol
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /**
+     * @dev Open a new request in storage at `id`.
+     *      Transfer token deposits from msg.sender to this inbox.
+     *      Duplicate token addresses are allowed.
+     */
+    function _openRequest(address from, Solve.Call calldata call, Solve.TokenDeposit[] calldata deposits)
+        internal
+        returns (Solve.Request storage req)
```
Zodomo, line 81: Leveraging a storage variable in the function return allows you to write clean-looking code, but is very inefficient when it comes to gas costs. Consider returning a memory object, and update storage at the very end. The costs for these incremental storage updates add up very quickly.

Here's an example:
```solidity
function _openRequest(address from, Solve.Call calldata call, Solve.TokenDeposit[] calldata deposits)
    internal
    returns (Solve.Request memory req)
{
    bytes32 id = _nextId();

    // Update the request in memory
    req.id = id;
    req.updatedAt = uint96(block.timestamp);
    req.from = from;
    req.call = call;

    if (msg.value > 0) {
        req.deposits.push(Solve.Deposit({ isNative: true, token: address(0), amount: msg.value }));
    }

    for (uint256 i = 0; i < deposits.length; i++) {
        if (deposits[i].amount == 0) revert InvalidDeposit();
        if (deposits[i].token == address(0)) revert InvalidDeposit();

        req.deposits.push(Solve.Deposit({ isNative: false, token: deposits[i].token, amount: deposits[i].amount }));

        // Transfer tokens to this contract
        IERC20(deposits[i].token).safeTransferFrom(msg.sender, address(this), deposits[i].amount);
    }

    // Now that req is fully populated in memory, assign it to storage with a single write
    _requests[id] = req;
}
```

contracts/core/src/solve/Solve.sol
```diff
@@ -0,0 +1,70 @@
...
+     * @param updatedAt     Timestamp request status was last updated.
+     * @param from          Address of the user who created the request.
+     * @param fulfilledBy   Address of the solver that fulfilled the request.
+     * @param status        Request status (open, accepted, cancelled, rejected, fulfilled, paid).
+     * @param call          Details of the call to be executed on another chain.
+     * @param deposits      Array of deposits backing the request.
+     */
+    struct Request {
+        bytes32 id;
+        uint96 updatedAt;
```
Zodomo, line 26: uint40 is all you need for a timestamp as it will work until the year 3593. The smaller variable will also let you move Status into the same slot.

contracts/core/src/solve/Inbox.sol
```diff
@@ -0,0 +1,114 @@
...
+        payable
+        nonReentrant
+        returns (bytes32 id)
+    {
+        if (call.target == address(0)) revert InvalidCall();
+        if (call.destChainId == 0) revert InvalidCall();
+        if (call.data.length == 0) revert InvalidCall();
+        if (deposits.length == 0 && msg.value == 0) revert NoDeposits();
+
+        Solve.Request storage req = _openRequest(msg.sender, call, deposits);
```
Zodomo, line 67: If `_openRequest` sets storage within its call (referenced [here](https://github.com/omni-network/omni/pull/2348/files#r1826000433)), this should be changed to a memory variable. Using storage here when storage is not being updated just leads to higher read costs.

contracts/core/src/solve/Inbox.sol
kevinhalliday: But the we can't push to array 😩

Still probably worth it

kevinhalliday: And we still have to push to the storage arraray in the end

contracts/core/src/solve/Inbox.sol
kevinhalliday: it's a question of when do we resolve the actualy `req.deposits`. Are we resolving it in memory then looping to push . or are we just resolving it once in storage and returning it

contracts/core/src/solve/Inbox.sol
kevinhalliday: so it's kinda a question of where to we want to resolve the true `req.deposits` - in memory then push storage. or just once in memory

kevinhalliday: Qs like these we can answer once we merge in and have gas snapshots

Zodomo (approved):

PR #2350, opened 2024-10-31
feat(monitor/xfeemngr): slow sync on ethereum

kevinhalliday:
Increase sync interval for eth mainnet and omega.

issue: none

### DIFF
--- a/monitor/xfeemngr/oracle.go
+++ b/monitor/xfeemngr/oracle.go
@@ -20,6 +20,7 @@ import (
 
 type feeOracle struct {
 	chain  evmchain.Metadata   // source chain
+	tick   ticker.Ticker       // ticker to sync fee oracle
 	toSync []evmchain.Metadata // chains to sync on fee oracle
 	gprice *gasprice.Buffer    // gas price buffer
 	tprice *tokenprice.Buffer  // token price buffer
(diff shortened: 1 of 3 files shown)


monitor/xfeemngr/oracle.go
```diff
@@ -34,6 +35,12 @@ func makeOracle(chain netconf.Chain, toSync []evmchain.Metadata, ethCl ethclient
 		return feeOracle{}, errors.New("chain metadata not found", "chain", chain.ID)
 	}
 
+	syncInterval := feeOracleSyncInterval
+	override, ok := chainSyncOverrides[chain.ID]
+	if ok {
+		syncInterval = override
+	}
+
```
kc1116, line 43: nit: Put this is a func 

```
// chainSyncInterval return chain sync interval. 
func chainSyncInterval() time.Duration {
	if override, ok := chainSyncOverrides[chain.ID]; ok {
		return override
	}
	
	return feeOracleSyncInterval
}
```

kc1116 (approved):

monitor/xfeemngr/xfeemngr.go
```diff
@@ -89,7 +95,7 @@ func Start(ctx context.Context, network netconf.Network, cfg Config, privKeyPath
 		gprice:  gprice,
 		tprice:  tprice,
 		oracles: oracles,
-		ticker:  ticker.New(ticker.WithInterval(feeOracleSyncInterval)),
+		// ticker:  ticker.New(ticker.WithInterval(feeOracleSyncInterval)),
```
chmllr, line 98: Why not removing?

chmllr (approved):

monitor/xfeemngr/xfeemngr.go
kevinhalliday: it should be removed

Issue #2321, opened 2024-10-29
Dedup relayer streams


corverroos:
#### Problem to Solve
See [ADR](https://www.notion.so/omni-network/Dedup-relayer-streams-12769a99643a80c7b78cea02dbea35ba)

kc1116:
Slack [thread](https://omni-chain.slack.com/archives/C02HCJN6KNE/p1730368439553649?thread_ts=1730356199.896509&cid=C02HCJN6KNE) related to original implementation

corverroos:
Closing this as "not done" as we did two implementations that didn't solve the problem. This feels quite complex (caching often is). Hopefully #1860 solves the problem in the short to medium term.

PR #2351, opened 2024-11-01
feat(relayer): add leader streamer

corverroos:
Implements a leader based attest stream cache that aims to avoid overlapping streams thereby deduping the query load on the archive node.

issue: #2321

### DIFF
--- a/relayer/app/leadstream.go
+++ b/relayer/app/leadstream.go
@@ -2,14 +2,17 @@ package relayer
 
 import (
 	"context"
+	"sync"
+	"time"
+
 	"github.com/omni-network/omni/lib/cchain"
 	"github.com/omni-network/omni/lib/errors"
 	"github.com/omni-network/omni/lib/log"
 	"github.com/omni-network/omni/lib/netconf"
+	"github.com/omni-network/omni/lib/umath"
 	"github.com/omni-network/omni/lib/xchain"
+
 	"github.com/tidwall/btree"
-	"sync"
-	"time"
 )
 
 // errStop is returned when the leader should stop streaming.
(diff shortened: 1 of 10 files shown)


relayer/app/leadstream.go
```diff
@@ -0,0 +1,281 @@
...
+		return time.Since(t0) > duration
+	}
+}
+
+// leader tracks a worker actively streaming attestations and adding them to the cache.
+// It "locks" a range of offsets as leader of that range.
+type leader struct {
+	mu     sync.RWMutex
+	from   uint64
+	latest uint64
```
chmllr, line 60: It feels like `latest` is not really needed because it's only used to _guess_ if the range was covered by somebody else or not (this does not directly prove that the range is consistenty filled). So in case of some invariant violation (for whatever reason) we can get a leader with inconsistent range coverage that will prevent others from starting on its range.

Instead we could not rely on ranges at all and simply start a new leader on _any_ height that is a cache miss — it will ultimately bump into somebody's range (or timeout).

chmllr (approved):

relayer/app/leadstream.go
corverroos: nope, this is actually critical, since, latest is increased as soon as previous is fetched, this ensure that this leader remains the leader and others don't also start streaming from next as soon as it isn't in the cache.

corverroos:
I do realise an issue with design though: “blast radius”:
- if leader is slow in processing X, others are slowed down artificially.
- this can happen when `mempoolLimit` is reached
- Risk of artificially slowing down xchain latency isn’t worth the optimisation.

Two options:
- 1️⃣ Follow a more “orchestration” based pattern where workers request stuff and the orchestrator does the actual streaming. It populates the cache while workers only read from the cache. The orchestrator is fast, since it only fetches. Slow workers blast radius is contained.
- 2️⃣ Since this is the second design that isn’t good enough, I suggest we first do #1860 which should solve the problem. We can come back to this if we see that caching historical height isn’t sufficient and we need more optimisations.

Issue #2275, opened 2024-10-24
Periodic xmsgs on mainnet


corverroos:
#### Problem to Solve
Mainnet xchain isn't busy yet. We'd like to get some load on the system.

#### Proposed Solution
- Create a `RoleTester` for mainnet, but not fireblocks based, rather GCP.
- Add thresholds for it and ensure it gets funded.
- Add a package `scripts/sendxmsg`
- Write Integration test that does the following:
  - Define a `-count`, `-network`, `-role` in a config
  - Download `role` key from GCP
  - Instantiate xconnector
  - Get the `ethbackend` for the source chain, add the private key to it.
  - Implement and call `GetPortal(chainID uint64, backend ethbackend.Backend) bindings.OmniPortal` on the xconnector
  - Loop `count` times:
    - Pick a random source and destination, excluding eth L1s.
    - Do a simple xcall to 0xdead
- Write a github action in ops repo that:
  - On some predefined cron schedule (few times a day)
 
Use `TestJoinNetwork` as an example.

Zodomo:
I am not yet ready to start making changes to the Go codebase yet, but alternatives external to our infrastructure include automation platforms like [Gelato Web3 Functions](https://docs.gelato.network/web3-services/web3-functions) (50% gas premium on L2s), [Chainlink Automation](https://docs.chain.link/chainlink-automation) (50% gas premium on L2s), and [OpenZeppelin Actions](https://docs.openzeppelin.com/defender/module/actions) (no cost, self-funded, requires JS/TS).

The primary counterpoint to this is that these platforms might not be quick to support the Omni EVM, so if we want to send messages into Omni, we will likely need to rely on the internal solution described in the issue. I just wanted to provide additional commentary from the perspective of what I can do personally to tackle this issue. Doing it internally costs less, and gives us more flexibility. It just requires Go knowledge to do.

kc1116:
Updates https://omni-chain.slack.com/archives/C02HCJN6KNE/p1730726059890279

PR #2352, opened 2024-11-01
feat(e2s): add RoleTester service account key for mainnet part 1

kc1116:
This PR adds a RoleTester key for mainnet so that we can submit xmsgs to mainnet via a script. 

issue: #2275


kc1116:
Currently the key generation tool assumes that all keys being created for a application service. This PR adds the `Service` key type so that we can generate keys for utility purposes.

```
 go run e2e/main.go key-create -f e2e/manifests/mainnet.toml --name=tester --type=service
```
We generated a new _service_ key for mainnet and mapped it to the RoleTester. Now we can access this key in the e2e app the same way we access RoleTester for other networks.

chmllr (approved):
Code LGTM, can't validate the logic yet.

e2e/app/eoa/static.go
```diff
@@ -46,5 +46,6 @@ var statics = map[netconf.ID][]Account{
 		remote("0x9496Bf1Bd2Fa5BCba72062cC781cC97eA6930A13", RoleDeployer),
 		secret("0xfD62020Cee216Dc543E29752058Ee9f60f7D9Ff9", RoleMonitor),
 		secret("0x6191442101086253A636aecBCC870e4778490AaB", RoleRelayer),
+		secret("0x84aC7eeC1e035Fa13a7DEBD5f6da69B82CBDC5B5", RoleTester),
```
sideninja, line 49: wouldn't it be better for this account to be remote type, managed by Fireblocks?

e2e/app/key/key.go
```diff
@@ -22,14 +22,15 @@ const (
 	P2PConsensus Type = "p2p_consensus"
 	P2PExecution Type = "p2p_execution"
 	EOA          Type = "eoa"
+	Service      Type = "service"
```
corverroos, line 25: no need for a new type, this is an eoa

e2e/manifests/mainnet.toml
```diff
@@ -4,6 +4,9 @@ public_chains = ["ethereum","arbitrum_one","base","optimism"]
 multi_omni_evms = true
 prometheus   = true
 
+[keys.tester]
```
corverroos, line 7: don't think this is the way to go, it deplicates whatever is in the eoa package. Follow the same patterns as relayer and monitor keys.

corverroos (changes requested):
Blocking review since christiaan aproved and this is the way to go

kc1116:
Did not have to do.

PR #2357, opened 2024-11-01
fix(ci): fix the log file label and run on bigger runner

chmllr:
Minor fix of log file and specifying a large instance to run the network join test.

issue: none

### DIFF
--- a/.github/workflows/ci-join.yaml
+++ b/.github/workflows/ci-join.yaml
@@ -45,6 +45,6 @@ jobs:
         uses: actions/upload-artifact@v4
         if: always()
         with:
-          name: failed-logs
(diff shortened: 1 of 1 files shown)


ttarsi (approved):

.github/workflows/ci-join.yaml
```diff
@@ -45,6 +46,6 @@ jobs:
         uses: actions/upload-artifact@v4
         if: always()
         with:
-          name: failed-logs
+          name: docker-logs
```
chmllr, line 49: These logs are uploaded on successful runs as well, so the label might be misleading for successful jobs.

Issue #2354, opened 2024-11-01
Fix Pnl metrics for OP chains


kevinhalliday:
#### Problem to Solve

OP chains return a receipt that includes L1 fee info. Currently we do not include that in our spend calculations, causing us to misreport spend.

#### Proposed Solution

Query and use OP fields when calculating spend

PR #2359, opened 2024-11-01
fix(relayer): fix tx spend calc for op chains

kevinhalliday:
Fix PnL metrics / logs for op chains.

OP receipts include L1GasUsed, L1GasPrice, and L1Fee (L1GasPrice * L1GasUsed).
We need to include these in spend calc.

issue: #2354


go.mod
```diff
@@ -294,6 +294,8 @@ require (
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
+require github.com/fjl/gencodec v0.0.0-20230517082657-f9840df7b83e
```
kevinhalliday, line 297: got tossed here with a go mod tidy, dunno why it's on it's own

lib/optypes/receipt.go
```diff
@@ -0,0 +1,89 @@
+// This file copies Receipt type from ethereum-optimism/op-geth/core/types/receipt.go
```
kevinhalliday, line 1: Couple things.

Would have imported, but op-geth is just a fork of geth that changes minimal things. So:
- we get an error that we import with one name but the module has another
- geth versions won't match

Figured copying a little is better.

As for comment at the top, would have added a copyright. But the [op file itself](https://github.com/ethereum-optimism/op-geth/blob/optimism/core/types/receipt.go) has go-ethereum copyright. Any thoughts? Is this comment fine?

lib/ethclient/ethclient.go
```diff
@@ -179,3 +181,28 @@ func (w Wrapper) EtherBalanceAt(ctx context.Context, addr common.Address) (float
 
 	return bf / params.Ether, nil
 }
+
+// OPTransactionReceipt returns the OP trasnaction receipt for the given hash.
+// OP receipts included additional fields for L1 fee info.
+func (w Wrapper) OPTransactionReceipt(ctx context.Context, txHash common.Hash) (*optypes.Receipt, error) {
+	const endpoint = "eth_getTransactionReceipt"
```
corverroos, line 188: should probably use a different name, what about `op_transaction_receipt`?

lib/optypes/receipt.go
corverroos: I don't think it matters that much, above is fine.

relayer/app/sender.go
```diff
@@ -15,19 +15,17 @@ import (
...
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
-	"github.com/ethereum/go-ethereum/params"
 )
 
 // Sender uses txmgr to send transactions a specific destination chain.
-type onSubmitFunc func(context.Context, *ethtypes.Transaction, *ethtypes.Receipt, xchain.Submission)
+type onSubmitFunc func(context.Context, ethclient.Client, *ethtypes.Transaction, *ethtypes.Receipt, xchain.Submission)
```
corverroos, line 28: why don't we replace the legacy `TransactionReceipt` with `OpTransactionReceipt` in `omni/lib/txmgr/txmgr.go:478`, then no need to query a second time?

lib/optypes/receipt.go
```diff
@@ -0,0 +1,89 @@
...
+const (
+	// ReceiptStatusFailed is the status code of a transaction if execution failed.
+	ReceiptStatusFailed = uint64(0)
+
+	// ReceiptStatusSuccessful is the status code of a transaction if execution succeeded.
+	ReceiptStatusSuccessful = uint64(1)
+)
+
+// Receipt represents the results of a transaction.
+type Receipt struct {
```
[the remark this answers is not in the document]
corverroos: Can maybe just name it `ethclient.Receipt` and then prefix the OP specific fields with `OP*`?

relayer/app/sender.go
kevinhalliday: I think I'd prefer that. but blast radius of that change felt pretty big.  i'll give it a try

relayer/app/pnl.go
```diff
@@ -166,23 +181,21 @@ func getFees(
 			fees.nETH += feesGwei
 			fees.nUSD += feesGwei * prices[tokens.ETH]
```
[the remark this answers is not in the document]
kevinhalliday: coingecko provider makes [those checks](https://github.com/omni-network/omni/blob/c0f4c682f846c603e2f3a64123219aa3b65d24a5/lib/tokens/coingecko/coingecko.go#L69)

Issue #2355, opened 2024-11-01
SolverNet Contracts Prototype


kevinhalliday:
Placeholder to be split into new tickets.

Golang:
 - https://github.com/omni-network/omni/issues/2363
 - https://github.com/omni-network/omni/issues/2364

Solidity:
- https://github.com/omni-network/omni/issues/2377

PR #2360, opened 2024-11-01
feat(contracts/core): add additional actions to inbox

Zodomo:
Replaced OpenZeppelin libs with gas-optimized Solady alternatives. Expanded solve Inbox to include accept, reject, and cancel logic. Also implemented tests for these flows.

issue: #2355


contracts/core/src/solve/Inbox.sol
```diff
@@ -104,6 +188,23 @@ contract Inbox is ReentrancyGuardUpgradeable {
...
+            if (deposit.isNative) {
+                (bool success,) = payable(msg.sender).call{ value: deposit.amount }("");
+                if (!success) revert TransferFailed();
+            } else {
+                IERC20(deposit.token).safeTransfer(msg.sender, deposit.amount);
+            }
+        }
+
+        // Delete request
+        delete _requests[req.id];
```
kevinhalliday, line 205: Was thinking a little more about this. Feel like it will make solver spec much easier if it can just query source inbox for status of of a request ID. Rather than indexing events and maintaining state in db. 

We can work with corver on this when he starts. But I propose:

- Let's keep things simple for noew and just keep requests, updating their status
- If it turns out we don't need view funcs for this, we can delete
- If it turns out we do, but we want gas rebase for deletions, we can move status to another mapping

kevinhalliday (commented):
- Let's make all methods `nonReentrant`. But definitely cancel. Otherwise we could have some malicious cancellations while opening a request. 
- Thinking we actually do keep requests around for now, and just change status (see my comment below). Think the view function will be helpful. We can switch later if we don't need it 

Also, can you move the function tests cases to their own files.

`Inbox_accept.t.sol`
`Inbox_reject.t.sol`
`Inbox_cancel.t.sol`

Hlps keeps tests files a bit smaller

kevinhalliday (approved):
Approving.

Let's switch to new state names tho from adr `Open -> Pending` and `Cancelled -> Reverted`

PR #2365, opened 2024-11-04
chore(e2e): add new role RoleXCaller

kc1116:
This PR adds a new role RoleXCaller that will be used to send xmsgs on differen't networks and source chain/dest chain combinations. We will create a github action that will periodically send xmsgs and it will use this account. 

issue: #2275


e2e/app/eoa/testdata/threshold_reference.json
```diff
@@ -14,8 +14,8 @@
     "target": "0.0100"
    },
    "hot": {
-    "min": "2.0080",
-    "target": "8.0800"
+    "min": "22.0080",
+    "target": "108.0800"
```
corverroos, line 18: these thresholds are too high for mainnet

corverroos (changes requested):

e2e/app/eoa/testdata/threshold_reference.json
sideninja: I think threshold for tester has to be decreased on mainnet, possibly we can do a dynamic overwrite of the threshold, see something similar here: https://github.com/omni-network/omni/blob/16feed3cd72b61f0b915ffe2fc5095edf35097ab/e2e/app/eoa/fund.go#L74

kc1116:
Closed for https://github.com/omni-network/omni/pull/2379

Issue #2384, opened 2024-11-05
Implement v0 solver architecture


corverroos:
#### Problem to Solve
With solver service running, next step is to implement basic architecture for handling events and submitting txs. This doens't need to be wired up, or tested. Lets first outline the high level approach.

#### Proposed Solution
See above

PR #2386, opened 2024-11-05
feat(solver/app): basic event processor

corverroos:
Basic solver architecture based on a well-abstracted event processor.

issue: #2384

### DIFF
--- a/solver/app/bindings.go
+++ b/solver/app/bindings.go
@@ -1,0 +1,182 @@
+package solver
+
+import (
+	"github.com/omni-network/omni/contracts/bindings"
+	"github.com/omni-network/omni/lib/errors"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+const (
+	statusInvalid   uint8 = 0
+	statusPending   uint8 = 1
+	statusAccepted  uint8 = 2
+	statusRejected  uint8 = 3
+	statusReverted  uint8 = 4
+	statusFulfilled uint8 = 5
+	statusClaimed   uint8 = 6
+)
+
+var (
+	inboxABI = mustGetABI(bindings.SolveInboxMetaData)
+
+	// Event log topics (common.Hash).
+	topicRequested = mustGetEventTopic(inboxABI, "Requested")
+	topicAccepted  = mustGetEventTopic(inboxABI, "Accepted")
+	topicRejected  = mustGetEventTopic(inboxABI, "Rejected")
+	topicReverted  = mustGetEventTopic(inboxABI, "Reverted")
+	topicFulfilled = mustGetEventTopic(inboxABI, "Fulfilled")
+	topicClaimed   = mustGetEventTopic(inboxABI, "Claimed")
+)
+
+// eventMeta contains metadata about an event.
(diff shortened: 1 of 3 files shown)


solver/app/processor.go
```diff
@@ -0,0 +1,74 @@
...
+			if err != nil {
+				return errors.Wrap(err, "parse id")
+			}
+
+			ctx := log.WithCtx(ctx, log.Hex7("req_id", reqID[:]))
+
+			req, _, err := deps.GetRequest(ctx, chainID, reqID)
+			if err != nil {
+				return errors.Wrap(err, "current status")
+			} else if event.Status != req.Status {
```
sideninja, line 46: does it make sense to also check for case where event status is further than request status and mark it as a bug? so event status is fullffilled but request status is pending, because if this happens (if it can) we might miss a bug here.

solver/app/processor_internal_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+const (
+	accept  = "accept"
+	reject  = "reject"
+	fulfill = "fulfill"
+	claim   = "claim"
+	ignored = ""
+)
+
+func TestEventProcessor(t *testing.T) {
```
sideninja, line 24: nice

sideninja (approved):

solver/app/processor.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/omni-network/omni/contracts/bindings"
+	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/log"
+	"github.com/omni-network/omni/lib/xchain"
+
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+// procDeps abstracts dependencies for the event processor allowed simplified testing.
+type procDeps struct {
```
kc1116, line 15: nit: this should be an interface, interface is used to do exactly what the comment says _"abstracts dependencies for ... allowed simplified testing"_

solver/app/processor.go
```diff
@@ -0,0 +1,74 @@
...
+			switch event.Status {
+			case statusPending:
+				reason, reject, err := deps.ShouldReject(ctx, chainID, req)
+				if err != nil {
+					return errors.Wrap(err, "should reject")
+				} else if reject {
+					return deps.Reject(ctx, chainID, req, reason)
+				}
+
+				return deps.Accept(ctx, chainID, req)
```
kc1116, line 60: Can you explain why we dont Fulfill at this step?

kc1116 (approved):
Couple nits related to Go styling, and one comment for clarification.

solver/app/bindings.go
```diff
@@ -0,0 +1,182 @@
...
+	statusPending   uint8 = 1
+	statusAccepted  uint8 = 2
+	statusRejected  uint8 = 3
+	statusReverted  uint8 = 4
+	statusFulfilled uint8 = 5
+	statusClaimed   uint8 = 6
+)
+
+var (
+	inboxABI = mustGetABI(bindings.SolveInboxMetaData)
```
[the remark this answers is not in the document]
corverroos: `init` should be avoided at all costs I think

solver/app/processor.go
corverroos: it is a style thing, I prefer funcs, since the implementations are not related, this allows better decoupling,

solver/app/processor.go
corverroos: First Accept, if that succeeds, then fulfil

PR #2388, opened 2024-11-05
chore(scripts/join): avoid building images

corverroos:
Add the `halo_tag` flag to `TestJoin`.

Default to using `main` docker images for `TestJoin`. This drops building docker images as a requirement, making it easier to run. Also just use `main` in CI.

issue: none

### DIFF
--- a/.github/workflows/ci-join.yaml
+++ b/.github/workflows/ci-join.yaml
@@ -27,19 +27,15 @@ jobs:
         with:
           go-version: 'stable'
 
-      - name: Build binaries
-        uses: goreleaser/goreleaser-action@v5
-        with:
-          version: 2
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-join.yaml
```diff
@@ -27,19 +27,15 @@ jobs:
...
-
       - name: Run join test
         run: |
           cd scripts/join
-          sudo go test . -v --integration --logs_file=docker_logs.txt --network="${{github.event.inputs.network || 'mainnet'}}" -timeout 0
+          sudo go test . -v \
+            --integration \
+            --timeout=0 \
+            --logs_file=docker_logs.txt \
+            --halo_tag="main" \
```
corverroos, line 37: this is just for clarity, main is the default in CI

PR #2389, opened 2024-11-05
test(e2e/app): disable testUpgradeBridgeNative

Zodomo:
Disable the test for OmniBridgeNative as it was failing because the bridge has not yet been upgraded. Post-upgrade the test can be reenabled.

issue: none

### DIFF
--- a/e2e/app/admin/test.go
+++ b/e2e/app/admin/test.go
@@ -60,9 +60,11 @@ func Test(ctx context.Context, def app.Definition) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


kevinhalliday (approved):
Try running admin tests locally. You can check gh workflow for command (right now they are only run on merge to main).

Zodomo:
Ran the `go run github.com/omni-network/omni/e2e admin test -f e2e/manifests/devnet1.toml` command from the `e2e-test-admin` workflow and it was successful.

PR #2399, opened 2024-11-06
chore(solver/app): add some wiring

corverroos:
Adds some wiring that constructs the `procDeps` and starts streaming.

issue: #2384

### DIFF
--- a/solver/app/deps.go
+++ b/solver/app/deps.go
@@ -1,0 +1,198 @@
+//nolint:dupl,unused // It's okay to have similar code for different events
+package solver
+
+import (
+	"context"
+
+	"github.com/omni-network/omni/contracts/bindings"
+	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/ethclient/ethbackend"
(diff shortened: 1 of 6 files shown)


kc1116 (approved):

solver/app/deps.go
```diff
@@ -0,0 +1,198 @@
...
+			return nil
+		}
+
+		// TODO(corver): Convert req.Deposits into TokenPreReqs
+		var prereqs []bindings.SolveTokenPrereq
+
+		tx, err := outbox.Fulfill(txOpts, req.Id, chainID, solverAddr, req.Call, prereqs)
+		if err != nil {
+			return errors.Wrap(err, "fulfill request")
+		} else if _, err := backend.WaitMined(ctx, tx); err != nil {
```
sideninja, line 95: nit: drop the else?

sideninja (approved):

solver/app/deps.go
corverroos: why? I feel it is part of the "fullfil" step, so locally part of the same thing.

Issue #2402, opened 2024-11-06
Use `prometheus.yaml` on live networks


ttarsi:
**Problem to Solve**

Converting from `yml` to `yaml` had some unintended consequences – vms on live networks have `.yml` files but the new images look for `.yaml` ones. We need to make sure all files on live networks / vms use `.yaml`

This needs to be done before any upgrades can happen.

**Proposed Solution**

When we upgrade, copy `prometheus.yaml` into the vm in the vmcompose upgrade command.

PR #2404, opened 2024-11-06
fix(e2e): copy prom config on upgrade

kevinhalliday:
Copy prometheus.yaml to vm on upgrade.

issue: #2402

### DIFF
--- a/e2e/vmcompose/provider.go
+++ b/e2e/vmcompose/provider.go
@@ -245,6 +245,20 @@ func (p *Provider) Upgrade(ctx context.Context, cfg types.ServiceConfig) error {
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


e2e/vmcompose/provider.go
```diff
@@ -245,6 +245,20 @@ func (p *Provider) Upgrade(ctx context.Context, cfg types.ServiceConfig) error {
 				}
 			}
 
+			log.Debug(ctx, "Copying prometheus.yaml", "vm", vmName)
+			agentFile := vmAgentFile(instance.IPAddress.String())
```
corverroos, line 249: nice, this is perfect, but I took the opportunity to simplify things a bit https://github.com/omni-network/omni/pull/2407/files

corverroos:
Not needed anymore

PR #2410, opened 2024-11-07
feat(solver/app): add cursor db

corverroos:
Adds a cursor DB to solver .

issue: #2384


solver/app/cursor.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	return &cursors{
+		table: dbStore.CursorTable(),
+	}, nil
+}
+
+// cursors provides a thread-safe persisted cursor store.
+type cursors struct {
+	mu    sync.Mutex
```
chmllr, line 57: Do we have to exclusively lock on reads?

chmllr (approved):

solver/app/cursor.go
kc1116, line 57:
```suggestion
	mu    sync.RWMutex
```

solver/app/cursor.go
```diff
@@ -0,0 +1,100 @@
...
+	table CursorTable
+}
+
+// Get returns the block height of the cursor for the given chain version, or false, or an error.
+func (c *cursors) Get(ctx context.Context, chainVer xchain.ChainVersion) (uint64, bool, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
kc1116, line 64:
```suggestion
	c.mu.RLock()
	defer c.mu.RUnlock()
```

solver/app/cursor.go
corverroos: to be safe yes, the cosmosDB is a single threaded thing, I'm not sure concurrent reads are supported or not. We only read from the DB on startup in any case.

corverroos: You are assuming this is safe? how do you know this?

PR #2417, opened 2024-11-07
chore(contracts/core): remove required paused statement

Zodomo:
Remove the require paused statement from the bridge tests as they fail when bridge is not paused.

issue: none


kevinhalliday (approved):
We can also just remove theses tests from `e2e/admin/test`

Issue #2412, opened 2024-11-07
Decrease relayer bootstrap duration


sideninja:
Relayer is taking a lot of time to bootstrap because it has to "catch-up" a lot of xblocks to find next attestation offsets it needs to relay.

This range is being searched using an exponential lookback and binary search which produces a lot of queries (~30 per offset). 

We have attempted to remove exponential lookback step using caching: https://github.com/omni-network/omni/issues/2044 

That introduced slower queries (for not yet known reason). 

We should look at alternatives to solve this. One idea is to persist the cache we've already introduced in the [previous PR](https://github.com/omni-network/omni/pull/2292).

sideninja:
The new proposed design is to store attestation offsets that were processed in a persistent storage. That makes the relayer bootstrap use that storage to narrow the search window on the source attestation module.

PR #2425, opened 2024-11-08
feat(relayer): persist cursors for attestation stream

sideninja:
Persisting cursors for each stream which is used during relayer bootstrap. 

We add a storage for cursors and monitor the stored cursors and confirm them once the submissions are finalized on the network. Confirmed cursors are also trimmed from the database. 

The confirmation rules are:
- Confirm an empty cursor if a previous cursor is confirmed
- Confirm a non empty cursor if a finalized submission from the network contains equal or higher attestation offset

Issue: #2412


relayer/app/app.go
```diff
@@ -1,4 +1,4 @@
-package relayer
```
sideninja, line 1: this was needed because generator will define the package as app (correctly) also I think it's better to have package named as they should be

scripts/buf.gen.orm.yaml
```diff
@@ -2,7 +2,7 @@ version: v1
 managed:
   enabled: true
   go_package_prefix:
-    default: github.com/omni-network/omni
+    standard: github.com/omni-network/omni
```
sideninja, line 5: seems like this had to be changed:

```
WARN	Category DEFAULT referenced in your buf.yaml is deprecated. It has been replaced by category STANDARD.
	The concept of a default rule has been introduced. A default rule is a rule that will be run
	if no rules are explicitly configured in your buf.yaml. Run buf config ls-lint-rules or
	buf config ls-breaking-rules to see which rules are defaults. With this introduction, having a category
	also named DEFAULT is confusing, as while it happpens that all the rules in the DEFAULT category
	are also default rules, the name has become overloaded.
	As with all buf changes, this change is backwards-compatible: DEFAULT will continue to work.
	We recommend replacing DEFAULT in your buf.yaml, but no action is immediately necessary.
```

relayer/app/cursors/cursor.go
```diff
@@ -0,0 +1,15 @@
+package cursors
```
corverroos, line 1: lets stick to singlular package names, not plural as per standard go practice?

relayer/app/cursors/watcher.go
```diff
@@ -0,0 +1,123 @@
...
+			uint32(stream.ConfLevel()),
+			stream.DestChainID,
+		))
+
+	if err != nil {
+		return nil, errors.Wrap(err, "list cursors")
+	}
+	defer iterator.Close()
+
+	var cursorsDesc []*Cursor
```
corverroos, line 39: what is `*Desc`? why not just `resp` or `cursors`?

relayer/app/cursors/cursors.proto
```diff
@@ -0,0 +1,24 @@
...
+import "cosmos/orm/v1/orm.proto";
+
+option go_package = "relayer/app";
+
+message Cursor {
+  option (cosmos.orm.v1.table) = {
+    id: 1;
+    primary_key: { fields: "src_chain_id,conf_level,dst_chain_id,attest_offset" }
+    index: { id: 1, fields: "confirmed" }
+    index: { id: 2, fields: "src_chain_id,conf_level,dst_chain_id" }
```
corverroos, line 14: not sure this index is required, since you have `CursorPrimaryKey{}.WithSrcChainIdConfLevelDstChainId` which is the same thing? Can remove?

relayer/app/cursors/cursors.proto
```diff
@@ -0,0 +1,24 @@
...
+  };
+
+  uint64 src_chain_id       = 1; // Chain ID as per https://chainlist.org
+  uint64 dst_chain_id       = 2; // Chain ID as per https://chainlist.org
+  uint32 conf_level         = 3; // Confirmation level of the cross-chain block
+  uint64 attest_offset      = 4; // Offset of the cross-chain block
+  uint64 first_xmsg_offset  = 5; // Offset of the first cross-chain message
+  uint64 last_xmsg_offset   = 6; // Offset of the last cross-chain message
```
corverroos, line 22: ah right, so this is not going to work. Multiple shards map to the same conf level. So we must store something like: `map[ShardID]map[XMsgRange]`. 

My suggestion, lets skip this xmsgOffset stuff, lets stick to original `bool empty`.
```
empty := Attestation > XBlock > Xmsgs > filter by shard ConfLevel > empty or not
confirmed := SubmitCursor(FinalizedRef).AtterstOffset >= AttestOffset
```

relayer/app/cursors/watcher.go
```diff
@@ -0,0 +1,123 @@
...
+	"context"
+	"slices"
+
+	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/xchain"
+)
+
+// todo mind the query limit on db
+
+type Watcher struct {
```
corverroos, line 13: please add godoc to all types of exported functions, also to package.

relayer/app/cursors/watcher.go
```diff
@@ -0,0 +1,123 @@
...
+// ConfirmStream checks all cursors for the provided stream and confirms them according to the rules:
+//   - if a cursor is not empty fetch finalized cursor from the network and confirm only if finalized
+//     message offset is higher to the cursor last message offset
+//   - if a cursor is empty, confirm it if the previous cursor was confirmed
+func (w *Watcher) ConfirmStream(ctx context.Context, stream xchain.StreamID) error {
+	cursors, err := w.getStreamCursors(ctx, stream)
+	if err != nil {
+		return err
+	}
+
```
corverroos, line 60: why not something like:
```
for i, c := range cursors {
  if c.Confirmed {
    continue
  }
  if !c.Empty() {
    finalized, err := isFinalized(ctx, c)
    if !finalized {
      return nil // Reached head
    }
  }
  
  setConfirmed(ctx, c)
}

```

relayer/app/cursors/cursors.proto
corverroos: also, `confirmed` index not used (yet)

relayer/app/cursors/watcher.go
sideninja: this is leftover from reverse ordered, keep in mind it's wip

relayer/app/cursors/watcher.go
sideninja: because you would have to fetch cursors from the network for each non-empty cursor even though you know what the latest finalized cursor is

sideninja: unless you cache the network request in isFinalized but then it's basically what I wrote. And if you suggest just extracting into more functions, that can be changed, I plan to go over that in the end again, but extracting can be improved

corverroos: at this point attestations are few and far between, not sure caching is required. I'd just stick to simple code rather?

corverroos: be careful of #prematureoptimisations

sideninja: sure we can always fetch, but in any case the logic above won't work because we can't just set to confirmed all, what if the cursors are all empty, you cant just confirm them, until you have one non-empty confirmed, above suggestion would just confirm empty ones.

relayer/app/cursors/cursors.proto
sideninja: Changed that now back, the issue I have is that cursors now don't persist message offset which is needed by the message filter https://github.com/omni-network/omni/blob/7fb91f4cf778e0b586aa85f0aa8098ab199c41f3/relayer/app/worker.go#L112

sideninja: I believe we still need to persist last msg offset

sideninja:
@corverroos let's sync on the shard / conf levels as well as the last message offset.

relayer/app/cursor/stream.go
```diff
@@ -0,0 +1,160 @@
...
+			return errors.Wrap(err, "delete cursor")
+		}
+	}
+
+	return nil
+}
+
+// isFinalized checks if the cursor is finalized on the network.
+func (s *StreamCursors) isFinalized(ctx context.Context, cursor *Cursor) (bool, error) {
+	final, ok, err := s.provider.GetSubmittedCursor(ctx, xchain.FinalizedRef, cursor.StreamID())
```
sideninja, line 152: Conf level to shard ID conversion, which I'm not sure about, but since we are not persisting shards now I wonder if that is ok

relayer/app/cursor/cursor.go
```diff
@@ -0,0 +1,47 @@
...
+	"cosmossdk.io/core/store"
+	"cosmossdk.io/orm/model/ormdb"
+	db "github.com/cosmos/cosmos-db"
+)
+
+func (c *Cursor) StreamID() xchain.StreamID {
+	return xchain.StreamID{
+		SourceChainID: c.GetSrcChainId(),
+		DestChainID:   c.GetDstChainId(),
+		ShardID:       xchain.ShardID(c.GetConfLevel()),
```
sideninja, line 19: used to fetch the cursors from network, want to confirm if it's ok for that usecase

corverroos: when fetching cursors from on-chain, we need to fetch all streams for the chain version, then use minimum attestOffset. We can't simply query the the first shard, there might be more and they might be lagging.

relayer/cmd/flags.go
```diff
@@ -14,4 +16,6 @@ func bindRunFlags(flags *pflag.FlagSet, cfg *relayer.Config) {
 	flags.StringVar(&cfg.PrivateKey, "private-key", cfg.PrivateKey, "The path to the private key e.g path/private.key")
 	flags.StringVar(&cfg.HaloURL, "halo-url", cfg.HaloURL, "The URL of the halo node e.g localhost:26657")
 	flags.StringVar(&cfg.MonitoringAddr, "monitoring-addr", cfg.MonitoringAddr, "The address to bind the monitoring server")
+	flags.StringVar(&cfg.DBDir, "db-dir", cfg.DBDir, "The path to the database directory")
+	flag.DurationVar(&cfg.ConfirmInterval, "attestaion-confirm-interval", cfg.ConfirmInterval, "Submitted attestations finalization check interval")
```
corverroos, line 20: no point in making this a flag, just hard code it.

corverroos (changes requested):

relayer/app/cursor/cursors.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// ConfirmedOffset loads confirmed cursor offset for the chain version and destination chain ID.
+// If a cursor doesn't exist a 0, false is returned.
+func (c *Cursors) ConfirmedOffset(
+	ctx context.Context,
+	chainVer xchain.ChainVersion,
+	destChainID uint64,
+) (uint64, bool, error) {
+	return newStreamCursors(chainVer.ID, destChainID, chainVer.ConfLevel, c.cursors, c.xProvider, "", "").
```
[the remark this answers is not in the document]
sideninja: Can you elaborate. If I understand you right you would completely remove the StreamCursors type and create functions that would all be passed in src chain, dest chain id, cursors table, xprovider, source name, destination name? That feels like they all operate on same data that's why it felt to me like they belong on an object owning the data and manipulating it. I can convert into functions if you think that's more appropriate tho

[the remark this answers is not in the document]
corverroos: yup, rather create a `stream` struct, which is the data:
```
type stream struct {
  SrcChain
  ConfLevel
  DestChain
}
```

then pass that into functions like `getConfirmed`
```
func getConfirmed(ctx, cursors, stream)
```

This splits logic from data. Instead of OO style which combines them.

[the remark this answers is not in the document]
sideninja: ok will change into this style

relayer/app/config.go
```diff
@@ -16,19 +17,23 @@ import (
...
-	HaloURL        string
-	Network        netconf.ID
-	MonitoringAddr string
+	RPCEndpoints    xchain.RPCEndpoints
+	PrivateKey      string
+	HaloURL         string
+	Network         netconf.ID
+	MonitoringAddr  string
+	DBDir           string
+	ConfirmInterval time.Duration
```
corverroos, line 26: Config is only for actual flags. This can be hardcoded in `NewCursors` or its caller.

relayer/app/cursor/cursor.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/xchain"
+
+	ormv1alpha1 "cosmossdk.io/api/cosmos/orm/v1alpha1"
+	"cosmossdk.io/core/store"
+	"cosmossdk.io/orm/model/ormdb"
+	db "github.com/cosmos/cosmos-db"
+)
+
+func (c *Cursor) StreamID(shardID xchain.ShardID) xchain.StreamID {
```
corverroos, line 15: this isn't great. Providing a shard, and then returning the stream feels like incorrect abstractions. Someone can provide any shardID, not matching the actual conf level. I don't think should be a method on the cursor.

relayer/app/cursor/cursors.go
```diff
@@ -0,0 +1,156 @@
...
+		Empty:        empty,
+	}
+
+	err := c.db.Save(ctx, cursor)
+	if err != nil {
+		return errors.Wrap(err, "insert cursor")
+	}
+
+	log.Debug(ctx, "New cursor saved",
+		"src_chain", srcVersion.ID,
```
corverroos, line 107: never log chain ID, always log human readable strings. I suggest you use `c.network.ChainVersionName(srcVersion` and `destChain.Name`.

relayer/app/cursor/cursors.go
```diff
@@ -0,0 +1,156 @@
...
+	latestOffset.
+		WithLabelValues(c.network.ChainVersionName(srcVersion), destChain.Name).
+		Set(float64(cursor.GetAttestOffset()))
+
+	return nil
+}
+
+// Monitor all existing cursors for the network in the storage and
+// confirm them once the submission is finalized.
+func (c *Cursors) Monitor(ctx context.Context) {
```
corverroos, line 121: lets stick to existing patterns, name this `ProcessForever` and then `processOnce`

relayer/app/cursor/cursors.go
```diff
@@ -0,0 +1,156 @@
...
+
+		if ctx.Err() != nil {
+			return
+		}
+
+		backoff()
+	}
+}
+
+func (c *Cursors) runOnce(ctx context.Context) error {
```
corverroos, line 136: this isn't running once....

I'd suggest following the pattern set in `monitor/xmonitor/indexer::deleteForever`

relayer/app/cursor/metrics.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	confirmedOffset = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "relayer",
+		Subsystem: "cursors",
+		Name:      "confirmed_offset",
+		Help:      "Confirmed cursor offset",
```
corverroos, line 13: Confirmed cursor offset by src chain version and dest chain

relayer/app/cursor/stream.go
```diff
@@ -0,0 +1,168 @@
...
+//   - if a cursor is not empty fetch finalized cursor from the network and confirm only if finalized
+//     message offset is higher to the cursor last message offset
+//   - if a cursor is empty, confirm it if the previous cursor was confirmed
+func (s *StreamCursors) Confirm(ctx context.Context) error {
+	cursors, err := s.getAllCursors(ctx)
+	if err != nil {
+		return err
+	}
+
+	// if we only have empty and not confirmed skip as there's nothing to do
```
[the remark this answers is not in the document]
corverroos: that is correct though?

relayer/app/cursor/stream.go
```diff
@@ -0,0 +1,137 @@
...
+
+// isFinalized checks if the cursor is finalized on the network.
+func isFinalized(ctx context.Context, network netconf.Network, xProvider xchain.Provider, cursor *Cursor) (bool, error) {
+	chain, ok := network.Chain(cursor.GetSrcChainId())
+	if !ok {
+		return false, errors.New("invalid dest chain id [BUG]")
+	}
+
+	lowestOffset := uint64(math.MaxUint64)
+	for _, shard := range chain.Shards {
```
corverroos, line 123: bug: not all shards belong to the cursor.ConfLevel

relayer/app/cursor/stream.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	return cursors, nil
+}
+
+// ConfirmStream checks all stream cursors and confirms them according to the rules:
+//   - if a cursor is not empty fetch finalized cursor from the network and confirm only if finalized
+//     message offset is higher to the cursor last message offset
+//   - if a cursor is empty, confirm it if the previous cursor was confirmed
+func ConfirmStream(ctx context.Context, db CursorTable, network netconf.Network, xProvider xchain.Provider, stream stream) error {
```
corverroos, line 51: nit: don't export types and fields by default, rather make them private by default

relayer/app/cursor/stream.go
```diff
@@ -0,0 +1,137 @@
...
+	"context"
+	"math"
+	"slices"
+
+	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/netconf"
+	"github.com/omni-network/omni/lib/xchain"
+)
+
+type stream struct {
```
corverroos, line 13: godoc pls

relayer/app/cursor/stream.go
```diff
@@ -0,0 +1,137 @@
...
+	SrcVersion xchain.ChainVersion
+	DestChain  netconf.Chain
+	Network    netconf.Network
```
corverroos, line 16: Network is a complex thing, not sure it should be part of "stream". Also, it isn't used...

relayer/app/cursor/cursors.go
```diff
@@ -0,0 +1,156 @@
...
+	xProvider xchain.Provider,
+	confirmInterval time.Duration,
+	network netconf.Network,
+) (*Cursors, error) {
+	cursorsTable, err := NewCursorsTable(db)
+	if err != nil {
+		return nil, err
+	}
+
+	var streams []stream
```
corverroos, line 37: nit: cursors.streams is only used in one place, in `Monitor`, but `Monitor` has access to all the data to create it. So I'd suggest to remove it from `cursors` and just create them in `Monitor`

Issue #2431, opened 2024-11-08
require that validator pubkeys are on the secp256k1 curve


ttarsi:
#### Problem to Solve

We currently require that a key is 33 bytes but don't verify that it's on the secp256k1 curve. 

#### Proposed Solution

Require that it's on the curve, either during `deliverCreateValidator` in `evmstaking` or in the staking contract.

Zodomo:
Reopening until Staking contracts are updated

PR #2433, opened 2024-11-08
feat(contracts/core): added secp256k1 pubkey validation

Zodomo:
Properly check if a validator's secp256k1 compressed pubkey is on the curve, instead of simply relying on a byte length check. Bindings were modified due to the introduction of witnet's [elliptic-curve-solidity](https://github.com/witnet/elliptic-curve-solidity) library.

issue: #2431


kevinhalliday (commented):
Maybe toss these all into a  `Secp256k1` library?  then just `Secp256k1.validatePubkey(.,.)` from staking

Don't love those constants taking up realestate in the staking contract

kevinhalliday (approved):

Issue #2430, opened 2024-11-08
validate consensus key during createValidator


ttarsi:
#### Problem to Solve

Validator pubkey is unchecked during registration. This is fine for now since the validator set is whitelisted and validation happens on the cosmos side, but this is required for permissionless validators.

#### Proposed Solution

require and verify consensus private key signature field in `Staking.sol::createValidator`. Might want to change to uncompressed pubkey for simplicity of verification

PR #2452, opened 2024-11-11
feat(contracts/core): refactored to verify secp256k1 signatures

Zodomo:
Implemented secp256k1 signature verification in Staking.sol's createValidator function.

issue: #2430


Zodomo:
In case it is relevant, here is some basic typescript I used to generate keys and signatures with to test against

https://gist.github.com/Zodomo/1b0227ed536c7bc6830a181c44f446c9

contracts/core/src/libraries/Secp256k1.sol
```diff
@@ -45,4 +76,75 @@ library Secp256k1 {
...
+            r := calldataload(signature.offset)
+            s := calldataload(add(signature.offset, 32))
+            v := byte(0, calldataload(add(signature.offset, 64)))
+        }
+
+        // v needs to be 0 or 1 for secp256k1 (not 27/28 like Ethereum)
+        require(v == 0 || v == 1, "Secp256k1: invalid v value");
+
+        // Add 27 to make it compatible with ecrecover
+        v += 27;
```
kevinhalliday, line 138: rather than translating here, i suggest we require one (probably 27 / 28) impose that requirement on signers (our lib/k1util::Sign already produces signatures with v 27/28)

that said I think we should just use a oz's escda library for this. It does all these things and others (like keeping s value to lower bound, to prevent malleable signatures).

best to default to battle tested / audited code when we can, particularly for crypography related stuff

kevinhalliday (commented):
Couple things

I think we should be using [eip 712](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md) signatures. We don't exactly need typedata. But we do need domain separation. So best to just follow common standard

When doing cryptography related operations, I suggest we use battetested code (oz's ecdsa library, in this case). Even tho it is fun to write these things ourselves, it's easy to miss things. And reduces audit surface area

contracts/core/src/octane/Staking.sol
```diff
@@ -82,16 +89,47 @@ contract Staking is OwnableUpgradeable {
...
+
+        emit CreateValidator(msg.sender, pubkey, msg.value);
+    }
+
+    /**
+     * @param x The x coordinate of the validators consensus public key
+     * @param y The y coordinate of the validators consensus public key
+     * @return Digest hash to be signed by the validators public key
+     */
+    function getValidatorDigest(bytes32 x, bytes32 y) external view returns (bytes32) {
```
kevinhalliday, line 115: maybe `getValidatorPubkeyDigest()`

contracts/core/src/libraries/Secp256k1.sol
```diff
@@ -45,4 +71,23 @@ library Secp256k1 {
...
         return EllipticCurve.isOnCurve(x, y, AA, BB, PP);
     }
+
+    /**
+     * @notice Convert public key coordinates to an Ethereum address
+     * @param x The x coordinate of the public key
+     * @param y The y coordinate of the public key
+     * @return The Ethereum address corresponding to the public key
+     */
+    function publicKeyToAddress(bytes32 x, bytes32 y) internal pure returns (address) {
```
kevinhalliday, line 81: maybe `pubkeyToAddress` cause we just use "pubkey" elsewhere?

kevinhalliday (dismissed):

kevinhalliday (commented):
Need to handle fresh initialization for devnet / staging. We only call initialize in AllocPredeploys

kevinhalliday (approved):

PR #2458, opened 2024-11-12
chore(contracts/core): fix uninitialized destination

Zodomo:
Upgrade bridge native dry-run was failing as destination chainId wasn't initialized yet.

issue: #565


contracts/core/script/admin/BridgeNativePostUpgradeTest.sol
```diff
@@ -38,7 +38,7 @@ contract BridgeNativePostUpgradeTest is Test {
     function _setup() internal {
         b = OmniBridgeNative(Predeploys.OmniBridgeNative);
         l1Bridge = b.l1Bridge();
-        l1ChainId = b.l1ChainId();
+        l1ChainId = b.l1ChainId() == 0 ? 1 : b.l1ChainId();
```
kevinhalliday, line 41: We'll likely need to do same thing for l1BridgeAddr too. But I would change to be more explicit right now. Rather than saying "if unset, do this". Let's say "if omni mainnet, set to these values"

```solidity
l1Bridge = b.l1Bridge();
l1ChainId = b.l1ChainId();

// manually setting for eth mainnet, to support upgrade of bridge pre "setup"
if (block.chainid == 166) {
   l1ChainId = 1
   l1Bridge = 0xbbb3f5bcb1c8b0ee932efaba2fdee566b83053a5
} 
```

kevinhalliday (approved):

PR #2465, opened 2024-11-12
chore(multi): use staking reinitializer and add upgrade tests

Zodomo:
The Staking.sol contract was updated, so the `updatedStaking` routine was updated to use the new initializer function, and additional tests were added.

issue: #2430


contracts/core/script/admin/Admin.s.sol
```diff
@@ -234,13 +235,23 @@ contract Admin is Script {
...
         address impl = address(new Staking());
         vm.stopBroadcast();
 
         _upgradeProxy(admin, Predeploys.Staking, impl, data);
 
-        // TODO: add post upgrade tests
+        // assert storage unchanged
+        require(staking.owner() == owner, "owner changed");
+        require(staking.isAllowlistEnabled() == isAllowlistEnabled, "isAllowlistEnabled changed");
+
```
kevinhalliday, line 253: Let's test that initializers are disabled on implementation (and fix if they are not)

Zodomo: I believe that check is performed globally for all Admin script actions [here](https://github.com/omni-network/omni/blob/8298b12d3104826aa85bba2ac4db87c5d32ec250/contracts/core/script/admin/Admin.s.sol#L388)

kevinhalliday (approved):

Issue #2461, opened 2024-11-12
Handle EngineAPI application errors


corverroos:
#### Problem to Solve
EngineAPI spec doesn't specify that `status==Invalid` MUST not also return an error. Geth actually returns both. Our retryForever logic doesn't distinghusih between the two.  

#### Proposed Solution
Check for `status==Invalid` BEFORE retrying all other errors.

PR #2469, opened 2024-11-13
fix(octane/evmengine): handle engine errors

corverroos:
Improve EngineAPI error handling:
 - Geth returns `status:INVALID` AND errors, so swallow errors if known status is returned.
 - Remaining errors (with unknown status) should only be temporary networking (or some unexpected geth error).
 - Timeout PrepareProposal after 10s, proposing an empty consensus block rather (other validators probably already moved on in any case.)
 - Timeout ProcessProposal after 1min, prevent blocking forever

issue: #2461

### DIFF
--- a/halo/app/prouter.go
+++ b/halo/app/prouter.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"context"
+	"time"
 
 	atypes "github.com/omni-network/omni/halo/attest/types"
 	"github.com/omni-network/omni/lib/errors"
@@ -15,6 +16,11 @@ import (
 	"github.com/cosmos/cosmos-sdk/client"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
+
+// processTimeout is the maximum time to process a proposal.
+// Timeout results in rejecting the proposal, which could negatively affect liveness.
+// But it avoids blocking forever, which also negatively affects liveness.
+const processTimeout = time.Minute
 
 // makeProcessProposalRouter creates a new process proposal router that only routes
 // expected messages to expected modules.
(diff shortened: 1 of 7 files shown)


halo/app/prouter.go
```diff
@@ -32,6 +38,11 @@ func makeProcessProposalRouter(app *App) *baseapp.MsgServiceRouter {
 // It also updates some external state.
 func makeProcessProposalHandler(router *baseapp.MsgServiceRouter, txConfig client.TxConfig) sdk.ProcessProposalHandler {
 	return func(ctx sdk.Context, req *abci.RequestProcessProposal) (*abci.ResponseProcessProposal, error) {
+		// Only allow 10s to process a proposal. Reject proposal otherwise.
```
sideninja, line 41: good idea

lib/ethclient/engineclient.go
```diff
@@ -77,50 +67,39 @@ func NewAuthClient(ctx context.Context, urlAddr string, jwtSecret []byte) (Engin
...
-	var resp engine.ForkChoiceResponse
-	err := c.cl.Client().CallContext(ctx, &resp, forkchoiceUpdatedV2, update, payloadAttributes)
-	if err != nil {
-		incError(c.chain, endpoint)
-		return engine.ForkChoiceResponse{}, errors.Wrap(err, "rpc forkchoice updated v2")
-	}
+	incError(c.chain, endpoint)
 
-	return resp, nil
+	return engine.PayloadStatusV1{}, errors.New("nil error and unknown status", "status", resp.Status)
```
sideninja, line 102: should this be a [BUG]?

sideninja (approved):

lib/ethclient/engineclient.go
```diff
@@ -129,30 +108,33 @@ func (c engineClient) ForkchoiceUpdatedV3(ctx context.Context, update engine.For
 	const endpoint = "forkchoice_updated_v3"
 	defer latency(c.chain, endpoint)()
 
+	// isStatusOk returns true if the response status is valid.
+	isStatusOk := func(resp engine.ForkChoiceResponse) bool {
```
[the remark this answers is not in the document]
corverroos: functions are easy to read and document. Also slightly more concise
```
if isStatusOk(resp) {

vs

if okStatuses[resp.PayloadStatus.Status] {
```

lib/ethclient/engineclient.go
corverroos: possibly yeah

Issue #2206, opened 2024-10-17
Deploy gas pump contracts on mainnet


fabtreb:
- Deploy [GasPump](https://github.com/omni-network/omni/blob/main/contracts/core/src/interfaces/IOmniGasPump.sol) contract on mainnet
- Document usage of the gas pumps in smart contracts
- Configure the [gas pump ui](https://bridge.omni.network/refuel) to work with the deployed contracts

Zodomo:
All contracts deployed and verified!

Gas Station [Implementation](https://omniscan.network/address/0xE49661b0a55542E20a39b2205a9883462A1725a4/contract/166/code) and [Proxy](https://omniscan.network/address/0x5A19272100949f9e8A4aD9D98f8C1070232de626/contract/166/code)

Gas Pumps

- Ethereum
    - Implementation [[ES]](https://etherscan.io/address/0xa2F831b93d1bA5eBf17803EF827189ed44efE51e#code) [[BS]](https://eth.blockscout.com/address/0xa2F831b93d1bA5eBf17803EF827189ed44efE51e?tab=contract) [[Sourcify]](https://repo.sourcify.dev/contracts/full_match/1/0xa2F831b93d1bA5eBf17803EF827189ed44efE51e/)
    - Proxy [[ES]](https://etherscan.io/address/0x3cdF44dC0548fDD9304A4020624A5e216D8A861F#code) [[BS]](https://eth.blockscout.com/address/0x3cdF44dC0548fDD9304A4020624A5e216D8A861F?tab=contract) [[Sourcify]](https://repo.sourcify.dev/contracts/full_match/1/0x3cdF44dC0548fDD9304A4020624A5e216D8A861F/)
- Arbitrum
    - Implementation [[ES]](https://arbiscan.io/address/0xa8faf21f43ee3ec1a3f98e8ccb595e32acf8d260#code) [[BS]](https://arbitrum.blockscout.com/address/0xa8faF21F43ee3Ec1a3f98E8cCB595E32Acf8d260?tab=contract) [[Sourcify]](https://repo.sourcify.dev/contracts/full_match/42161/0xa8faF21F43ee3Ec1a3f98E8cCB595E32Acf8d260/)
[...]

PR #2474, opened 2024-11-13
chore(e2e/app): disable fundGasStation on mainnet

Zodomo:
Separate funding gas station action from deployment on mainnet only.

issue: #2206

### DIFF
--- a/e2e/app/gaspump.go
+++ b/e2e/app/gaspump.go
@@ -12,6 +12,7 @@ import (
 	"github.com/omni-network/omni/lib/contracts/gasstation"
 	"github.com/omni-network/omni/lib/errors"
 	"github.com/omni-network/omni/lib/log"
+	"github.com/omni-network/omni/lib/netconf"
 	"github.com/omni-network/omni/lib/txmgr"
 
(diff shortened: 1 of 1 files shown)


e2e/app/gaspump.go
```diff
@@ -30,8 +31,10 @@ func DeployGasApp(ctx context.Context, def Definition) error {
 		return errors.Wrap(err, "deploy gas station")
 	}
 
-	if err := fundGasStation(ctx, def); err != nil {
-		return errors.Wrap(err, "fund gas station")
+	if def.Testnet.Network != netconf.Mainnet {
+		if err := fundGasStation(ctx, def); err != nil {
+			return errors.Wrap(err, "fund gas station")
+		}
```
kevinhalliday, line 37: Maybe `if def.Testnet.Nework.IsEphemeral()`

kevinhalliday (approved):

e2e/app/gaspump.go
Zodomo: That will disable it for Omega too.

PR #2478, opened 2024-11-14
feat(solver): check target before accept

kevinhalliday:
Only accept if know target, and call / deposits are verified.

issue: none


solver/types/target.go
```diff
@@ -15,11 +15,23 @@ type Target interface {
...
 	// Verify returns an error if the call should not be fulfilled.
-	Verify(srcChainID uint64, call *bindings.SolveCall, deposits []*bindings.SolveDeposit) error
+	Verify(srcChainID uint64, call bindings.SolveCall, deposits []bindings.SolveDeposit) error
+}
+
+type Targets []Target
+
+func (t Targets) ForCall(call bindings.SolveCall) (Target, bool) {
+	for _, target := range t {
+		if target.IsAllowedCall(call) {
```
corverroos, line 31: I think we should error if multiple targets are matched?

corverroos: mmm, should we use the inbox or outbox contract address to identify the target? Picking any target that matches feels a bit loose?

corverroos: non matching calls will also be hard to debug since we can't see what rules are expected and why they are failing

corverroos (approved):

solver/types/target.go
kevinhalliday: > I think we should error if multiple targets are matched?

Yeah would help us catch bugs / bad target config.

> mmm, should we use the inbox or outbox contract address to identify the target? Picking any target that matches feels a bit loose?

We will still be executing through the outbox. And there is a difference between outbox allowing or not allowing a call (which it does). And a target saying "yes this is an allowed call". 

We can also add an outbox check. 

> non matching calls will also be hard to debug since we can't see what rules are expected and why they are failing

Yeah good point we do need to consider debugging.

"is this call one we support?" (`target.IsAllowedCall`)

Here we filter out chains / contracts / functions selectors we don't support. Here I think we can
```
Request for unsupported call   dest_chain="holeskey" target_contract="0x..." target_selector="0x..."
```

then "should I make this supported call?" (`target.Verify`)

Here we can log the error returned by verify

```
Request rejected for target      reason="insufficient deposits"   target="symbiotic" dest_chain="holeskey" ...
```

PR #2485, opened 2024-11-14
chore(monitor/xfeemngr): add gas price buffer updates counter metric

kevinhalliday:
Track number of times gas price buffer is updated per chain.

issue: none


kevinhalliday:
Will use `changes` query on existing metric instead

PR #2489, opened 2024-11-15
ci(e2e): support devnet chaos slashing

corverroos:
Support slash testing in e2e:
- Enable it via `halo_chaos = true` in e2e manifest.
- This results in `--devnet-chaos-test=true` flag in halo.
- Which results in halo injecting deterministic double sign evidence once

The aim is to improve the edge-case testing when slashing occurs.

issue: none


corverroos:
closing in favour of cometbft InjectEvidence

PR #2493, opened 2024-11-15
feat(monitor): add support for monitoring additional contracts

Zodomo:
contracts.ToFund didn't quite work well for monitoring contracts we want to withdraw from instead of adding funding to, so I added contracts.ToWithdraw and incorporated it into the monitor. Initially prepared it for the gas pumps.

issue: none


kevinhalliday (commented):
I like separation. But I do think we can do some combination. What are your thoughts on something like a `Contract` type?

```golang
type Contract struct {
	Name              string
	Address           common.Address
	FundThreshold     *FundThreshold
	WithdrawThreshold *WithdrawThreshold
        // ...
}
```

then we can define them all in one place?

lib/contracts/withdraw.go
```diff
@@ -0,0 +1,52 @@
...
+			Address:     addrs.GasPump,
+			OnlyOmniEVM: false,
+			NotOmniEVM:  true,
+			Thresholds:  WithdrawThresholds{minEther: 10},
+		},
+	}, nil
+}
+
+type WithdrawThresholds struct {
+	minEther float64
```
kevinhalliday, line 41: This would be `maxEther` for withdraw right?

Zodomo: I was reading it like "minimum Ether required for withdrawal" but that perspective works too

lib/contracts/contracts.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil, err
+	}
+
+	return []Contract{
+		{
+			Name:           "gas-station",
+			Address:        addrs.GasStation,
+			OnlyOmniEVM:    true,
+			NotOmniEVM:     false,
+			FundThresholds: &FundThresholds{minEther: 200, targetEther: 1000}, // GasStation funds user GasPump requests, and needs a large OMNI balance.
```
kevinhalliday, line 43: Change thresholds

lib/contracts/contracts.go
```diff
@@ -0,0 +1,124 @@
...
+	FundThresholds     *FundThresholds
+	WithdrawThresholds *WithdrawThresholds
+}
+
+func ToFund(ctx context.Context, network netconf.ID) ([]Contract, error) {
+	// GasStation will not deployed initially on mainnet
+	// TODO: remove this when mainnet GasStation
+	if network == netconf.Mainnet {
+		return []Contract{}, nil
+	}
```
kevinhalliday, line 30: need to remove

lib/contracts/contracts.go
Zodomo: Resolved in #2501

lib/contracts/contracts.go
Zodomo: Resolved in #2501

kevinhalliday (commented):
Let's put the whole contracts list in a `contracts.All()` function, then the other views (`ToFund()`, `ToWithdraw()`) can be filters on top of that

lib/contracts/contracts.go
```diff
@@ -0,0 +1,130 @@
...
+	Name               string
+	Address            common.Address
+	OnlyOmniEVM        bool
+	NotOmniEVM         bool
+	FundThresholds     *FundThresholds
+	WithdrawThresholds *WithdrawThresholds
+}
+
+// All returns all contracts for the given network relevant to the monitor.
+func All(ctx context.Context, network netconf.ID) ([]Contract, error) {
```
kevinhalliday, line 27: I know I suggested the name `All`, but do you think we should rename to `ToMonitor()`? because this is really just a list of contracts to monitor. it does not include all contracts

kevinhalliday (approved):

PR #2498, opened 2024-11-18
feat(octane/evmengine): support blob txs

corverroos:
Adds minimal support for blob txs.

Only include `BlobCommitments` in consensus chain blocks. This is required to calculate blob hashes and provide to `engine.NewPayload`. This adds minimum support for blobs, i.e., that they can be submitted to EVM and included in blocks without problems.

But we don't include blobs themselves (or proofs) in the consensus chains. Neither do we provide an API to query them. This limits consensus chain size by not including blobs. Submitted blobs can therefore not be queried from the consensus chain. This does render BlobTxs useless, but this isn't a problem since we don't have a use-case that require it.

Minimal support for blobs does means that stock standard geth is still supported and that forking geth to disable blob txs isn't required.

issue: https://github.com/omni-network/ops/issues/613


octane/evmengine/keeper/helpers.go
```diff
@@ -41,3 +47,43 @@ func retryForever(ctx context.Context, fn func(ctx context.Context) (bool, error
...
+	sidecar, err := bundle.EthSidecar() // Nil bundle is valid but empty.
+	if err != nil {
+		return nil, err
+	}
+
+	hashes := make([]common.Hash, 0, len(sidecar.Blobs))
+
+	hasher := sha256.New()
+	for i := range sidecar.Blobs {
+		if err := kzg4844.VerifyBlobProof(&sidecar.Blobs[i], sidecar.Commitments[i], sidecar.Proofs[i]); err != nil {
```
corverroos, line 80: Is verifying the whole blob, proof, and commitment required for all validators? Since only `Commitments` are required to calculate `versioned hashes` which are provided to geth via `NewPayload`, other fields (`Blobs and Proofs`) are technically unused. Can they not be removed?

chmllr (approved):

kevinhalliday (approved):
Worth having blob tx in e2e?

octane/evmengine/keeper/helpers.go
kc1116: Did you find the answer to this?

corverroos: Yeah, we only include Commitments for minimal but safe support

corverroos:
@kevinhalliday yeah, I'll add some E2E txs in next PR

PR #2507, opened 2024-11-18
fix(lib/contracts): correct deployment of feeoraclev2

Zodomo:
Generation of FeeParams for L2 Sepolia testnets was failing as Sepolia isn't in the manifests for staging or omega. Default gas prices to 1 gwei for non-mainnet manifests.

issue: #1951


kevinhalliday (commented):
Why do this just for mainnet? Can we just stub a 1 gwei gas price if we don't find the backend for the chain (i.e sepola)

kevinhalliday (approved):

PR #2524, opened 2024-11-20
fix(halo/attest): delete empty attestations on double-sign detection

### LINKED ISSUE #2488: Don't store attestation with no signatures
#### Problem to Solve
When storing votes in the attest module DB, we first create the attestation if it doesn't exist, then we add signatures. If we detect double signing, we log a warn and don't store the signature. The attestations is however created. This can be exploited by malicious validators to spam the DB with invalid data.

#### Proposed Solution
In `addOne`, delete the attestation if it was created but signatures were inserted.

raehat:
- Issue: Attestations were created even if double signing was detected, allowing malicious validators to spam the database with invalid data.
- Cause: Attestation creation was not rolled back if no valid signature was stored due to double sign detection.
- Solution: Added logic to delete the attestation if it was created but no signatures were added, preventing the spam of invalid data in the DB as suggested by @corverroos in https://github.com/omni-network/omni/issues/2488#:~:text=Proposed%20Solution-,In%20addOne%2C%20delete%20the%20attestation%20if%20it%20was%20created%20but%20signatures%20were%20inserted.,-corverroos%20added%20the

issue: fixes #2488

### DIFF
--- a/halo/attest/keeper/keeper.go
+++ b/halo/attest/keeper/keeper.go
@@ -176,7 +176,8 @@ func (k *Keeper) addOne(ctx context.Context, agg *types.AggVote, valSetID uint64
 	// Get existing attestation (by unique key) or insert new one.
 	var attID uint64
 	existing, err := k.attTable.GetByAttestationRoot(ctx, attRoot[:])
-	if ormerrors.IsNotFound(err) {
+	insertAtt := ormerrors.IsNotFound(err)
+	if insertAtt {
 		// Insert new attestation
 		attID, err = k.attTable.InsertReturningId(ctx, &Attestation{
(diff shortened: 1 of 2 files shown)


corverroos:
Hey @raehat, thanks for your PR, this is great!

Sorry, I didn't see this until now. I actually also pushed a PR for this earlier today. I actually prefer my implementation, as that avoid doing additional DB queries, which can be significant if the validator set if large. You are welcome to follow my example and then we can merge your PR and I'll abandon mine.

Let me know what you think

raehat:
Hey @corverroos, I chose the db query approach because I wasn't sure of race conditions. I'll follow your example and push a commit then 👍

corverroos:
Great work @arajasek approved 🚀

PR #2527, opened 2024-11-21
fix(halo/attest): dont insert attestations without sigs

corverroos:
Revert insertion of new attestations without valid signatures. This prevents an attack spamming DB with invalid attestations.

issue: #2488

### DIFF
--- a/halo/attest/keeper/keeper.go
+++ b/halo/attest/keeper/keeper.go
@@ -176,7 +176,8 @@ func (k *Keeper) addOne(ctx context.Context, agg *types.AggVote, valSetID uint64
 	// Get existing attestation (by unique key) or insert new one.
 	var attID uint64
 	existing, err := k.attTable.GetByAttestationRoot(ctx, attRoot[:])
-	if ormerrors.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


halo/attest/keeper/keeper.go
```diff
@@ -254,6 +256,15 @@ func (k *Keeper) addOne(ctx context.Context, agg *types.AggVote, valSetID uint64
 			)
 		} else if err != nil {
 			return errors.Wrap(err, "insert signature")
+		} else {
+			insertSig = true
+		}
+	}
+
+	if insertAtt && !insertSig {
```
chmllr, line 264: Shouldn't we account for early returns in line 230 or 258?

Moreover, should we only revert the attestations and not all the other inserted signatures before one of them generates an error?

corverroos: In general, errors inside cosmos modules are not supported or expected. Most errors returned by keeper logic will result in the chain stalling (or proposals being rejected). So no need to worry about cleanup or ordering in these cases, as the whole block will basically fail.

corverroos:
Closed in favour of @2524

Issue #2525, opened 2024-11-21
Valsync buffering


chmllr:
Implement #1857.

Corresponding ADR: https://www.notion.so/omni-network/Valsync-Buffering-14469a99643a809989a2e503733373e9

PR #2534, opened 2024-11-21
feat(halo/evmstaking2): adds a new empty module

chmllr:
Add the boilerplate for a new Cosmos SDK module (step 1 of the Valsync Buffering ADR).

issue: #2525


halo/evmstaking2/types/keys.go
```diff
@@ -0,0 +1,6 @@
+package types
+
+const (
+	// ModuleName defines the module name.
+	ModuleName = "evmstaking2"
```
corverroos, line 5: this can be `evmstaking`, then less to change later

kc1116 (approved):

PR #2536, opened 2024-11-21
feat(e2e): allow forking public chains

kevinhalliday:
Allow forking public chains on devnet anvil chains.

This requires using new dev accounts, rather then well known defaults.
Well known anvil dev accounts have been used on public chains.

issue: none

### DIFF
--- a/lib/anvil/accounts.go
+++ b/lib/anvil/accounts.go
@@ -10,29 +10,27 @@ import (
 
 //nolint:gochecknoglobals // Static keys and addresses
 var (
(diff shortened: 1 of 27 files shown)


lib/anvil/accounts.go
```diff
@@ -79,13 +77,9 @@ func DevAccount9() common.Address {
 	return acc9
 }
 
-func DevAccount10() common.Address {
```
kevinhalliday, line 82: Removed account 10 because it was not actually an anvil dev account

kevinhalliday:
will split into two prs

PR #2537, opened 2024-11-21
refactor(contracts/core): improved FeeOracleV2 update efficiency

Zodomo:
Refactored FeeOracleV2 such that data cost fee updates can be made once and update all execution layer pricing for all chains that post to any given data layer. Also, conversion rates are set independently, which updates pricing for all routes that rely on them.

issue: #1951


contracts/core/src/xchain/FeeOracleV2.sol
```diff
@@ -14,30 +14,35 @@ contract FeeOracleV2 is IFeeOracle, IFeeOracleV2, OwnableUpgradeable {
...
      */
-    uint24 public baseGasLimit;
+    address public manager;
 
     /**
-     * @notice Address allowed to set gas prices and to-native conversion rates.
+     * @notice Conversion rate from this chain's gas token to another native token, by gas token ID.
      */
-    address public manager;
+    mapping(uint8 gasToken => uint256) public tokenToNativeRate;
```
kevinhalliday, line 27: Converion rate from `gasToken` to this chain's native token (normalized by CONVERSION_RATE_DENOM)

contracts/core/src/xchain/FeeOracleV2.sol
```diff
@@ -168,62 +265,130 @@ contract FeeOracleV2 is IFeeOracle, IFeeOracleV2, OwnableUpgradeable {
...
+    function _bulkSetToNativeRate(NativeRateParams[] calldata params) internal {
+        for (uint256 i = 0; i < params.length; i++) {
+            NativeRateParams memory n = params[i];
+
+            if (n.gasToken == 0) revert IFeeOracleV2.ZeroGasToken();
+            if (n.nativeRate == 0) revert IFeeOracleV2.ZeroNativeRate();
+
+            tokenToNativeRate[n.gasToken] = n.nativeRate;
+
+            emit ToNativeRateSet(n.gasToken, n.nativeRate);
```
kevinhalliday, line 309: `_setToNativeRate(n.gasToken, n.nativeRate)`?

contracts/core/src/interfaces/IFeeOracleV2.sol
```diff
@@ -9,81 +9,185 @@ import { IConversionRateOracle } from "./IConversionRateOracle.sol";
...
+    }
+
+    /**
+     * @notice Parameters for a gas token's to-native conversion rate.
+     * @custom:field gasToken       The gas token ID.
+     * @custom:field nativeRate     The to-native conversion rate for that gas token.
+     */
+    struct NativeRateParams {
+        uint8 gasToken;
+        uint256 nativeRate;
```
kevinhalliday, line 111:
```
struct ToNativeRateParams {
    uint8 gasToken;
    uint256 rate;
}
```

feel like we lose semantic meaning when we drop to prefix and just say nativeRate

lib/contracts/feeoraclev2/config.go
```diff
@@ -0,0 +1,66 @@
...
+const defaultBaseDataBuffer = 100
+const defaultGasPerByte = 16
+
+const (
+	// Mainnets.
+	IDEthereum uint64 = 1
+
+	// Testnets.
+	IDSepolia uint64 = 11155111
+)
```
kevinhalliday, line 19: Are these used?

lib/contracts/feeoraclev2/config.go
```diff
@@ -0,0 +1,66 @@
...
+
+type Config struct {
+	ID             uint64
+	BaseGasLimit   uint32
+	BaseDataBuffer uint32
+	GasPerByte     uint64
+}
```
kevinhalliday, line 26: We should ascribe more meaning this name here. And maybe not expose it publicly. When thinking of exposed names think of import path. This would be `feeoraclev2.Config`. But this is data cost config, not fee oracle config.

Though most structs & functions we do not need to expose yet. Best to default to keeping them private to module until somethings needed

lib/contracts/feeoraclev2/config.go
```diff
@@ -0,0 +1,66 @@
+package feeoraclev2
+
+import (
+	"sort"
+
+	"github.com/omni-network/omni/lib/evmchain"
+)
+
+const defaultBaseGasLimit = 100_000
+const defaultBaseDataBuffer = 100
```
kevinhalliday, line 10: Mention units in name or comment

contracts/core/src/interfaces/IFeeOracleV2.sol
```diff
@@ -9,81 +9,185 @@ import { IConversionRateOracle } from "./IConversionRateOracle.sol";
...
+     * @notice Data cost parameters for a data cost ID.
+     * @custom:field gasToken       The gas token ID.
+     * @custom:field baseDataBuffer The base data buffer for that data cost ID.
+     * @custom:field dataCostId     The data cost ID.
+     * @custom:field gasPrice       The data gas price for that data cost ID (denominated in chains native token).
+     * @custom:field gasPerByte     The gas per byte for that data cost ID.
+     */
+    struct DataCostParams {
+        uint8 gasToken;
+        uint32 baseDataBuffer;
```
kevinhalliday, line 98: We should mention units in name here.

IMO "buffer" and "base" are trying getting at the same thing here (ie. overhead). So we can drop one of them. And "data' is implied by namespace of the struct.

`baseBytes` or `bufferBytes`

contracts/core/src/interfaces/IFeeOracleV2.sol
```diff
@@ -9,81 +9,185 @@ import { IConversionRateOracle } from "./IConversionRateOracle.sol";
...
+        uint64 dataCostId;
```
kevinhalliday, line 99: Same thing here. I think we can drop `dataCostId` prefix because that namespaced by the struct typename.

Could just say `id`

lib/contracts/feeoraclev2/config.go
```diff
@@ -0,0 +1,66 @@
...
+	for _, cfg := range static {
+		resp = append(resp, cfg)
+	}
+
+	sort.Slice(resp, func(i, j int) bool {
+		return resp[i].ID < resp[j].ID
+	})
+
+	return resp
+}
```
kevinhalliday, line 64: If we don't need yet we don't need to write it

lib/contracts/feeoraclev2/config.go
```diff
@@ -0,0 +1,66 @@
...
+		if metadata.PostsTo == 0 {
+			config.BaseDataBuffer = defaultBaseDataBuffer
+			config.GasPerByte = defaultGasPerByte
+		}
+
+		configs[metadata.ChainID] = config
+	}
+
+	return configs
+}
```
kevinhalliday, line 46: I feel like we should keep data cost params and fee params separate in static config, as we do in the contract.

contracts/core/src/interfaces/IFeeOracleV2.sol
```diff
@@ -9,81 +9,185 @@ import { IConversionRateOracle } from "./IConversionRateOracle.sol";
...
+    /// @notice Returns the gas per byte for a data cost ID.
+    function dataGasPerByte(uint64 dataCostId) external view returns (uint64);
+
+    /// @notice Returns the to-native conversion rate for a destination chain.
+    function toNativeRate(uint64 chainId) external view returns (uint256);
 
-    /// @notice Returns the base gas limit.
-    function baseGasLimit() external view returns (uint24);
+    /// @notice Returns the to-native conversion rate for a data cost ID.
+    function toNativeRateData(uint64 dataCostId) external view returns (uint256);
```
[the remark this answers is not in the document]
Zodomo: Tbh it isn't necessary. Someone could call `dataCostParams()` to get a dataCostId's `gasToken`, and then call `tokenToNativeRate()` to get the value.

contracts/core/src/interfaces/IFeeOracleV2.sol
Zodomo: Settled for baseBytes

lib/contracts/feeoraclev2/config.go
Zodomo: Removed for refactored static mappings

lib/contracts/feeoraclev2/config.go
Zodomo: Refactoring these to feeConfig and dataCostConfig. Making them private as they're only needed in the fee oracle for now.

kevinhalliday (approved):
👌 👏

Issue #2533, opened 2024-11-21
Strict execution payload encoding


corverroos:
#### Problem to Solve
Currently, octane encodes the execution payload obtained from engineAPI into the `MsgExecutionPayload.ExecutionPayload` field as json bytes. 

Json is not strict, in the sense that arbitrary data can be added to json struct which will be ignored when unmarshalled. This provides an attack vector for malicious validators to inject arbitrary large data (up to max block size) into consensus chain blocks. Which could lead to range of operational and performance issues. 

#### Proposed Solution
Use a strict serialization format like [ssz](https://github.com/attestantio/go-eth2-client/blob/master/spec/deneb/executionpayload.go) or [protobuf](https://github.com/prysmaticlabs/prysm/blob/develop/proto/engine/v1/execution_engine.proto) which has the added benefit of decreasing block sizes.

Or use [json.Decoder.DisallowUnknownFields](https://pkg.go.dev/encoding/json#Decoder.DisallowUnknownFields). Note this is tricky since geth does custom unmarshalling which bypasses `DisallowUnknownFields`

Add backwards compatibility tests with supported geth versions

corverroos:
Milestone: Permissionless validators

raehat:
is someone working on this issue? If not, is it up for grabs?

corverroos:
Note that cosmosSDK uses `github.com/cosmos/cosmos-sdk/codec/unknownproto.RejectUnknownFieldsStrict` to mitigate against this issue in their proto encodings.

PR #2541, opened 2024-11-22
fix(octane/evmengine): prevent json data injection

corverroos:
Mitigates malicious proposers including unknown fields in execution payload JSON.

Maintain backwards compatibility with previous versions of geth that include `"deposits":null`.

issue: #2533


corverroos:
This doesn't actually work since geth types do custom unmarshalling which bypass DisallowUnknownFields

PR #2543, opened 2024-11-22
feat(halo/evmstaking2): implement EvmEventProcessor for keeper

chmllr:
- rename proto type (singular <> plural)
- implement `EvmEventProcessor` for evmstaking2 keeper
- add basic implementation for `Deliver`

issue: #2525


halo/evmstaking2/keeper/keeper.go
```diff
@@ -3,17 +3,74 @@ package keeper
...
+}
+
+// Deliver processes a omni deposit log event, which must be one of:
+// - CreateValidator
+// - Delegate.
+// Note that the event delivery is not immediate. Instead, every event is
+// first stored in keeper's state. Then all stored events are periodically delivered
+// from `EndBlock` at once.
+func (k Keeper) Deliver(ctx context.Context, _ common.Hash, elog evmenginetypes.EVMEvent) error {
+	err := k.eventsTable.Insert(ctx, &EVMEvent{
```
corverroos, line 68: tests as we go would also help ensure everything is working as expected

halo/evmstaking2/keeper/evmstaking2.proto
```diff
@@ -7,8 +7,8 @@ import "octane/evmengine/types/tx.proto";
 
 option go_package = "halo/evmstaking2/keeper";
```
corverroos, line 8: I'd rename the proto file to `evmstaking.proto` since then we don't need to rename later. The name might be included in some underlying identifiers.

corverroos (approved):

kc1116 (approved):

Issue #2490, opened 2024-11-15
Test slashing in e2e


corverroos:
#### Problem to Solve
We currently don't test slashing in E2E tests. 

#### Proposed Solution
Use [cometbft/e2e/app.InjectEvidence](https://github.com/cometbft/cometbft/blob/main/test/e2e/runner/evidence.go#L31) in e2e app.

PR #2544, opened 2024-11-22
ci(e2e): add slashing testcase

corverroos:
Add support for `manifest.evidence` that submits a double signing evidence and asserts the validator is slashed.

Also speed up e2e by doing stuff in parralel.

issue: #2490

### DIFF
--- a/e2e/app/evidence.go
+++ b/e2e/app/evidence.go
@@ -1,0 +1,182 @@
+package app
+
+import (
+	"context"
+	"time"
+
+	"github.com/omni-network/omni/e2e/types"
+	"github.com/omni-network/omni/lib/cchain/provider"
+	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/log"
+
+	"github.com/cometbft/cometbft/crypto"
+	"github.com/cometbft/cometbft/crypto/tmhash"
+	"github.com/cometbft/cometbft/evidence"
+	e2e "github.com/cometbft/cometbft/test/e2e/pkg"
+	cmttypes "github.com/cometbft/cometbft/types"
+)
+
+// awaitSlashed returns nil when the provided validator is slashed.
+func awaitSlashed(ctx context.Context, def Definition, valAddr crypto.Address) error {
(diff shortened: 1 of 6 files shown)


e2e/app/run.go
```diff
@@ -83,49 +85,34 @@ func Deploy(ctx context.Context, def Definition, cfg DeployConfig) (*pingpong.XD
...
+	eg2.Go(func() error { return DeployBridge(ctx, def) })
+	eg2.Go(func() error { return maybeSubmitNetworkUpgrade(ctx, def) })
+	eg2.Go(func() error { return FundValidatorsForTesting(ctx, def) })
 	if def.Manifest.DeploySolve {
-		if err := solve.DeployContracts(ctx, NetworkFromDef(def), def.Backends()); err != nil {
-			return nil, errors.Wrap(err, "deploy solve contracts")
-		}
+		eg2.Go(func() error { return solve.DeployContracts(ctx, NetworkFromDef(def), def.Backends()) })
+	}
+	if err := eg2.Wait(); err != nil {
```
chmllr, line 114: Just for my understanding: this refactoring assumes that there is no sequential dependency on these deployments, right? Clarifying because `initPortalRegistry` being the first deployment might indicate that there are some dependencies.

e2e/app/evidence.go
```diff
@@ -0,0 +1,182 @@
...
+
+	return ev, nil
+}
+
+// makeRandomBlockID was copied from cometbft/test/e2e/runner/evidence.go.
+func makeRandomBlockID() cmttypes.BlockID {
+	return makeBlockID(crypto.CRandBytes(tmhash.Size), 100, crypto.CRandBytes(tmhash.Size))
+}
+
+// makeRandomBlockID was copied from cometbft/test/e2e/runner/evidence.go.
```
chmllr, line 166:
```suggestion
// makeBlockID was copied from cometbft/test/e2e/runner/evidence.go.
```

chmllr (approved):

e2e/app/run.go
corverroos: indeed, no dependencies.

PR #2552, opened 2024-11-22
fix(lib/contracts): add one block wait for holesky deployments

Zodomo:
Add 12 second/~1 block wait for Holesky deployments in an effort to diagnose Staging deployment failures on Holesky.

issue: none


lib/ethclient/ethbackend/backend.go
```diff
@@ -177,6 +177,12 @@ func (b *Backend) WaitMined(ctx context.Context, tx *ethtypes.Transaction) (*eth
 		return rec, errors.New("receipt status unsuccessful", "status", rec.Status, "tx", tx.Hash())
 	}
 
+	// If b.chainID == holesky, wait a little longer for the tx to be indexed.
+	if b.chainID == 17000 {
```
kevinhalliday, line 181: `can use lib/evmchain::IDHolesky`

kevinhalliday (approved):

PR #2577, opened 2024-11-26
feat(contracts/avs): enabled operator deregistration

Zodomo:
Enabled operator deregistration in the AVS contracts.

issue: https://github.com/omni-network/ops/issues/601


kevinhalliday (approved):
Let's follow this up with a fork upgrade test?

PR #2591, opened 2024-11-29
feat(halo/evmstaking2): unit tests for unhappy paths

chmllr:
More unit tests for unhappy paths.

issue: #2525


halo/evmstaking2/keeper/keeper_internal_test.go
```diff
@@ -88,18 +88,112 @@ func TestInsertAndDeleteEVMEvents(t *testing.T) {
 	}
 }
 
-func TestHappyPathDelivery(t *testing.T) {
+func TestDeliveryWithBrokenServer(t *testing.T) {
```
kc1116, line 91: Need godocs for all tests

halo/evmstaking2/keeper/keeper_internal_test.go
```diff
@@ -88,18 +88,112 @@ func TestInsertAndDeleteEVMEvents(t *testing.T) {
...
+
+	expectDelegates := 1
+	expectCreates := 1
+	expectTotalEvents := expectDelegates + expectCreates
+
+	require.Len(t, events, expectTotalEvents)
+
+	for _, event := range events {
+		err := keeper.parseAndDeliver(ctx, &event)
+		require.True(t, strings.Contains(err.Error(), sServer.err.Error()))
```
kc1116, line 120: nit: Incorrect syntax usage of **_require_** , update all usages.

```suggestion
	require.Error(t, err)
	require.Contains(t, err.Error(),"create validator: pubkey to cosmos")
```
Using require.Error and require.Contains is preferred as it provides clearer intent, better failure messages, and avoids potential panics if err is nil. It also aligns with idiomatic testing practices for improved readability and debugging.

chmllr: Nice! Thanks!

halo/evmstaking2/keeper/keeper_internal_test.go
```diff
@@ -202,19 +296,21 @@ func setupKeeper(
...
 }
 
-// GetValidator returns no errors on the first call, because it is called on delegation
-// event delivery for the first time and the validator should be available.
-// Second time it is called on a validator creation event and it should return an error
-// on the pubkey of the new validator.
-func (m *stakingKeeperStub) GetValidator(context.Context, sdk.ValAddress) (stypes.Validator, error) {
-	m.calls++
-	if m.calls == 1 {
+// GetValidator memorizes all addresses and returns an error for calls it never seen before.
```
kc1116, line 302:
```suggestion
// GetValidator tracks accessed addresses, returning an error for unseen addresses.
```

halo/evmstaking2/keeper/keeper_internal_test.go
```diff
@@ -202,19 +296,21 @@ func setupKeeper(
...
+func (m *stakingKeeperStub) GetValidator(ctx context.Context, addr sdk.ValAddress) (stypes.Validator, error) {
```
kc1116, line 303:
```suggestion
func (m *stakingKeeperStub) GetValidator(_ context.Context, addr sdk.ValAddress) (stypes.Validator, error) {
```

halo/evmstaking2/keeper/keeper.go
```diff
@@ -240,6 +240,10 @@ func (k Keeper) deliverDelegate(ctx context.Context, ev *bindings.StakingDelegat
 		return errors.New("validator does not exist", "validator", valAddr.String())
 	}
 
+	if ev.Amount == nil {
```
corverroos, line 243: lets add a "verifyStakingDelegate` function. Also, we can rename the `ev` variable, it was copy paste bug.

halo/evmstaking2/keeper/keeper_internal_test.go
corverroos: we don't really do this for any tests in the code base. So not sure we can start a new convention simply in the middle of a PR review?

corverroos (approved):

halo/evmstaking2/keeper/keeper_internal_test.go
```diff
@@ -202,19 +296,21 @@ func setupKeeper(
...
+	}
+
+	hexAddr := string(addr)
+
+	if _, found := m.validators[hexAddr]; found {
 		return stypes.Validator{}, nil
 	}
+	m.validators[hexAddr] = true
 
 	return stypes.Validator{}, errors.New("validator does not exist")
```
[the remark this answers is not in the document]
chmllr: I assume from your description correction that you understood the logic.

halo/evmstaking2/keeper/keeper_internal_test.go
chmllr: I did some quick googling and it does not look like this convention is a thing, in general...

I think comments inside the test doing the "handholding" and explaining the steps are more useful, aren't they? If the godoc would contain all these details instead, you'd still need to do mental work by mapping the docs to the actual code. But if we remove the details, then the test name is already explaining the high level idea, doesn't it?

kc1116 (approved):
LGTM, make sure to add the command to the makefile and move the mocks to a _**mocks**_ dir not testutil.

halo/evmstaking2/keeper/keeper_internal_test.go
kc1116: > we don't really do this for any tests in the code base. So not sure we can start a new convention simply in the middle of a PR review?
I think this is fine starting good standards as we go along, especially in small PR's. 

> I did some quick googling and it does not look like this convention is a thing, in general...
> 
> I think comments inside the test doing the "handholding" and explaining the steps are more useful, aren't they? If the godoc would contain all these details instead, you'd still need to do mental work by mapping the docs to the actual code. But if we remove the details, then the test name is already explaining the high level idea, doesn't it?

Brief godoc that concisely describes the test and expected behavior is very useful. Example [geth](https://github.com/ethereum/go-ethereum/blob/master/core/state_processor_test.go#L43). That way without any context a dev can see what the test does, just looking at the code is not enough. I dont think there is any cost to adding godocs for all tests (atleast the more involved ones) and all exported funcs.

halo/evmstaking2/types/interfaces.go
```diff
@@ -1,3 +1,4 @@
+//go:generate mockgen -source ./interfaces.go -package testutil -destination ../testutil/mock_interfaces.go
```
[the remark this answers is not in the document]
chmllr: I would leave this out here and instead create one PR doing this for all other tests later.

PR #2592, opened 2024-11-29
feat(falo/evmstaking2): integrate evmstaking2 if feature enabled

chmllr:
Adds a new manifest and module wiring.

issue: #2525


halo/app/app_config.go
```diff
@@ -94,113 +100,127 @@ var (
...
+						Config: appconfig.WrapAny(&portalmodule.Module{}),
+					},
+					{
+						Name:   registrytypes.ModuleName,
+						Config: appconfig.WrapAny(&registrymodule.Module{}),
+					},
+				}
+
+				// TODO(christian): integrate into the list above
+				if feature.FlagEVMStakingModule.Enabled(context.Background()) {
```
corverroos, line 207: you cant use context background.

corverroos (commented):
You also need to add it to endblockers

halo/app/app_config.go
```diff
@@ -53,6 +59,12 @@ const (
 
 //nolint:gochecknoglobals // Cosmos-style
 var (
+	stakingLogsDeliveryInterval = map[netconf.ID]int64{
```
corverroos, line 62: nit, more concise is more better: `stakingDeliveryInterval`

halo/app/app_config.go
```diff
@@ -53,6 +59,12 @@ const (
...
+		netconf.Mainnet: 21_600, // roughly ~12h assumping 0.5bps finalization rate
```
corverroos, line 63: lets make this a round number, e.g. 20k blocks

halo/app/app_config.go
```diff
@@ -53,6 +59,12 @@ const (
...
+		netconf.Omega:   21_600,
+		netconf.Devnet:  2,
```
corverroos, line 65: suggest doing protected vs ephemeral rather

halo/app/app_config.go
```diff
@@ -94,113 +107,134 @@ var (
...
+					if !found {
+						deliveryInterval = stakingLogsDeliveryInterval[netconf.Devnet]
+					}
+
+					configs = append(configs, &appv1alpha1.ModuleConfig{
+						Name: evmstaking2types.ModuleName,
+						Config: appconfig.WrapAny(&evmstaking2module.Module{
+							DeliveryInterval: deliveryInterval,
+						}),
+					})
```
corverroos, line 225: lets make this more consice:
```
configs = append(configs, &appv1alpha1.ModuleConfig{
	Name: evmstaking2types.ModuleName,
	Config: appconfig.WrapAny(&evmstaking2module.Module{
		DeliveryInterval: deliveryInterval(network),
	}),
})
```

halo/app/app_config.go
```diff
@@ -53,6 +59,14 @@ const (
 
 //nolint:gochecknoglobals // Cosmos-style
 var (
+	deliveryInterval = func(network netconf.ID) int64 {
+		if network.IsProtected() {
+			return 20_000 // roughly ~12h assumping 0.5bps finalization rate
+		}
+
+		return 2
+	}
```
kc1116, line 68: Move this to a normal func declaration it doesn't have to be a variable, remove the magic numbers. 

```
var (
    protectedDeliveryInterval = 20_000 // roughly ~12h assumping 0.5bps
    ephermeralDeliveryInterval = 2
    ....
)

func deliveryInterval(network netconf.ID) int64 {
	if network.IsProtected() {
		return protectedDeliveryInterval
	}

	return ephermeralDeliveryInterval
}
```

kc1116 (approved):
LGTM

halo/app/app.go
```diff
@@ -78,6 +81,7 @@ type App struct {
 	EVMEngKeeper          *evmengkeeper.Keeper
 	AttestKeeper          *attestkeeper.Keeper
 	ValSyncKeeper         *valsynckeeper.Keeper
+	StakingEventKeeper    *evmstaking2.Keeper
```
corverroos, line 84: Suggest naimng it `EVMStakingKeeper` to align with `evmstaking2.Keeper` and with other module names

halo/app/app.go
```diff
@@ -136,9 +142,19 @@ func newApp(
...
 		&app.UpgradeKeeper,
 		&app.SlashingEventProc,
-		&app.StakingEventProc,
 		&app.UpgradeEventProc,
-	); err != nil {
+	}
+
+	// TODO (christian): remove feature check with logs
+	if feature.FlagEVMStakingModule.Enabled(ctx) {
+		log.Info(ctx, "EVMStakingModule feature ENABLED")
```
corverroos, line 150: We already log which feature flags are enabled in `feature.WithFlags`, so can remove these logs as they are duplicate.

halo/app/app.go
```diff
@@ -136,9 +142,19 @@ func newApp(
...
+		dependencies = append(dependencies, &app.StakingEventKeeper)
+	} else {
+		log.Info(ctx, "EVMStakingModule feature DISABLED")
```
corverroos, line 153: No need to log if flags are disabled, this is default behaviour

halo/app/app_config.go
corverroos: agreed. But rather use the same prefix for similar consts/vars to group them together.
```
deliveryIntervalProtected = 20_000 // roughly ~12h assumping 0.5bps 
deliveryIntervalEphemeral = 2 // Fast updates while testing
```

halo/app/app_config.go
```diff
@@ -78,6 +92,7 @@ var (
 		attesttypes.ModuleName,
 		valsynctypes.ModuleName, // Wraps staking module end blocker (must come after attest module)
 		upgradetypes.ModuleName,
+		evmstaking2types.ModuleName,
```
corverroos, line 95: you shouldn't add this modile if feature is disabled. Also, the ordering isn't correct

halo/evmstaking2/keeper/keeper.go
```diff
@@ -162,6 +165,9 @@ func (k Keeper) Deliver(ctx context.Context, _ common.Hash, elog evmenginetypes.
 		return errors.Wrap(err, "insert evm event")
 	}
 
+	log.Debug(ctx, "Scheduled a new staking event")
```
corverroos, line 168: best log this where you can add more context, event type, key fields, etc. This doesn't provide enough information, it only logs that something happened, not what actually happened, it will not help when debugging

halo/evmstaking2/keeper/metrics.go
```diff
@@ -0,0 +1,29 @@
+package keeper
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	eventDeliveryHeight = promauto.NewGauge(prometheus.GaugeOpts{
```
corverroos, line 9: not sure about this metric. What is the "insight" here? What question is this answering. We already now that, events are processed every 20k blocks, this doesn't add any additional information that we don't already know.

halo/evmstaking2/keeper/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+		Namespace: "halo",
+		Subsystem: "evmstaking",
+		Name:      "event_delivery_height",
+		Help:      "The height at which all scheduled staking events were delivered",
+	})
+
+	scheduledEvents = promauto.NewGauge(prometheus.GaugeOpts{
+		Namespace: "halo",
+		Subsystem: "evmstaking",
+		Name:      "pending_events",
```
corverroos, line 19: `buffered_events` maybe?

halo/app/app_config.go
corverroos: Also maybe add a note the const block in `app_config`:
```
// Note that these constants form part of consensus logic and can only be changed during network upgrades.
const (
   ...
)
```

halo/app/app_config.go
corverroos: remember to run normal tests to ensure they still work (regression)

halo/evmstaking2/keeper/keeper.go
```diff
@@ -31,15 +31,15 @@ var (
...
-	deliverInterval int64
+	eventsTable      EVMEventTable
+	ethCl            ethclient.Client
+	address          common.Address
+	contract         *bindings.Staking
+	aKeeper          types.AuthKeeper
+	bKeeper          types.BankKeeper
+	sKeeper          types.StakingKeeper
+	sServer          types.StakingMsgServer
+	deliveryInterval int64
```
corverroos, line 42: I would avoid longer names `deliverInterval` is fine and simple and shorter.Shorter names FTW!

halo/app/app.go
chmllr: Yep that was only for my debugging to make sure we're going though this exact code path. Will remove.

halo/evmstaking2/keeper/keeper.go
chmllr: Not sure this is easily possible as we indeed have no additional information ad this point: the event is and stays unparsed until we start processing it, right?

halo/evmstaking2/keeper/metrics.go
chmllr: Wouldn't it be useful to have a dashboard with the delivery events? I will change this to a counter, which is probably more natural.

halo/evmstaking2/keeper/keeper.go
corverroos: then lets remove the log as it doesn't add sufficient value

halo/evmstaking2/keeper/metrics.go
```diff
@@ -0,0 +1,29 @@
+package keeper
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	eventDeliveries = promauto.NewGauge(prometheus.GaugeOpts{
```
corverroos, line 9: this should be a counter I think. Also see [prometheus](https://prometheus.io/docs/practices/naming/) docs on naming (counters should have `total` suffix)

halo/evmstaking2/keeper/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+	})
+
+	bufferedEvents = promauto.NewGauge(prometheus.GaugeOpts{
+		Namespace: "halo",
+		Subsystem: "evmstaking",
+		Name:      "buffered_events",
+		Help:      "The number of buffered staking events to be delivered",
+	})
+
+	failedEvents = promauto.NewGauge(prometheus.GaugeOpts{
```
corverroos, line 23: this should be a counter too

corverroos (approved):

PR #2599, opened 2024-12-02
feat(sdk): init sdk package with biome and a precommit hook

ga-reth:
initialise the SDK package with:

- SDK subdir
- typescript dependency
- compilation / clean steps
- gitignore
- typed abi file
- top level readme

issue: none


sdk/src/utils/package.json
```diff
@@ -0,0 +1,3 @@
+{
+    "type": "module"
+}
```
kevinhalliday, line 3: Why do we need these (this and src/types/package.json)?

sdk/pnpm-workspace.yaml
```diff
@@ -0,0 +1,2 @@
+packages:
+  - 'src'
```
kevinhalliday, line 2: I'm good with using a workspace but like we said live, might be best just to start with single package until we have need for multiple

kevinhalliday (approved):
Couple thoughts in comments.

Looks like we've got pre-commit hooks failing. Fix those. And then not would be a good time to introduce some ts precommit / ci hooks. typechecking, linting, formatting, tests

sdk/src/utils/package.json
ga-reth: for the pnpm workspaces setup, the presence of each `package.json` indicates a new submodule within the workspace

sdk/pnpm-workspace.yaml
ga-reth: sounds good, i'll revert to a more simple setup

ga-reth:
hey @kevinhalliday i've dropped the workspaces stuff, and added compilation / clean steps, i'll add precommit stuff in a followup

ga-reth:
@kevinhalliday i've added biome for linting/formatting and a precommit hook

PR #2600, opened 2024-12-02
feat(contracts/misc): add pledge deploy script

Zodomo:
Added foundry deploy script for Pledge contract

issue: none

### DIFF
--- a/contracts/misc/script/pledge/Pledge.s.sol
+++ b/contracts/misc/script/pledge/Pledge.s.sol
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity =0.8.24;
+
+import { Pledge } from "../../src/pledge/Pledge.sol";
+import { Script, console2 } from "forge-std/Script.sol";
+import { ICreateX } from "createx/src/ICreateX.sol";
+
+contract PledgeScript is Script {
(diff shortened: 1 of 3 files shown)


contracts/misc/script/pledge/Pledge.s.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity =0.8.24;
+
+import { Pledge } from "../../src/pledge/Pledge.sol";
+import { Script, console2 } from "forge-std/Script.sol";
+import { ICreateX } from "createx/src/ICreateX.sol";
+
+contract PledgeScript is Script {
+    ICreateX public createX = ICreateX(0xba5Ed099633D3B313e4D5F7bdc1305d3c28ba5Ed);
+    address public omni = 0x5e9A8Aa213C912Bf54C86bf64aDB8ed6A79C04d1;
```
kevinhalliday, line 10: Worth tossing our portal addresses in a library?

kevinhalliday (approved):

contracts/misc/script/pledge/Pledge.s.sol
Zodomo: Not my style personally when it comes to scripts. I like references within scripts to remain static, as they are run at a set point of time. If a script needs to be run on a recurring basis on the most recent address, that would make sense, but this script is a one-off. I like preserving the context in which it was run, especially as the deployment address here would change if the portal address did.

PR #2601, opened 2024-12-02
fix(cli): add flag for rpc server URL

kc1116:
Currently, the CLI relies on static consensus URLs. This morning, we disabled one of these URLs, which caused the _operator init-nodes_ command to fail because the static URL was no longer reachable. As a result, the instructions in our live documentation ([Run a Full Node - Quick Start](https://docs.omni.network/operate/run-full-node#quick-start)) are now broken. This PR introduces the ability for users to override the static consensus URL.

```
omni operator init-nodes --network=omega --moniker=foo --rpc-server=https://omega.omni.network --home=./temp
```

issue: none

### DIFF
--- a/cli/cmd/flags.go
+++ b/cli/cmd/flags.go
@@ -30,7 +30,7 @@ func bindInitConfig(cmd *cobra.Command, cfg *InitConfig) {
 	netconf.BindFlag(cmd.Flags(), &cfg.Network)
 	cmd.Flags().StringVar(&cfg.Moniker, "moniker", "", "Human-readable node name used in p2p networking")
 	cmd.Flags().StringVar(&cfg.Home, "home", "", "Home directory. If empty, defaults to: $HOME/.omni/<network>/")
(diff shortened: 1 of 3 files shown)


cli/cmd/flags.go
```diff
@@ -30,6 +30,7 @@ func bindInitConfig(cmd *cobra.Command, cfg *InitConfig) {
 	netconf.BindFlag(cmd.Flags(), &cfg.Network)
 	cmd.Flags().StringVar(&cfg.Moniker, "moniker", "", "Human-readable node name used in p2p networking")
 	cmd.Flags().StringVar(&cfg.Home, "home", "", "Home directory. If empty, defaults to: $HOME/.omni/<network>/")
+	cmd.Flags().StringVar(&cfg.RPCServerURL, "rpc-server", "", "RPC server url to use when querying")
```
kevinhalliday, line 33: maybe `--consensus-rpc`?

kevinhalliday (approved):

corverroos (changes requested):
mmm, not sure this PR is going to help, since we don’t have a public consensus RPC to override the default with…?

PR #2605, opened 2024-12-03
feat(e2e/solve): add symbiotic target

kevinhalliday:
Add symbiotic devnet and staging targets.

issue: none

### DIFF
--- a/contracts/solve/script/MintMockToken.s.sol
+++ b/contracts/solve/script/MintMockToken.s.sol
@@ -1,0 +1,13 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity =0.8.24;
+
+import { MockToken } from "test/utils/MockToken.sol";
+import { Script } from "forge-std/Script.sol";
+
+contract MintMockToken is Script {
+    function run(address token, address to) public {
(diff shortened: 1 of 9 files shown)


contracts/solve/script/MintMockToken.s.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity =0.8.24;
+
+import { MockToken } from "test/utils/MockToken.sol";
+import { Script } from "forge-std/Script.sol";
+
+contract MintMockToken is Script {
+    function run(address token, address to) public {
+        uint256 pk = vm.envUint("PRIVATE_KEY");
```
Zodomo, line 9: Why is there a private key import here? Will other people need to populate this env variable to run the script?

kevinhalliday: so we can run this one public chains .

though now i'm thinking we can just run `forge script --private-key ...`

kevinhalliday: yeah we can just do that

corverroos (approved):

PR #2606, opened 2024-12-03
fix(cli): static peer info 

kc1116:
Currently, the CLI relies on static consensus URLs. This morning, we disabled one of these URLs, which caused the operator init-nodes command to fail because the static URL was no longer reachable. As a result, the instructions in our live documentation ([Run a Full Node - Quick Start](https://docs.omni.network/operate/run-full-node#quick-start)) are now broken. This PR introduces static peer info for persistent networks

issue: none


corverroos (changes requested):
yeah, I don’t think this is going to work. Not only is the list of peers dynamic, but fields like connection_status , channels ,etc and also dynamic and making them static is incorrect

PR #2613, opened 2024-12-03
feat(contracts/portal): refund xcall fee overpayment

RyanSea:
Added a refund on fee overpayment & made a small optimization that removes the second SSLOAD & is imo cleaner

Requesting feedback because there is a needless warning on unused low level return value


Zodomo:
This was a good suggestion. @kevinhalliday and I discussed and we decided that we will refrain from merging it for now. As it stands, integrating apps have to handle when users overpay as it is, so sending refunds to those apps seems to enforce not checking proper payment was made as a norm. Additionally, fee oracle updates are not frequent enough to cause user transactions to revert, so quotes should be accurate between a frontend retrieving one and a user's wallet submitting the transaction.

PR #2615, opened 2024-12-03
(refactor/core): Optimized loops in OmniPortal._clearNetwork()

RyanSea:
Removed subsequent SSLOAD's for length

### DIFF
--- a/contracts/core/src/xchain/OmniPortal.sol
+++ b/contracts/core/src/xchain/OmniPortal.sol
@@ -457,8 +457,9 @@ contract OmniPortal is
      * @notice Clear the network of supported chains & shards
      */
     function _clearNetwork() private {
-        XTypes.Chain storage c;
-        for (uint256 i = 0; i < _network.length; i++) {
+        XTypes.Chain storage c; uint256 totalShards;
+        uint256 totalNetworks = _network.length;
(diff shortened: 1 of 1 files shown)


contracts/core/src/xchain/OmniPortal.sol
```diff
@@ -457,8 +457,9 @@ contract OmniPortal is
      * @notice Clear the network of supported chains & shards
      */
     function _clearNetwork() private {
-        XTypes.Chain storage c;
-        for (uint256 i = 0; i < _network.length; i++) {
+        XTypes.Chain storage c; uint256 totalShards;
+        uint256 totalNetworks = _network.length;
+        for (uint256 i = 0; i < totalNetworks; i++) {
```
kevinhalliday, line 462: These aren't exaclty "totals" (network is a list of chains, each chain has a list of shards) so I'd suggest a rename.

Either literal

```
uint256 lenNetwork;
uint256 lenShards;
```

Or something shorthand / contexual

```
uint256 nChains;
uint256 nShards;
```

kevinhalliday (commented):
Nice - thanks 🙏 One comment on naming

kevinhalliday:
Closing for inactivity. And this function is not called frequently. We can gas optimize if / when it is

PR #2616, opened 2024-12-03
(feat/core): added OmniPortal.updateValidatorSet() SysCall

RyanSea:
Not sure if this is necessary!

The external + internal pattern isn't required but felt right considering other SysCalls. Safe cast should be considered for production


kevinhalliday:
This one isn't necessary but I appreciate the initiative. Here's the context

- syscalls are xmsgs produced by our consensus chain (halo)
- `addValidatorSet` xmsgs are downstream of comet bft validator set changes
- any change in a validator set produces a new validator set. i.e. if a validator's power changes, a new validator set is created (with valsetid = prev valset id + 1). so we only ever need to add validator sets to the portal

PR #2644, opened 2024-12-05
draft(AltVM): explore replacing addrs with bytes32

Zodomo:
This PR explores the impact radius of changing addresses in XMsg to bytes32 values. This change is critical for alt-VM support, as many blockchains (Aptos, Solana, etc) use 32-byte addresses. These changes are substantial, and include: AVS, bridges, XApp templates (including upgradeable storage slots), portals, gas pump/station, and admin actions.

This only encompasses changes to the Solidity smart contracts. These type changes extend far into the Go codebase as well. Everything that uses `xcall` and `xsubmit` basically.

I also introduced a helper library that translates addresses <> bytes32 values, and its use is demonstrated throughout the codebase.

issue: none


Zodomo:
Closing this as it was an exploratory branch. These changes will not be applied today.

PR #2650, opened 2024-12-06
feat(e2e): added avs impl deployer

Zodomo:
Added a deploy routine for the new AVS implementation. Automatic proxy upgrade is not included as the AVS proxy is managed by a manual multisig.

issue: https://github.com/omni-network/ops/issues/601


Zodomo:
Closing because we are going to manually deploy instead

PR #2653, opened 2024-12-08
Update evmstaking.go

orexcoder:
Update some code reusing
Refactor Prepare method

issue: none or #12345


corverroos:
Hey @orexcoder thanks so much for the PR, we really appreciate it. But the `halo/evmstaking` has been deprecated and will be removed soon, so we don't want to do any more changes to it.

So, you have infinite loop bug in your code 😉

PR #2665, opened 2024-12-10
chore(solver/app): instrument inbox processed offset

corverroos:
Add metric tracking latest req offset processed by status. 

issue: none


ga-reth (approved):
Approving to unblock as it's staging, but my go knowledge is nil fyi in case you'd like another to review also

PR #2673, opened 2024-12-10
chore(halo/evmstaking): return error on unexpected code path

chmllr:
Error on old code paths if a feature flag is set.

issue: #2525


corverroos (approved):

corverroos (approved):

lib/feature/feature.go
```diff
@@ -11,14 +11,16 @@ const (
 	FlagEVMStakingModule Flag = "evm-staking-module"
 )
 
+var enabledFlags = make(map[Flag]bool)
```
corverroos, line 14: this will maybe fail when run with -race. Well need to wrap it in an atomic or use a lock possibly

corverroos (approved):

lib/feature/feature.go
corverroos: also maybe add a comment why we use globals here. Saying something about cometbft that doesn't expose the its root context, and the cosmos doesn't actually use the context provided in abci, so using context inside modules isn't possible.

chmllr: SG!

PR #2681, opened 2024-12-11
feat(contracts/core): support user stake delegation

Zodomo:
Added _initial_ support for user stake delegation to the `Staking` contract, and updated related Go. I also added support for delegating on behalf of other users. The omni CLI delegate command was only updated for the new delegate function interface, so CLI still only supports self-delegation, for now.

It has yet to be determined what contract changes, if any, are necessary for unstaking. This contract should not be upgraded yet until the entire scope is known.

issue: #567


contracts/core/src/octane/Staking.sol
```diff
@@ -150,18 +150,13 @@ contract Staking is OwnableUpgradeable, EIP712Upgradeable {
 
     /**
      * @notice Increase your validators self delegation.
-     *         NOTE: Only self delegations to existing validators are currently supported.
-     *         If msg.sender is not a validator, the delegation will be lost.
      * @dev Proxies x/staking.MsgDelegate
      */
-    function delegate(address validator) external payable {
+    function delegate(address delegator, address validator) external payable {
```
corverroos, line 155: I'd suggest rather creating a new method, since the signature is changing. This means that bindings will not be compatbile with our onchain contracts, which isn't good. Main brnach code should remain backwards compatible with onchain deployed contracts. Once they have been upgraded, then only can we remove methods that we dont need.

corverroos: we should have called the method `selfDelegate`, then we could simply have added `delegate` now... 🤦

kevinhalliday: we don't need to change the signature.

This should be keep `delegate(address validator)`, we just have to remove the check below that enforces self delegation.

We do not want anyone to be able to choose delegations for anyone else

kevinhalliday (changes requested):

contracts/core/src/octane/Staking.sol
Zodomo: Rolled the function back to its original implementation, minus the self-delegation check. Implemented a new delegateFor function. Changes are reflected in the Go codebase as well, restoring those as they were before.

kevinhalliday (approved):
Consolidate to `_delegate`?

PR #2687, opened 2024-12-12
chore(solver): fix solver typo

corverroos:
Fixes typo in solver.

Also makes cli `delegate` command public so it can be used in scripts.

issue: none


solver/app/targets.go
```diff
@@ -11,7 +11,7 @@ import (
 
 var targetsByNetwork = map[netconf.ID][]types.Target{
 	netconf.Devnet:  {devapp.MustGetApp(netconf.Devnet), symbiotic.MustGetApp(netconf.Devnet)},
-	netconf.Staging: {devapp.MustGetApp(netconf.Devnet), symbiotic.MustGetApp(netconf.Staging)},
```
corverroos, line 14: typo

PR #2692, opened 2024-12-13
feat(e2e): periodic xmsgs part 1

kc1116:
This PR adds the skeleton app that will generate xmgs on a specified network.

issue: #2275

### DIFF
--- a/e2e/loadgen/app/app.go
+++ b/e2e/loadgen/app/app.go
@@ -1,0 +1,15 @@
+package app
+
+import (
+	"context"
+
+	"github.com/omni-network/omni/lib/log"
+)
+
+func Run(ctx context.Context, cfg Config) error {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	log.Info(ctx, "Loadgen config:", cfg.String())
+	// @TODO: add load generation logic
+	return nil
+}
--- a/e2e/loadgen/app/config.go
+++ b/e2e/loadgen/app/config.go
@@ -1,0 +1,71 @@
+package app
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/omni-network/omni/e2e/app/eoa"
+	"github.com/omni-network/omni/lib/evmchain"
+	"github.com/omni-network/omni/lib/netconf"
+)
+
+type Config struct {
+	Count        int
+	Network      string
+	Role         string
+	ChainIDPairs string
+}
+
+type ChainIDPair struct {
+	From string
+	To   string
+}
+
+func (c Config) String() string {
+	var s strings.Builder
+	_, _ = s.WriteString(fmt.Sprintf("count:%d\n", c.Count))
+	_, _ = s.WriteString(fmt.Sprintf("metwork:%s\n", c.Network))
+	_, _ = s.WriteString(fmt.Sprintf("role:%s\n", c.Role))
+	_, _ = s.WriteString("Chain ID XCall Pairs from->to:\n")
+	for _, pair := range c.GetChainIDPairs() {
+		_, _ = s.WriteString(fmt.Sprintf(" %s -> %s\n", pair.From, pair.To))
+	}
+
+	return s.String()
+}
+
+func (c Config) GetChainIDPairs() []ChainIDPair {
+	pairs := make([]ChainIDPair, 0)
+	for _, pair := range strings.Split(c.ChainIDPairs, ",") {
+		chainIDs := strings.Split(pair, ":")
+		pairs = append(pairs, ChainIDPair{
+			From: chainIDs[0],
+			To:   chainIDs[1],
+		})
+	}
+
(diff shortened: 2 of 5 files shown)


e2e/loadgen/app/app.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"context"
+
+	"github.com/omni-network/omni/lib/log"
+)
+
+func Run(ctx context.Context, cfg Config) error {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	log.Info(ctx, "Loadgen config:", cfg.String())
```
corverroos, line 12: no need to log config, it is already done in `cmd.LogFlags`.

e2e/loadgen/app/config.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+type Config struct {
+	Count        int
+	Network      string
+	Role         string
+	ChainIDPairs string
+}
+
+type ChainIDPair struct {
```
corverroos, line 19: It is better to work with ChainIDs, ie uint64s, not chain names, so lets change this to from string to uint64.

e2e/loadgen/app/config.go
```diff
@@ -0,0 +1,71 @@
...
+	"strings"
+
+	"github.com/omni-network/omni/e2e/app/eoa"
+	"github.com/omni-network/omni/lib/evmchain"
+	"github.com/omni-network/omni/lib/netconf"
+)
+
+type Config struct {
+	Count        int
+	Network      string
```
corverroos, line 14: can use proper type, `netconf.ID`, see examples.

e2e/loadgen/app/config.go
```diff
@@ -0,0 +1,71 @@
...
+	Role         string
```
corverroos, line 15: I think we'll always use the XCaller role right? Not sure it has to be configurable. 

Can also use proper type otherwise.

e2e/loadgen/app/config.go
```diff
@@ -0,0 +1,71 @@
...
+	ChainIDPairs string
```
corverroos, line 16: this should probably be `[]ChainIDPair` not string

e2e/loadgen/cmd/cmd.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	return cmd
+}
+
+func bindFlags(flags *pflag.FlagSet, cfg *app.Config) {
+	flags.IntVar(&cfg.Count, "count", cfg.Count, "number of xcalls to make per chain id pair")
+	flags.StringVar(&cfg.Network, "network", cfg.Network, "network to generate load on")
+	flags.StringVar(&cfg.Role, "role", cfg.Role, "role to use to pay for xcalls")
+	flags.StringVar(&cfg.ChainIDPairs, "chainid-pairs", cfg.ChainIDPairs, "comma separated list of pairs of chain ids to make xcalls from:to e.g: ethereum:arbitrum_one,arbitrum_one:optimism,base:arbitrum_one")
```
corverroos, line 45: I'd simplify this. Since it currently requires string parsing of custom data format. Best stick to proper types as for as possible.

What about just:`ChainIDPairs map[uint64]uint6`, mapping source to dest chains. Sure it doesn't allow sending from the same source to multiple dest chains, but this isn't super important. We just want something simple as v1. Then no parsing required.

e2e/loadgen/cmd/cmd.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/omni-network/omni/e2e/loadgen/app"
+	libcmd "github.com/omni-network/omni/lib/cmd"
+	"github.com/omni-network/omni/lib/log"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+// New returns a new root cobra command that runs the xcall load generator.
+func New() *cobra.Command {
```
corverroos, line 13: I'd actually just follow the `TestJoinNetwork` example. Not sure a whole app is required. Lets keep it simple.

lib/evmchain/evmchain.go
```diff
@@ -83,6 +83,54 @@ func All() []Metadata {
 	return resp
 }
 
+// Mainnets returns all mainnet evmchain metadatas ordered by chain ID.
+func Mainnets() []Metadata {
+	var resp []Metadata
+
+	mainnets := []uint64{IDEthereum, IDOmniMainnet, IDArbitrumOne, IDBase, IDOptimism}
```
corverroos, line 90: I'd avoid defaulting to Ethereum mainnet, it is very expensive. I think the test can just pick the chains it wants, no need for this exhausting default values. We don't want to build a super generic loadgen tool, we just want to test something specific, lets keep the scope small and solution direct to the point.

kc1116:
CLOSED will move to integration test

PR #2693, opened 2024-12-13
fix(cli/cmd): add feature flags to the joint-test

chmllr:
Adds feature flags to the join test

issue: #2525

### DIFF
--- a/scripts/join/join_test.go
+++ b/scripts/join/join_test.go
@@ -16,8 +16,11 @@ import (
 	"time"
 
 	clicmd "github.com/omni-network/omni/cli/cmd"
+	"github.com/omni-network/omni/e2e/manifests"
+	"github.com/omni-network/omni/e2e/types"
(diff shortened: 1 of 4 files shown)


scripts/join/join_test.go
```diff
@@ -61,10 +62,11 @@ func TestJoinNetwork(t *testing.T) {
...
 	cfg := clicmd.InitConfig{
-		Network: networkID,
-		Home:    home,
-		Moniker: t.Name(),
-		HaloTag: haloTag,
+		Network:          networkID,
+		Home:             home,
+		Moniker:          t.Name(),
+		HaloTag:          haloTag,
+		HaloFeatureFlags: strings.Split(*featureFlags, ","),
```
corverroos, line 69: I think this test can automatically read the feature flags from the network manifest. No need to specify it via flags.

corverroos (approved):

PR #2694, opened 2024-12-13
chore(solver/app): attempt to fix rejections

corverroos:
Attempt to fix (and test) rejections:
 - Fix error handling bug.
 - Add invalid call test
 - But rejections are still failing with: "tx reverted", not sure why.
 
 Also some general cleanups:
  - Stop dumping warns logs in E2E, bad signal-to-noise.
  - Add stypes.ReqID type for better formatting and conversion to Offset.
  - Few renames
  - Fix IsDeposited check to use non-zero start balance.
  - Add target app to metrics.

issue: none


kevinhalliday (approved):
Approving so we can merge. We can change reject reason to uint8 on solidity in separate pr

Issue #2679, opened 2024-12-11
Improve request context in SolveInbox


Zodomo:
#### Problem to Solve
In order to determine north star metrics like solve latency, we need historical timestamps as a request goes through status updates. We also need to see which request was most recently switched to any given status.

#### Proposed Solution
Add update history to the requests and include a mapping binding each status value to the most recent request to apply it.

PR #2704, opened 2024-12-16
feat(contracts/solve): add status update history

Zodomo:
Added status update history, so the timestamps of status update progression can be retrieved. This is important for metrics involving solver latency.

issue: #2679


kevinhalliday (approved):

kevinhalliday:
Let's add some tests for this

contracts/solve/src/SolveInbox.sol
```diff
@@ -210,6 +226,9 @@ contract SolveInbox is OwnableRoles, ReentrancyGuard, Initializable, XAppBase, I
 
         req.updatedAt = uint40(block.timestamp);
```
kevinhalliday, line 227: `updatedAt` is now redundant, we can just use last element of updateHistory.

Also, thoughts on just `history`?

Zodomo: Agreed. I'll also use `history` instead.

kevinhalliday (approved):

PR #2719, opened 2024-12-19
chore(contracts/solve): refactor order state

kevinhalliday:
Refactor order state types / logic.

- rename "params" -> "state", more reflective of purpose
- removed `SolverNet` prefix to typenames
- return tuple from `getOrder`, instead of forcing name on wrapped struct
- shorten names when reasonable
- introduce upsert (update / insert) pattern

issue: none

### DIFF
--- a/contracts/solve/src/ERC7683/SolverNetInbox.sol
+++ b/contracts/solve/src/ERC7683/SolverNetInbox.sol
@@ -47,7 +47,7 @@ contract SolverNetInbox is OwnableRoles, ReentrancyGuard, Initializable, Deploye
     /**
      * @notice Map order ID to order parameters.
      */
-    mapping(bytes32 id => SolverNetOrderParams) internal _orderParams;
+    mapping(bytes32 id => OrderState) internal _orderState;
 
     /**
      * @notice Map order ID to order history.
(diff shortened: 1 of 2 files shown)


contracts/solve/src/ERC7683/SolverNetInbox.sol
```diff
@@ -80,24 +80,12 @@ contract SolverNetInbox is OwnableRoles, ReentrancyGuard, Initializable, Deploye
...
-     */
-    function getOrderHistory(bytes32 id) external view returns (StatusUpdate[] memory) {
-        return _orderHistory[id];
+    function getOrder(bytes32 id)
+        external
+        view
+        returns (ResolvedCrossChainOrder memory order, OrderState memory state, StatusUpdate[] memory history)
+    {
+        return (_orders[id], _orderState[id], _orderHistory[id]);
     }
```
kevinhalliday, line 89: Struggled thinking of a more appropraite "wrapping" struct name. Everything felt forced. I actually think `SolvertNetOrder` is decent, in that a `SolverNetOrder` wraps a `ResolvedCrossChainOrder`

But didn't love that we us overloading the use of "order". And we just needed this to name a return type. Tuple should suffice, as this is view is mostly for off chain and easy to unpack.

contracts/solve/src/ERC7683/interfaces/ISolverNetInbox.sol
```diff
@@ -158,11 +138,6 @@ interface ISolverNetInbox is IOriginSettler, ISolverNet {
      */
     function validateOrder(OnchainCrossChainOrder calldata order) external view returns (bool);
 
-    /**
-     * @dev Resolve the onchain order.
-     */
-    function resolve(OnchainCrossChainOrder calldata order) external view returns (ResolvedCrossChainOrder memory);
```
kevinhalliday, line 164: Removed this as it was in `ERC7683` intercace, which we inherit

Zodomo (approved):

PR #2725, opened 2024-12-19
chore(contracts/solve): refactor fill and order data

kevinhalliday:
Refactor fill and order data.

- collocate order / fill types
- refactor fill logic 
- rename / organize some errors

issue: none

### DIFF
--- a/contracts/solve/src/ERC7683/SolverNetOutbox.sol
+++ b/contracts/solve/src/ERC7683/SolverNetOutbox.sol
@@ -111,80 +111,82 @@ contract SolverNetOutbox is OwnableRoles, ReentrancyGuard, Initializable, Deploy
         onlyRoles(SOLVER)
         nonReentrant
     {
-        SolverNetIntent memory intent = abi.decode(originData, (SolverNetIntent));
+        FillOriginData memory fillData = abi.decode(originData, (FillOriginData));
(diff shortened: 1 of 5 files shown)


contracts/solve/src/ERC7683/SolverNetOutbox.sol
```diff
@@ -111,80 +111,82 @@ contract SolverNetOutbox is OwnableRoles, ReentrancyGuard, Initializable, Deploy
...
+            gasLimit: MARK_FILLED_GAS_LIMIT
+        });
+        if (msg.value - call.value < fee) revert InsufficientFee();
+
+        // refund any overpayment in native currency
+        uint256 refund = msg.value - call.value - fee;
+        if (refund > 0) msg.sender.safeTransferETH(refund);
 
-        return call.value;
+        emit Filled(orderId, fillHash, msg.sender);
```
kevinhalliday, line 189: kinda co-mingled marking fulfillment w/ enforcing native payment here. could be improve. might be worth just not using a help for this and keeping it flat in `fill()`

Zodomo:
Noticed that in `_openOrder` you increment the ID state before calling `resolve`, which also increments the ID (in memory). `resolve` should be called first.

Zodomo (approved):

PR #2749, opened 2025-01-06
test(contracts/solve): added simple e2e SolverNet test

Zodomo:
Added a simple end-to-end test confirming ERC7683 intents can be processed from opening the order through the solver claiming the deposits.

issue: none


kevinhalliday (approved):
👏

PR #2751, opened 2025-01-07
chore(docs): add cantina report

ttarsi:
add cantina report

issue: none


kevinhalliday (approved):

ga-reth:
@ttarsi Are we okay with this report report being publicly accessible? If we add it to this static file it can be found

chmllr (approved):

Issue #2106, opened 2024-10-09
Simplify Octane EventLogs


corverroos:
#### Problem to Solve
Currently the Octane `MsgExecutionPayload` contains both the new payload/block being proposed/added, as well as the previous block event logs. 

```
type MsgExecutionPayload struct {
	ExecutionPayload  []byte     
	PrevPayloadEvents []*EVMEvent 
}
```

@chmllr indicated that this could actually be simplified. EVMEvents are deterministic given a finalized block. They can simply be queried after `ForkChoiceUpdated` in `FinalizeBlock` and provided to the `EVMEventProcessors`. 

This reduces the size of consensus blocks. It also simplifies the code.
This also avoids the 1 block lag, with execution event logs being applied in the same consensus block.

Note that querying EVM EventLogs is generally only possible for finalized blocks (post `ForkchoiceUpdate`). So this would mean that verifying event logs during `ProcessProposal` wouldn't be possible. But since it is deterministic and event logs cannot be modified or blocked or omitted, they all need to be processed in anycase. Failing if they are not valid for some reason. So this ins't a problem I think.

> Note this would require a `network upgrade`.

#### Proposed Solution
- Remove `PrevPayloadEvents` from `MsgExecutionPayload`.
- Don't generate them during `PrepareProposal` and don't verify them during `ProcessProposal`.
- During `FinalizeBlock` in `EVMEngine` `msgServer.ExecutionPayload`, after calling `ForkchoiceUpdatedV3`
- Fetch the newly finalized block evm events via `s.evmEvents` and pass those to `s.deliverEvents`
- Ensure events are processing in correct order

Also take the opportunity to improve the `EVMEventProcessor` interface:
```
type EvmEventProcessor interface {
	// Name returns the name of the processor.
	Name() string
	// FilterParams returns the addresses and topics to filter events for.
	FilterParams() ([]common.Address, [][]common.Hash)
	// Deliver processes the EVM log event.
	Deliver(ctx context.Context, blockHash common.Hash, log *EVMEvent) error
}
```

> Note that on the network upgrade height itself, we should process both legacy previous block events as well as current block events. Otherwise we will miss events for `upgrade height-1`.

corverroos:
Note that this should fix the [issue](https://github.com/piplabs/story/issues/301) raised by story

PR #2757, opened 2025-01-08
refactor(octane/evmengine): simplify event proc interface

corverroos:
Simplify the `evmengine.EvmEventProcessor` interface to return filter query params, instead of doing fetch logic. This reduces lots of duplication and removes EthCl dependency from event procs

issue: #2106

### DIFF
--- a/halo/registry/keeper/keeper.go
+++ b/halo/registry/keeper/keeper.go
@@ -8,7 +8,6 @@ import (
 	ptypes "github.com/omni-network/omni/halo/portal/types"
 	"github.com/omni-network/omni/halo/registry/types"
 	"github.com/omni-network/omni/lib/errors"
-	"github.com/omni-network/omni/lib/ethclient"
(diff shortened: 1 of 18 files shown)


halo/registry/keeper/keeper.go
```diff
@@ -51,15 +48,14 @@ func NewKeeper(
 	}
 
 	address := common.HexToAddress(predeploys.PortalRegistry)
-	protalReg, err := bindings.NewPortalRegistryFilterer(address, ethCl)
+	protalReg, err := bindings.NewPortalRegistryFilterer(address, nil) // Passing nil backend if safe since only Parse functions are used.
```
chmllr, line 51:
```suggestion
	protalReg, err := bindings.NewPortalRegistryFilterer(address, nil) // Passing nil backend is safe since only Parse functions are used.
```

chmllr (approved):
Nice!

Issue #2768, opened 2025-01-10
Enable permissionless staking


chmllr:
#### Problem to Solve

Currently only self-delegations are possible. Enable delegations for everyone.

#### Proposed Solution

https://www.notion.so/omni-network/Permissionless-Staking-Unstaking-and-Rewards-V1-16169a99643a8035b573dec79eb250ef?d=17569a99643a80a1a457001c2063ce80#17469a99643a80c2b200d42862a8c2d3

PR #2770, opened 2025-01-10
feat(halo): enable self-delegation

chmllr:
Enable permissionless staking.

issue: #2768


halo/evmstaking2/keeper/keeper.go
```diff
@@ -316,10 +315,6 @@ func (k Keeper) deliverCreateValidator(ctx context.Context, createValidator *bin
 }
 
 func verifyStakingDelegate(delegate *bindings.StakingDelegate) error {
-	if delegate.Delegator != delegate.Validator {
```
corverroos, line 319: this must be behind a feature flag, since it changes consensus logic

corverroos: remember to add tests 🙏

lib/ethclient/enginemock.go
```diff
@@ -98,15 +98,20 @@ func WithMockValidatorCreation(pubkey crypto.PubKey) func(*engineMock) {
 	}
 }
 
-// WithMockSelfDelegation returns an option to add a self-delegation Delegate event to the mock.
-func WithMockSelfDelegation(pubkey crypto.PubKey, ether int64) func(*engineMock) {
+// WithMockDelegation returns an option to add a delegation event to the mock from the specified address.
+func WithMockDelegation(validatorPubkey crypto.PubKey, delegatorPukey crypto.PubKey, ether int64) func(*engineMock) {
```
corverroos, line 102: nit: only require what you need. Pubkeys are not required, only addresses.

corverroos (approved):
unit test is great, but we also need a e2e test. Lets add one to `e2e/test`

chmllr:
> unit test is great, but we also need a e2e test. Lets add one to `e2e/test`

As discussed in Slack, the e2e test will be added in a separate PR.

PR #2775, opened 2025-01-10
chore(e2e/manifests): update omega version pins

kevinhalliday:
Update omega version pins to commit ead6061

issue: none


corverroos (approved):

e2e/manifests/omega.toml
```diff
@@ -5,8 +5,8 @@ multi_omni_evms = true
 prometheus = true
 
 pinned_halo_tag = "v0.11.0"
```
corverroos, line 7: you probs also want to include halo pin in this

Issue #2791, opened 2025-01-13
Add executor contract in outbox


kevinhalliday:
By removing the `allowedCalls` requirement in the outbox, we allow users to request any calldata from the solver. This means users can request direct calls to ERC20 contracts. Given that solves maintain outbox allowances, this would mean users could steal solver funds. To address this, we should introduce a new contract deployed along with the outbox that actually executes the call. The solver should not give this contract any allowances.

PR #2794, opened 2025-01-13
feat(contracts/solve): implemented SolverNetExecutor and AddrUtils

Zodomo:
SolverNetExecutor was implemented to migrate call execution outside of the contract solvers approve to manage funds. Because arbitrary calldata will eventually be executed by the Outbox, it is important that it cannot be misused to steal solver funds (by calling transfer functions on tokens solvers have approved). This is done by having the Outbox leverage solver approvals to send tokens to the Executor, such that the Executor is outside of those approvals.

address <> bytes32 conversions were also moved into a separate AddrUtils library.

issue: #2791


contracts/solve/src/ERC7683/SolverNetExecutor.sol
```diff
@@ -0,0 +1,72 @@
...
+        outbox = _outbox;
+    }
+
+    /**
+     * @notice Approves a spender (usually call target) to spend a token.
+     * @dev Called prior to `executeCall` in order to ensure tokens can be spent.
+     */
+    function tokenApproval(address token, address spender, uint256 amount) external onlyOutbox {
+        token.safeApprove(spender, amount);
+    }
```
kevinhalliday, line 35: `tokenApproval` does not feel like an imperative function name telling the executor to do something.

maybe something like
```solidity
executor.approve(...); // my preference
executor.doApproval(...);
executor.approveToken(...);
```

contracts/solve/src/ERC7683/SolverNetExecutor.sol
```diff
@@ -0,0 +1,72 @@
...
+        token.safeTransfer(to, amount);
+    }
+
+    /**
+     * @notice Refunds native currency.
+     * @dev Called after `executeCall` in order to refund any native currency sent back to the executor.
+     */
+    function refundNative(address to) external onlyOutbox {
+        to.safeTransferETH(address(this).balance);
+    }
```
kevinhalliday, line 61: For both of these, it feels like the fact that these are refunds is outside the context of the executor. The executor just knows that it let's the outbox make it transfer native, and transfer / approve tokens. I think we can just match those functions 1-to-1 and let the outbox use with contxt 

```solidity
function transfer(token, to, amount);
function approve(token, spender, amount);
function transferETH(to, amount); // or transferNative
```

contracts/solve/src/ERC7683/SolverNetExecutor.sol
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * @dev Allows target contracts to arbitrarily return native tokens to the executor.
+     */
+    receive() external payable { }
+
+    /**
+     * @dev Allows target contracts to arbitrarily return native tokens to the executor.
+     */
+    fallback() external payable { }
+}
```
kevinhalliday, line 72: nice thinking

contracts/solve/src/ERC7683/SolverNetExecutor.sol
```diff
@@ -0,0 +1,72 @@
...
+     * @dev Called prior to `executeCall` in order to ensure tokens can be spent.
+     */
+    function tokenApproval(address token, address spender, uint256 amount) external onlyOutbox {
+        token.safeApprove(spender, amount);
+    }
+
+    /**
+     * @notice Executes a call.
+     */
+    function executeCall(ISolverNet.Call memory call) external payable onlyOutbox {
```
kevinhalliday, line 40: The fact that it's a call is made clear by args IMO

```solidity
function execute(call);
```

contracts/solve/src/ERC7683/SolverNetExecutor.sol
Zodomo: Went with `approve`

contracts/solve/src/ERC7683/SolverNetExecutor.sol
Zodomo: Went with `transfer` and `transferNative`.

kevinhalliday (approved):

Issue #2790, opened 2025-01-13
Test v2 ERC7683 contracts


kevinhalliday:
ERC7683 contracts need full test suite.

PR #2800, opened 2025-01-14
test(contracts/solve): comprehensive ERC7683 SolverNet tests

Zodomo:
Implemented full test suite of ERC7683 SolverNet contracts with 100% coverage.

issue: #2790


contracts/solve/test/ERC7683/inbox/Accept.t.sol
```diff
@@ -0,0 +1,181 @@
...
+            assertEq(storedOrder.maxSpent[i].amount, resolvedOrder.maxSpent[i].amount, "stored order: max spent amount");
+            assertEq(
+                storedOrder.maxSpent[i].recipient,
+                resolvedOrder.maxSpent[i].recipient,
+                "stored order: max spent recipient"
+            );
+            assertEq(
+                storedOrder.maxSpent[i].chainId, resolvedOrder.maxSpent[i].chainId, "stored order: max spent chain id"
+            );
+        }
```
kevinhalliday, line 74: here we are really just testing that `_resolve` is called by inbox both in `open` (`storedOrder`). and in `resolve` (`resolvedOrder`). but if `_resolve` has a mistake, our tests will not catch it. so these tests are not super useful.. let's have our tests derive expectations from user input.

so in this case, instead of comparing maxSpent of two instances of a resolve order, and we test that the stored resolved order has a maxSpent that matches the call expenses we specified when opening the order

same thing goes for minReceived below, and other assertions in these case

I think the right approach here is to write some util called like `_assertResolved(opened, resolved)`that basically does what our `Resolve.t.sol` tests do, but a bit more dynamic. So, for any user input, is the resolved order what we expect.

then, in `Resolve.t.sol` we should have multiple different tests cases of different order opens.

then in something like `Accept.t.sol`, we can call `_assertResolve()` as a sanity check, but leave the majority of the tests contents about the state transition from `pending -> accept`. that way we can look at the accept test and think mostly about the state transition, rather than order resolution

contracts/solve/src/ERC7683/SolverNetInbox.sol
```diff
@@ -277,10 +284,16 @@ contract SolverNetInbox is OwnableRoles, ReentrancyGuard, Initializable, Deploye
...
 
+        if (call.chainId == 0) revert NoCallChainId();
         if (call.target == bytes32(0)) revert NoCallTarget();
         if (call.data.length == 0) revert NoCallData();
         if (deposits.length == 0) revert NoDeposits();
 
+        for (uint256 i; i < deposits.length; ++i) {
+            Deposit memory deposit = deposits[i];
+            if (deposit.amount == 0) revert NoDepositAmount();
+        }
```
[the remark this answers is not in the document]
Zodomo: Added the duplicate native check to `_parseOrder()`. The checks in `_processDeposits()` need to remain there as well though because `_parseOrder()` must continue to work when msg.value is 0 (such as when calling `resolve()`).

contracts/solve/test/ERC7683/TestBase.sol
```diff
@@ -65,21 +88,56 @@ contract TestBase is Test {
...
+            expenses: expenses
+        });
+
+        ISolverNet.Deposit[] memory deposits = new ISolverNet.Deposit[](1);
+        deposits[0] = ISolverNet.Deposit({ token: address(token1).toBytes32(), amount: rand * 1 ether });
+
+        ISolverNet.OrderData memory orderData = ISolverNet.OrderData({ call: call, deposits: deposits });
+
+        return IERC7683.OnchainCrossChainOrder({
+            fillDeadline: uint32(block.timestamp + 1 minutes),
```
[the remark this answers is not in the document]
Zodomo: Lets fix this in a separate PR.

contracts/solve/test/ERC7683/inbox/Accept.t.sol
Zodomo: Implemented a `assertResolved` helper function and replaced the duplicate logic in the tests to rely on that as a sanity check. This helper checks the original ERC-7683 intent against the stored resolved order, instead of comparing the output of `resolve` against itself.

contracts/solve/src/ERC7683/SolverNetInbox.sol
```diff
@@ -277,10 +284,21 @@ contract SolverNetInbox is OwnableRoles, ReentrancyGuard, Initializable, Deploye
         Call memory call = orderData.call;
         Deposit[] memory deposits = orderData.deposits;
 
+        if (call.chainId == 0) revert NoCallChainId();
         if (call.target == bytes32(0)) revert NoCallTarget();
         if (call.data.length == 0) revert NoCallData();
         if (deposits.length == 0) revert NoDeposits();
 
+        bool nativeDeposite;
```
kevinhalliday, line 292: let's match language to below,`hasNative`

contracts/solve/test/ERC7683/TestBase.sol
```diff
@@ -130,12 +194,72 @@ contract TestBase is Test {
         return keccak256(abi.encode(orderId, originData));
     }
 
-    function addressToBytes32(address a) internal pure returns (bytes32) {
-        return bytes32(uint256(uint160(a)));
-    }
+    function assertResolved(
+        address addr,
```
kevinhalliday, line 198: `address user`? give it some meaning

contracts/solve/test/ERC7683/TestBase.sol
```diff
@@ -130,12 +194,72 @@ contract TestBase is Test {
...
+        ISolverNet.TokenExpense[] memory orderExpenses = orderCall.expenses;
+        ISolverNet.Deposit[] memory orderDeposits = orderData.deposits;
+        ISolverNet.FillOriginData memory fillOriginData =
+            abi.decode(resolvedOrder.fillInstructions[0].originData, (ISolverNet.FillOriginData));
+
+        assertEq(addr, resolvedOrder.user, "assertResolved: user");
+        assertEq(srcChainId, resolvedOrder.originChainId, "assertResolved: origin chain id");
+        assertEq(uint32(block.timestamp), resolvedOrder.openDeadline, "assertResolved: open deadline");
+        assertEq(order.fillDeadline, resolvedOrder.fillDeadline, "assertResolved: fill deadline");
+        assertGe(uint256(inbox.getNextId()), uint256(resolvedOrder.orderId), "assertResolved: order id");
```
kevinhalliday, line 213: this is somewhat state depending, maybe pass in expected order id

contracts/solve/test/ERC7683/TestBase.sol
```diff
@@ -130,12 +194,72 @@ contract TestBase is Test {
...
+
+        assertEq(orderExpenses.length, resolvedOrder.maxSpent.length, "assertResolved: max spent length");
+        assertEq(orderExpenses.length, fillOriginData.call.expenses.length, "assertResolved: call expense length");
```
kevinhalliday, line 216: this is not true if `maxSpent` includes native deposit

contracts/solve/test/ERC7683/TestBase.sol
```diff
@@ -130,12 +194,72 @@ contract TestBase is Test {
...
+            );
+            assertEq(bytes32(0), resolvedOrder.minReceived[i].recipient, "assertResolved: min received recipient");
+            assertEq(srcChainId, resolvedOrder.minReceived[i].chainId, "assertResolved: min received chain id");
+        }
 
-    function bytes32ToAddress(bytes32 b) internal pure returns (address) {
-        return address(uint160(uint256(b)));
+        assertEq(1, resolvedOrder.fillInstructions.length, "assertResolved: fill instructions length");
+        assertEq(
+            destChainId,
```
kevinhalliday, line 249: can we derive this from order data, instead of test fixtures. fine to leave if it's a hassle. just makes this util more flexible.

contracts/solve/test/ERC7683/inbox/Cancel.t.sol
```diff
@@ -0,0 +1,177 @@
...
+            inbox.getLatestOrderIdByStatus(ISolverNetInbox.Status.Pending), expectedOrderId, "latest pending order id"
+        );
+        assertEq(
+            inbox.getLatestOrderIdByStatus(ISolverNetInbox.Status.Rejected), expectedOrderId, "latest rejected order id"
+        );
+
+        // Verify token balances after cancellation (deposits should be returned)
+        assertEq(token1.balanceOf(user), initialUserBalance, "user balance after");
+        assertEq(token1.balanceOf(address(inbox)), initialInboxBalance, "inbox balance after");
+    }
```
kevinhalliday, line 134: Nice. Maybe in separate PR, let's rework `reject` to also refund the user. then this test can be different. we can block cancelling an order that's rejected, because there would be no need for it

contracts/solve/test/ERC7683/inbox/MarkFulfilled.t.sol
```diff
@@ -0,0 +1,196 @@
...
+
+        bytes32 orderId = inbox.getLatestOrderIdByStatus(ISolverNetInbox.Status.Pending);
+
+        // Accept the order as solver
+        vm.prank(solver);
+        inbox.accept(orderId);
+
+        // Mock xcall with wrong fill hash
+        bytes32 wrongFillHash = bytes32(uint256(1)); // Some random hash
+        vm.expectRevert(ISolverNetInbox.WrongCallHash.selector);
```
kevinhalliday, line 158: test case name and error selector should be `FillHah` not `CallHash`

contracts/solve/test/ERC7683/inbox/Resolve.t.sol
```diff
@@ -0,0 +1,29 @@
...
+        assertResolved(user, order, resolvedOrder);
+    }
+
+    function test_resolveOrder_native_deposit_succeeds() public {
+        IERC7683.OnchainCrossChainOrder memory order = randNativeOrder();
+        vm.prank(user);
+        IERC7683.ResolvedCrossChainOrder memory resolvedOrder = inbox.resolve(order);
+        assertResolved(user, order, resolvedOrder);
+    }
+}
```
kevinhalliday, line 29: can be in separate PR. but other things to test

- multi deposit
- multi expense

kevinhalliday (approved):
Nice. Left some comments. Feel free to merge and address in new prs

contracts/solve/test/ERC7683/TestBase.sol
Zodomo: Can't use `user` because we already initialize an address assigned to that. Opted for `userAddr`.

contracts/solve/test/ERC7683/TestBase.sol
Zodomo: Can't use `user` because we already initialize an address variable to that in test setup. Opted for `userAddr`.

contracts/solve/test/ERC7683/TestBase.sol
Zodomo: In `SolverNetInbox.resolve()` we always populate a `FillInstruction` array of length 1, and the entire `FillOriginData` struct is packed inside. Currently, our design doesn't point towards extending this array, so leaving it hardcoded at a length check of 1 is probably best, as it will draw attention to our tests if our utilization of this field changes.

contracts/solve/test/ERC7683/inbox/Resolve.t.sol
Zodomo: Will save this for a separate PR.

contracts/solve/test/ERC7683/TestBase.sol
Zodomo: Yes. now retrieving from `orderCall.chainId`.

contracts/solve/test/ERC7683/TestBase.sol
Zodomo: Good catch. `maxSpent` did not include native payments yet. Corrected that, and fixed the first line to take it into account. However, the second line is correct, as `call.expenses` will only ever contain ERC20 information. The native payment is kept in `call.value`. So `orderExpenses` should always match `call.expenses`.

contracts/solve/test/ERC7683/inbox/Cancel.t.sol
Zodomo: Addressed in #2810

contracts/solve/test/ERC7683/TestBase.sol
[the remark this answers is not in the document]
Zodomo: Addressed on #2812

contracts/solve/test/ERC7683/inbox/Resolve.t.sol
Zodomo: Resolved in #2853

PR #2801, opened 2025-01-15
feat(app/solver): v2 solve by liquidity

kevinhalliday:
Solve v2 ERC6783 orders by managing on-chain liquidity.

Liquidity only checked on ShouldReject, existing commitments not accounted for.
Restricted to hardcoded list of supported tokens.

issue: none

### DIFF
--- a/solver/app/v2/reject.go
+++ b/solver/app/v2/reject.go
@@ -1,9 +1,16 @@
 package appv2
 
 import (
+	"bytes"
 	"context"
 
-	"github.com/omni-network/omni/lib/netconf"
+	"github.com/omni-network/omni/lib/cast"
+	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/ethclient/ethbackend"
+	"github.com/omni-network/omni/lib/log"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
 )
 
 //go:generate stringer -type=rejectReason -trimprefix=reject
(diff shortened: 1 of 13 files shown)


solver/app/v2/types.go
```diff
@@ -23,8 +23,15 @@ type FillOriginData = bindings.ISolverNetFillOriginData
...
-	Resolved   OrderResolved
+	ID                 OrderID
+	FillInstruction    bindings.IERC7683FillInstruction
+	FillOriginData     []byte
+	DestinationSettler common.Address
+	DestinationChainID uint64
+	SourceChainID      uint64
+	MaxSpent           []bindings.IERC7683Output
+	MinReceived        []bindings.IERC7683Output
+
```
kevinhalliday, line 34: Was easier to pull and validate everything we need from resolved when `newOrder`, instead of validating on read

solver/app/v2/tokens.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/omni-network/omni/lib/netconf"
+
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// TODO: consider moving this to lib/tokens
+
+type Token struct {
+	symbol      string
```
corverroos, line 20: fields of simple structs can generally be public, since they are used by things outside the struct. See go style guide: public-is-public-private-is-private

solver/app/v2/tokens.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+type Expense struct {
+	token  Token
+	amount *big.Int
+}
+
+type Tokens []Token
+
+func (t Token) isNative() bool {
```
corverroos, line 34: same here, private functions should not be used by other types (even in same package).

solver/app/v2/tokens.go
```diff
@@ -0,0 +1,110 @@
...
+	return Token{
+		symbol:      "OMNI",
+		name:        "Omni Network",
+		chainID:     netconf.EthereumChainID(network),
+		address:     contracts.TokenAddr(network),
+		coingeckoID: "omni-network",
+	}
+}
+
+func (t Token) balanceOf(
```
corverroos, line 94: avoid OO style coding which links logic to data. Make this a simple function that takes token as parameter along with all other params

solver/app/v2/types.go
```diff
@@ -40,63 +47,82 @@ func (id OrderID) String() string {
...
-		"src_chain_id", o.SourceChainID(),
-		"dst_chain_id", o.DestinationChainID(),
+		"src_chain_id", o.SourceChainID,
+		"dst_chain_id", o.DestinationChainID,
 	}
 
-	fill, err := parseFillOriginData(o.FillInstruction().OriginData)
+	fill, err := parseFillOriginData(o.FillOriginData)
 	if err != nil {
 		log.Warn(ctx, "Failed to parse fill origin data", err, attrs...)
```
corverroos, line 110: logging shouldn't happen in deep inside helper functions. Logging should only happen in main logic loops. Helper functions should return errors rather.

solver/app/v2/reject.go
```diff
@@ -14,12 +21,112 @@ const (
...
+		}
+
+		// check liquidty, reject if insufficient
+		for _, expense := range expenses {
+			bal, err := expense.token.balanceOf(ctx, backend, solverAddr)
+			if err != nil {
+				return returnErr(errors.Wrap(err, "get balance", "token", expense.token.symbol))
+			}
+
+			sampleBalance(chainName(destChainID), expense.token, solverAddr, bal)
```
corverroos, line 112: avoid general monitoring inside logic, since it will not monitor if no orders, rather have periodic sampling of balances regardless of orders

corverroos (approved):

Issue #2803, opened 2025-01-15
Add join network test for a node synced from a backup snapshot


powerslider:
#### Problem to Solve

Testing #2709.

#### Proposed Solution

After finishing the implementation of #2709 we want to extend the support of the [Join Network Nightly Workflow](https://github.com/omni-network/omni/blob/main/.github/workflows/ci-join.yaml) to support testing the scenario of a node synced from a backup snapshot to join a network.

PR #2804, opened 2025-01-15
ci(scripts/join): add join test for node snapshots

powerslider:
- Support testing the scenario of a node synced from a backup snapshot to join a network utilizing the new `--from-latest-snapshot` option of the `init-nodes` command.

issue:  #2803

### DIFF
--- a/.github/workflows/ci-join.yaml
+++ b/.github/workflows/ci-join.yaml
@@ -27,20 +27,39 @@ jobs:
         with:
           go-version: 'stable'
 
-      - name: Run join test
+      - name: Run Join Network Test - Full Sync Node
         run: |
           cd scripts/join
           sudo go test . -v \
             --integration \
             --timeout=0 \
-            --logs_file=docker_logs.txt \
+            --logs_file=docker_logs_full_sync.txt \
             --halo_tag="main" \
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-join.yaml
```diff
@@ -27,20 +27,39 @@ jobs:
...
+      - name: Upload Docker Logs - Full Sync Node
         uses: actions/upload-artifact@v4
         if: always()
         with:
           name: docker-logs
-          path: scripts/join/docker_logs.txt
+          path: scripts/join/docker_logs_full_sync.txt
+          retention-days: 3
+
+      - name: Run Join Network Test - Snapshot Sync Node
```
corverroos, line 48: we also need to run this test periodically, for multiple networks preferrably

corverroos (approved):

.github/workflows/ci-join.yaml
powerslider: Currently we support a manual choice between `omega` and `mainnet` and `mainnet` as the sane default. As far as I understand your idea is no make separate jobs for `omega`, `mainnet` and possibly `staging`?

PR #2808, opened 2025-01-15
feat(docs): added FAQ question for AVS deregistration

Zodomo:
Added a FAQ question providing info about how to deregister from the AVS.

issue: #601

### DIFF
--- a/docs/content/operate/faq.md
+++ b/docs/content/operate/faq.md
@@ -111,12 +111,14 @@ Omni currently has a validator whitelist. The whitelist applies to both native *
 
 ### How can I deregister from the old AVS contract?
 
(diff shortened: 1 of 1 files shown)


ttarsi (approved):
Approving with some change requests. I think we can make this much simpler – with these instructions, they'll need to infer that they have to build from source, copy the commit, check it out, learn the command to build from source, etc. Let's make it simpler.

"To deregister, you need to build the CLI from source using `main`:

```
git clone <repo>
cd <repo>
go install <whatever the command is>
omni operator deregister <path>
```"

Make it as simple as possible. Operators are smart but they're not protocol devs

Issue #2811, opened 2025-01-15
SolverNet: enforce fillDeadline


Zodomo:
#### Problem to Solve
`fillDeadline` is currently not being enforced. All we currently do is prevent it from including a timestamp from the past.

#### Proposed Solution
Send `block.timestamp` along with `markFulfilled` call from Outbox.

PR #2812, opened 2025-01-15
feat(contracts/solve): enforce fillDeadline and allow unaccepted fills

Zodomo:
`fillDeadline` is now enforced by making the Outbox tell the inbox the timestamp at which fulfillment occurred. Also, fulfillment when an order was not accepted will now succeed, as the filler can provide an address to claim as during fulfillment. The aforementioned timestamp is leveraged to prevent potential claim frontrunning by tossing an invalid accept transition if it occurred after fulfillment.

issue: #2811


kevinhalliday (changes requested):
let's not do this now

PR #2825, opened 2025-01-17
chore(cli): improve init-nodes flags

corverroos:
Improve omni cli `--init-nodes` flags:
 - Remove `--feature-flags` since it wasn't actually used
 - Default `--halo-tag` to version of git commit

issue: none

### DIFF
--- a/cli/cmd/initnodes.go
+++ b/cli/cmd/initnodes.go
@@ -43,16 +43,35 @@ const (
 	gethJWTSecretFileName = "jwtsecret"
 )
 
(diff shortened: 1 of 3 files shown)


cli/cmd/initnodes.go
```diff
@@ -123,7 +142,6 @@ func InitNodes(ctx context.Context, cfg InitConfig) error {
 	if err != nil {
 		return err
 	}
-	cfg.HaloFeatureFlags = featureFlags
```
corverroos, line 126: initnodes overwrites flag always... so no point in exposing the flag...

PR #2826, opened 2025-01-17
feat(octane/evmengine): proto encode execution payload

corverroos:
Swap octane evmengine ExecutionPayload encoding from json to protobuf which is strictly decoded by cosmos.

This mitigates arbitrary data injection by proposers. 
It also decreases consensus block size since protobufs serialize smaller than json.

Note this feature is behind the `proto-evm-payload` feature flag and will only be enabled in next network upgrade.

issue: #2533

### DIFF
--- a/octane/evmengine/keeper/abci.go
+++ b/octane/evmengine/keeper/abci.go
@@ -117,9 +117,18 @@ func (k *Keeper) PrepareProposal(ctx sdk.Context, req *abci.RequestPreparePropos
 	}
 
 	// Create execution payload message
-	payloadData, err := json.Marshal(payloadResp.ExecutionPayload)
-	if err != nil {
-		return nil, errors.Wrap(err, "encode")
+	var payloadData []byte
+	var payloadProto *types.ExecutionPayloadDeneb
+	if feature.FlagProtoEVMPayload.Enabled(ctx) {
+		payloadProto, err = types.PayloadToProto(payloadResp.ExecutionPayload)
+		if err != nil {
+			return nil, errors.Wrap(err, "encode")
+		}
+	} else {
+		payloadData, err = json.Marshal(payloadResp.ExecutionPayload)
+		if err != nil {
+			return nil, errors.Wrap(err, "encode")
+		}
 	}
 
 	// Convert blobs bundle.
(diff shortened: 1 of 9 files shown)


octane/evmengine/types/customproto.go
```diff
@@ -0,0 +1,120 @@
...
+
+import (
+	"encoding/json"
+
+	"github.com/omni-network/omni/lib/errors"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// customProtobufType defines the interface custom gogo proto types must implement
```
chmllr, line 11:
```suggestion
// customProtobufType defines the interface custom go proto types must implement
```

octane/evmengine/keeper/abci.go
```diff
@@ -117,9 +117,18 @@ func (k *Keeper) PrepareProposal(ctx sdk.Context, req *abci.RequestPreparePropos
 	}
 
 	// Create execution payload message
-	payloadData, err := json.Marshal(payloadResp.ExecutionPayload)
-	if err != nil {
-		return nil, errors.Wrap(err, "encode")
+	var payloadData []byte
```
powerslider, line 120: Just a suggestion:
```
	var (
		payloadData  []byte
		payloadProto *types.ExecutionPayloadDeneb
	)
```

chmllr (approved):

octane/evmengine/types/customproto.go
powerslider: I think the original is correct, because this is [gogo protobuf](https://github.com/gogo/protobuf). This is a more Go opinionated implementation of protobuf, that cosmos sdk very heavily relies on.

powerslider (approved):

octane/evmengine/types/customproto.go
corverroos: indeed, it is called gogo

octane/evmengine/keeper/abci.go
corverroos: if it was 3 variables, sure, but with only 2, your solution is 4 lines of code, while mine is only 2 #lessismore

PR #2848, opened 2025-01-21
feat(contracts/solve): enforce fillDeadline

Zodomo:
`fillDeadline` previously had no enforcement, allowing solvers to fill stale orders. This blocks fulfillment if the `fillDeadline` has passed.

issue: #2811


kevinhalliday:
Holding on merging this until we can handle in solver. Need to
- check fill deadline before accept / fill
- if it's passed after accepting, we need to reject

Either that, or do not merge this. And continue to operate without fill deadline for v1

Zodomo:
Redid the PR in #2929

PR #2860, opened 2025-01-21
feat(contracts/solve): modify TokenExpense handling for transfers

Zodomo:
Allowing the `spender` field in `TokenExpense` to be `bytes32(0)` enables us to treat that like a transfer edge case so direct transfers can be executed without an unnecessary approval.

issue: https://github.com/omni-network/ops/issues/740


contracts/solve/src/ERC7683/SolverNetOutbox.sol
```diff
@@ -149,7 +149,7 @@ contract SolverNetOutbox is OwnableRoles, ReentrancyGuard, Initializable, Deploy
             uint256 tokenBalance = token.balanceOf(address(_executor));
 
             if (tokenBalance > 0) {
-                _executor.approve(token, expense.spender.toAddress(), 0);
+                if (expense.spender.toAddress() != address(0)) _executor.approve(token, expense.spender.toAddress(), 0);
```
kevinhalliday, line 152:
```suggestion
                if (expense.spender != bytes32(0)) _executor.approve(token, expense.spender.toAddress(), 0);
```

kevinhalliday: just a thought

kevinhalliday (approved):

Issue #2115, opened 2024-10-09
Address remaining informational and low security audit findings by Spearbit


chmllr:
#### Problem to Solve

We have some low-prio security audit findings which should be addressed at some point.

#### Proposed Solution

https://cantina.xyz/code/953c0c22-8e30-47d0-8567-07a3dadbef77/findings/6
https://cantina.xyz/code/953c0c22-8e30-47d0-8567-07a3dadbef77/findings/8
https://cantina.xyz/code/953c0c22-8e30-47d0-8567-07a3dadbef77/findings/38

raehat:
Is this up for grabs?

corverroos:
https://cantina.xyz/code/953c0c22-8e30-47d0-8567-07a3dadbef77/findings/38 fix in #2137

PR #2875, opened 2025-01-23
chore(*): fix typos found during cantina

corverroos:
Fixes typos raised during cantina issue 6 and 8.

issue: #2115

### DIFF
--- a/e2e/app/portalregistry.go
+++ b/e2e/app/portalregistry.go
@@ -252,7 +252,7 @@ func allowStagingValidators(ctx context.Context, def Definition) error {
 
 	contract, err := bindings.NewStaking(common.HexToAddress(predeploys.Staking), backend)
 	if err != nil {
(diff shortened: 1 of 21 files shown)


e2e/app/portalregistry.go
```diff
@@ -252,7 +252,7 @@ func allowStagingValidators(ctx context.Context, def Definition) error {
 
 	contract, err := bindings.NewStaking(common.HexToAddress(predeploys.Staking), backend)
 	if err != nil {
-		return errors.Wrap(err, "new staking")
+		return errors.Wrap(err, "new portal registry")
```
chmllr, line 255: Without a broader context and looking only at the surrounding code, I do not understand why this error message text make sense. The staking contract is not the portal registry, is it?

chmllr (approved):

Issue #2876, opened 2025-01-23
Enable parallel download and extract of node snapshots


powerslider:
#### Problem to Solve
`omega` and `mainnet` at this point have very large node snapshot archives on GCP. Currently we download and extract them sequentially which is a waste of time given that there are no dependencies in that process.

#### Proposed Solution
Parallelize with goroutines in order to improve waiting time almost in half.

PR #2877, opened 2025-01-23
feat(cli/cmd): add parallel snapshot download

powerslider:
- Optimize download and extract of node snapshots for geth and halo by parallelizing the process.

issue:  #2876

### DIFF
--- a/cli/cmd/initnodes.go
+++ b/cli/cmd/initnodes.go
@@ -452,7 +452,11 @@ func maybeDownloadSnapshots(ctx context.Context, cfg InitConfig) error {
 	})
 
 	// Wait for all downloads to complete.
-	return errors.Wrap(g.Wait(), "parallel download snapshots")
+	if err := g.Wait(); err != nil {
+		return errors.Wrap(err, "parallel download snapshots")
+	}
+
+	return nil
 }
 
 func downloadSnapshot(ctx context.Context, network netconf.ID, outputDir string, clientName string) error {
--- a/cli/cmd/initnodes.go
+++ b/cli/cmd/initnodes.go
@@ -144,15 +144,15 @@ func InitNodes(ctx context.Context, cfg InitConfig) error {
 		return err
 	}
 
-	if err = maybeDownloadSnapshots(ctx, cfg); err != nil {
-		return err
+	if err := maybeDownloadSnapshots(ctx, cfg); err != nil {
+		return errors.Wrap(err, "download snapshots")
 	}
 
-	if err = maybeDownloadGenesis(ctx, cfg.Network); err != nil {
+	if err := maybeDownloadGenesis(ctx, cfg.Network); err != nil {
 		return errors.Wrap(err, "download genesis")
 	}
 
-	if err = maybeGethInit(ctx, cfg, filepath.Join(cfg.Home, gethClientName)); err != nil {
+	if err := gethInit(ctx, cfg, filepath.Join(cfg.Home, gethClientName)); err != nil {
 		return errors.Wrap(err, "init geth")
 	}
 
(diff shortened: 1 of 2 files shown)


lib/netconf/network.go
```diff
@@ -12,6 +12,18 @@ func (i ID) IsEphemeral() bool {
 	return i == Simnet || i == Devnet || i == Staging
 }
 
+// NodeSnapshotGethStateScheme returns the supported Geth node snapshot state scheme for the given network.
+func (i ID) NodeSnapshotGethStateScheme() string {
```
corverroos, line 16: Remember, AHA, if only used in one place, just put the func there.

But more important, we take archive node snapshots, they are always hash and never path in all networks.

cli/cmd/initnodes.go
```diff
@@ -442,6 +442,29 @@ func gethInit(ctx context.Context, cfg InitConfig, dir string) error {
 	return nil
 }
 
+func downloadSnapshots(ctx context.Context, cfg InitConfig) error {
```
corverroos, line 445: rename to `maybeDownloadSnapshots` since it doesn't always do this.

Also, avoid nesting, rather do:
```
if !cfg.NodeSnapshot {
  return nil
}

...
```

corverroos (changes requested):

cli/cmd/initnodes.go
```diff
@@ -416,9 +411,14 @@ func gethInit(ctx context.Context, cfg InitConfig, dir string) error {
 	{
 		image := "ethereum/client-go:" + geth.Version
 		stateScheme := "path"
+		if cfg.NodeSnapshot {
```
corverroos, line 414: when restoring node snapshots, we don't need to do geth init, so you shouldn't need to do any of this

lib/netconf/network.go
powerslider: on staging that is not the case

cli/cmd/initnodes.go
powerslider: There is some inconsistency with staging that I am currently investigating, so that is why I put it here, but will be removed.

cli/cmd/initnodes.go
```diff
@@ -143,26 +144,16 @@ func InitNodes(ctx context.Context, cfg InitConfig) error {
...
-
-		if err := downloadSnapshot(ctx, cfg.Network, cfg.Home, haloClientName); err != nil {
-			return err
-		}
+	if err = maybeDownloadSnapshots(ctx, cfg); err != nil {
+		return err
 	}
 
-	if err := maybeDownloadGenesis(ctx, cfg.Network); err != nil {
+	if err = maybeDownloadGenesis(ctx, cfg.Network); err != nil {
```
corverroos, line 151: no need to change existing code like this. try to touch the least amount of code in PR. This is in anycase a subjective thing (err var shadowing isn't a problem), no point in changing to your preferred subjective style, because by that logic, other devs will just change it back again.

cli/cmd/initnodes.go
```diff
@@ -442,6 +435,26 @@ func gethInit(ctx context.Context, cfg InitConfig, dir string) error {
...
+	g.Go(func() error {
+		return downloadSnapshot(ctx, cfg.Network, cfg.Home, gethClientName)
+	})
+
+	g.Go(func() error {
+		return downloadSnapshot(ctx, cfg.Network, cfg.Home, haloClientName)
+	})
+
+	// Wait for all downloads to complete.
+	return errors.Wrap(g.Wait(), "parallel download snapshots")
```
corverroos, line 455: you risk wrapping nil error.... which panics.... 

nit: this should have paniced during happy flow testing....

corverroos (approved):

cli/cmd/initnodes.go
powerslider: Wanted to spare an additional error check and make the linter happy at the same time, but you are right and it still needs the check. I will fix it.

PR #2907, opened 2025-01-24
chore(e2e/solve): add mock token deployments

kevinhalliday:
Add mock solvernet token deployments, to test solver in devnet / staging.

Relies on #2906

issue: none

### DIFF
--- a/e2e/solve/mocktokens.go
+++ b/e2e/solve/mocktokens.go
@@ -1,0 +1,155 @@
+package solve
+
+import (
+	"context"
+
+	"github.com/omni-network/omni/contracts/bindings"
+	"github.com/omni-network/omni/e2e/app/eoa"
+	"github.com/omni-network/omni/lib/contracts"
+	"github.com/omni-network/omni/lib/create3"
+	"github.com/omni-network/omni/lib/errors"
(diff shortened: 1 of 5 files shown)


e2e/solve/mocktokens.go
```diff
@@ -0,0 +1,156 @@
...
+	NetworkID netconf.ID
+}
+
+var mocks = []MockToken{
+	// staging mock wstETH
+	{Token: tokens.WSTETH, ChainID: evmchain.IDBaseSepolia, NetworkID: netconf.Staging},
+
+	// devnet mock wstETH
+	{Token: tokens.WSTETH, ChainID: evmchain.IDMockL2, NetworkID: netconf.Devnet},
+	{Token: tokens.WSTETH, ChainID: evmchain.IDMockL2, NetworkID: netconf.Devnet},
```
Zodomo, line 35: Why is this repeated? Is one supposed to be `IDMockL1`?

Zodomo (approved):

e2e/solve/mocktokens.go
kevinhalliday: yeah should be IDMockL1

kevinhalliday: actually just need mock l2

Issue #2880, opened 2025-01-23
Fully test RLUSD bridge


Zodomo:
#### Problem to Solve
RLUSD bridge contracts need to be comprehensively tested.

#### Proposed Solution
Implement tests proving all of the logic in a focused manner.

PR #2908, opened 2025-01-24
test(contracts/misc): implemented comprehensive test suite

Zodomo:
Added comprehensive testing of all lockbox and bridge logic, including revert cases.

issue: #2880


contracts/misc/test/bridge/bridge/ReceiveToken.t.sol
```diff
@@ -0,0 +1,97 @@
...
+        vm.stopPrank();
+
+        vm.startPrank(user);
+        // `srcBridge` must have `BURNER_ROLE` for `srcWrapper`.
+        vm.expectRevert(
+            abi.encodeWithSelector(
+                IAccessControl.AccessControlUnauthorizedAccount.selector, address(srcBridge), minterRole
+            )
+        );
+        omni.mockXCall(DEST_CHAIN_ID, address(destBridge), address(srcBridge), data, DEFAULT_GAS_LIMIT);
```
kevinhalliday, line 48: So what we're doing here is sending a message to `srcBridge` and then `destBridge` after just revoking roles, checking they revert. Both for same reason.

It took me a little bit to figure out what we were testing. Part of that's on me (lol). But a couple reactions

If we are just testing that a bridge reverts when it doesn't have permission, let's just test one bridge? The fact that there was both `srcBridge` and `destBridge` threw me off a little at first, thinking that each was representing a different test case.

This is also a general reaction for the whole tests. Having `srcBridge` and `destBridge` as main fixture names when we are doing bridge unit tests makes it a little confusing. Let's just have one `bridge` and test it? 

Use key word args, maybe with a comment pointing our the failure case
```solidity
omni.mockXCall({
    sourceChainId: DEST_CHAIN_ID,
    sender: destBridge,
    to: srcBridge, // does not have MINTER_ROLE
    data: data,
    gasLimit: DEFAULT_GAS_LIMIT
}}
```

Key word args always help readabilty in cases like that. Cause other wise you have to rememeber order of positional args. For ex in this case, is it an `xcall` to `destBridge` with `xmsg.sender == srcBridge`? Or the other way around?

contracts/misc/test/bridge/bridge/ReceiveToken.t.sol
```diff
@@ -0,0 +1,97 @@
...
+        bytes memory data = abi.encodeCall(Bridge.receiveToken, (user, INITIAL_USER_BALANCE));
+        bytes32 minterRole = srcWrapper.MINTER_ROLE();
+
+        vm.startPrank(admin);
+        srcWrapper.revokeRole(minterRole, address(srcBridge));
+        destWrapper.revokeRole(minterRole, address(destBridge));
+        vm.stopPrank();
+
+        vm.startPrank(user);
+        // `srcBridge` must have `BURNER_ROLE` for `srcWrapper`.
```
kevinhalliday, line 42: `MINTER_ROLE`?

contracts/misc/test/bridge/bridge/ReceiveToken.t.sol
```diff
@@ -0,0 +1,97 @@
...
+        omni.mockXCall(DEST_CHAIN_ID, unknownSender, address(srcBridge), data, DEFAULT_GAS_LIMIT);
+
+        // Unauthorized direct call
+        vm.expectRevert(abi.encodeWithSelector(IBridge.Unauthorized.selector, SRC_CHAIN_ID, user));
+        vm.prank(user);
+        srcBridge.receiveToken(user, 1);
+
+        // Message delivers if sender is `OmniPortal`
+        omni.mockXCall(DEST_CHAIN_ID, address(destBridge), address(srcBridge), data, DEFAULT_GAS_LIMIT);
+    }
```
kevinhalliday, line 30: If we have a single `_reverts()` test case, let's put all reverts in there. So essentially, take test case below `test_receiveToken_reverts_without_wrapper_minter_role` and put it in here

```solidity
// Bridge needs MINTER_ROLE on token
vm.prank(admin);
vm.removeRole(token.MINTER_ROLE(), bridge);
...
vm.grantRole(token.MINTER_ROLE(), bridge);
```

Or, if you want to keep it separate, add more context to this test cast name.

kevinhalliday (commented):
Started leaving inline comments but will just leave general ones to address then we can go from there

for unit tests, let's just test a contract call `bridge` (rather than `srcBridge`). `srcBridge` makes it feels like there's some "dest" side of the tests that we are missing. You can keep `srcBridge` and `destBridge` distinction for e2e test, it makes sense there.

We use `srcWrapper` and `destWrapper` to reference the token contracts we bridge. Same comment on `src` prefix applies here for unit tests. But also, we switched names in contract to just call this `token` instead of `wrapper`. Let's have our tests do the same

TLDR: run unit tests on `bridge` & `token`, rather than `srcBridge` & `srcWrapper` and sometimes `destBridge` &`destWrapper`

contracts/misc/test/bridge/bridge/ReceiveToken.t.sol
Zodomo: I will use keyword args as I do agree about that making the tests more readable. However, I disagree about only testing one side. I think our tests should adequately reflect what the system looks like onchain, and test edge cases accordingly. Especially when the token handling is different based on whether a lockbox is present.

contracts/misc/test/bridge/bridge/ReceiveToken.t.sol
Zodomo: good catch

contracts/misc/test/bridge/bridge/ReceiveToken.t.sol
Zodomo: Condensed all revert cases to a single test.

contracts/misc/test/bridge/bridge/ReceiveToken.t.sol
Zodomo: Dropped src/dest conventions after our conversation. I was going too far with E2E testing instead of focusing on unit testing. Tried to reduce this as much as possible.

kevinhalliday (approved):

PR #2918, opened 2025-01-27
refactor(e2e): poc extensions

kevinhalliday:
POC e2e extions.

e2e extentions will help us trim down e2e runs to only deploy & test what's required for whatever we're working on.

issue: none

### DIFF
--- a/e2e/app/definition.go
+++ b/e2e/app/definition.go
@@ -15,6 +15,7 @@ import (
 	"github.com/omni-network/omni/e2e/types"
 	"github.com/omni-network/omni/e2e/vmcompose"
 	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/ethclient"
 	"github.com/omni-network/omni/lib/ethclient/ethbackend"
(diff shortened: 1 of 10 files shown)


e2e/app/definition.go
```diff
@@ -188,7 +194,105 @@ func newBackends(ctx context.Context, cfg DefinitionConfig, testnet types.Testne
...
+		if err != nil {
+			return ethbackend.Backends{}, errors.Wrap(err, "new public Backend")
+		} else if err := backend.EnsureSynced(ctx); err != nil {
+			return ethbackend.Backends{}, errors.Wrap(err, "ensure public chain synced", "chain", chain.Chain().Name)
+		}
+
+		inner[chain.Chain().ChainID] = backend
+	}
+
+	return ethbackend.BackendsFrom(inner), nil
```
kevinhalliday, line 295: Moved these from `lib/ethclient/ethbackend` to here to resolve import cycles (types in e2e/types relying on `ethbackend.Backends`)

They were only used here so would make sense

Issue #2899, opened 2025-01-24
Deploy RLUSD bridge to devnet


Zodomo:
#### Problem to Solve
RLUSD needs to be deployed to devnet for e2e testing.

#### Proposed Solution
Add a deployer to e2e.

PR #2919, opened 2025-01-27
feat(e2e/xbridge): add RLUSD xbridge deployer

Zodomo:
Added a deployment routine for the RLUSD XBridge contracts. Everything was intentionally isolated to `e2e/xbridge` as much as possible, as opposed to using `lib/contracts`. Due to this decision, some conventions used here (such as the shared logic functions seen in `e2e/xbridge/common.go`) were required by the linter. 

issue: #2899


e2e/app/xbridge.go
```diff
@@ -0,0 +1,141 @@
...
+
+import (
+	"context"
+
+	"github.com/omni-network/omni/e2e/xbridge"
+	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/log"
+)
+
+func DeployEphemeralXBridge(ctx context.Context, def Definition) error {
```
kevinhalliday, line 11: can we put this in `e2e/xbridge`? relates to "e2e extension" convo in slack

lib/contracts/address.go
```diff
@@ -258,6 +274,22 @@ func feeOracleV2(network netconf.ID, version string) common.Address {
...
+}
+
+func lockbox(network netconf.ID, version string) common.Address {
+	return create3.Address(Create3Factory(network), lockboxSalt(network, version), eoa.MustAddress(network, eoa.RoleDeployer))
+}
+
+func bridge(network netconf.ID, version string) common.Address {
+	return create3.Address(Create3Factory(network), bridgeSalt(network, version), eoa.MustAddress(network, eoa.RoleDeployer))
+}
+
```
kevinhalliday, line 292: Let's not put these here. we can put this in `e2e/xbridge`. can do similar golden tests there if you'd like

kevinhalliday (commented):
comments on code location. can you address those first and then I'll do another review?

e2e/app/xbridge.go
Zodomo: This triggers circular imports, as `e2e/xbridge` will need to inherit `e2e/app`, which will inherit `e2e/xbridge` for the deployment function.

lib/contracts/address.go
Zodomo: Migrated salt and address generation to `e2e/xbridge`.

e2e/app/xbridge.go
Zodomo: Resolved this by using the method employed in `e2e/solve`.

e2e/xbridge/token.go
```diff
@@ -0,0 +1,181 @@
...
+	Name       string
+	Symbol     string
+	Minter     common.Address
+	Admin      common.Address
+	Upgrader   common.Address
+	Pauser     common.Address
+	Clawbacker common.Address
+}
+
+func (cfg tokenDeploymentConfig) validateTokenConfig() error {
```
kevinhalliday, line 29:
```suggestion
func (cfg tokenDeploymentConfig) validate() error {
```

the receive `(cfg tokenDepoloymentConfig)` provides enough context

e2e/xbridge/token.go
```diff
@@ -0,0 +1,181 @@
...
+	return contracts.Create3Salt(ctx, network, deployment.symbol)
+}
+
+// deployTokenIfNeeded deploys a new rlusd contract if it is not already deployed.
+// If the contract is already deployed, the receipt is nil.
+func DeployTokenIfNeeded(ctx context.Context, network netconf.ID, backend *ethbackend.Backend, wrapper bool, deployment xBridgeDeployment) (common.Address, *ethtypes.Receipt, error) {
+	targetDeployment := deployment.token
+	if wrapper {
+		targetDeployment = deployment.wrapped
+	}
```
kevinhalliday, line 74: Instead of passing a flag `wrapped` to modify it's behavior. Consider passing in target deployment explicitly

```golang
DeployIfNeededI(ctx, network, backend, deployment.token)
// or
DeployIfNeededI(ctx, network, backend, deployment.wrapped)
```

I see these args are passed down a number of times so might not be as easy as that.

kevinhalliday (approved):

PR #2922, opened 2025-01-28
feat(solver/app): require deposits match quote for expenses

kevinhalliday:
Require deposits match expenses 1:1.

- introduce concept of a quote. will be useful in api (e.g GET /quote)
- only support single expense / deposit, require tokens match (by symbol)

Additionally, refector reject.go to use `RejectOrErr`. Makes it easier to split `shouldReject` into utils.

issue: #2920

### DIFF
--- a/solver/app/v2/quote.go
+++ b/solver/app/v2/quote.go
@@ -1,0 +1,93 @@
+package appv2
+
+import (
+	"math/big"
+
+	"github.com/omni-network/omni/lib/errors"
+)
+
+// getQuote returns payment in `depositTkns` required to pay for `expenses`.
+//
+// For now, this is a simple quote that requires a single expense, paid
+// for by an equal amount of an equivalent deposit token. Token equivalence is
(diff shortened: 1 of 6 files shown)


solver/app/v2/quote.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil, RejectOrErr{
+			Reason: rejectInvalidExpense,
+			Err:    errors.New("only single expense supported"),
+		}
+	}
+
+	expense := expenses[0]
+	depositTkn := depositTkns[0]
+
+	if expense.Token.Symbol != depositTkn.Symbol {
```
kevinhalliday, line 32: will add a check here that makes sure a testnet token cannot pay for mainnet token (like holesky ETH for mainnet ETH)

technically this wouldn't be able to happen because the networks would not overlap, so solver would reject for "unsupported dest chain".

but probably still good to check here

corverroos (approved):

PR #2925, opened 2025-01-28
feat(cli): add edit validator to cli

corverroos:
Adds the following command the cli: `omni operator edit-validator`.

Also adds a e2e test for this.

Switches solidity to signed ints to use -1 as "do not modify". This is required since the previous approach of setting the existing value failed, since CommissionRate can only be updated very 24 hours (even to same value). So we have to identify "nulls".

issue: #2886


contracts/core/src/octane/Staking.sol
```diff
@@ -192,8 +193,15 @@ contract Staking is OwnableUpgradeable, EIP712Upgradeable {
...
-        require(params.commission_rate_percentage <= 100, "Staking: invalid commission rate");
-        require(params.min_self_delegation > 0, "Staking: invalid min self delegation");
+        if (params.min_self_delegation != -1) {
+            require(
+                params.commission_rate_percentage <= 100 && params.commission_rate_percentage >= 0,
+                "Staking: invalid commission rate"
+            );
+        }
+        if (params.min_self_delegation != -1) {
+            require(params.min_self_delegation > 0, "Staking: invalid min self delegation");
```
Zodomo, line 203: Move this require statement into the if statement above it. Both if statements have the same condition.

chmllr (approved):

powerslider (approved):

Issue #2930, opened 2025-01-29
Add a progress indicator for restoring node snapshots


powerslider:
#### Problem to Solve
`omega` and `mainnet` have really big node snapshots at this point and require time to download and restore those snapshots. This might result is bad UX for the user, because of the unpredictability of the wait time.

#### Proposed Solution
Implement a progress indicator with elapsed time, eta time, percentage of completion, etc. that will inform the user of the current progress in restoring a node snapshot according to his selected network.

PR #2931, opened 2025-01-29
feat(cli/cmd): add progress bar to node snapshots

powerslider:
improve UX by informing the user on load status for restoring node snapshots via a progress bar and time metrics

issue: #2930


powerslider:
For all reviewers - this is how the progress bar looks:
Screenshot 2025-01-29 at 10 40 26

In order to check it for yourself:
- checkout this branch
- `mkdir snapshot_staging`
- `make install-cli`
- `omni operator init-nodes --network=staging --moniker=foobar --node-snapshot` which will result in checking the progress bar in real time with real metrics.

corverroos (changes requested):
lessismore, lets avoid this.

Issue #2947, opened 2025-01-30
Integrate SolverNet migration into genesis staking claim contract.


Zodomo:
#### Problem to Solve
We want to enable migrating stake and rewards to Omni in one click.

#### Proposed Solution
Integrate SolverNet into Genesis staker reward distribution contract that allows a user to migrate their rewards and/or stake to Omni in one click.

PR #2948, opened 2025-01-30
feat(contracts/core): genesis staker solvernet migration

Zodomo:
Added the `MerkleDistributorWithDeadline` contract and modified it to facilitate migration of rewards and/or stake to Omni via a SolverNet order.

issue: #2947


contracts/core/src/token/MerkleDistributorWithDeadline.sol
```diff
@@ -0,0 +1,129 @@
...
+contract MerkleDistributorWithDeadline is MerkleDistributor, Ownable {
+    using LibBitmap for LibBitmap.Bitmap;
+    using SafeTransferLib for address;
+    using AddrUtils for address;
+
+    error EndTimeInPast();
+    error ClaimWindowFinished();
+    error NoWithdrawDuringClaim();
+
+    bytes32 internal constant ORDER_DATA_TYPEHASH = keccak256(
```
kevinhalliday, line 23: should we make this readable from inbox

contracts/core/src/token/GenesisStake.sol
```diff
@@ -157,14 +171,33 @@ contract GenesisStake is IGenesisStake, OwnableUpgradeable, PausableUpgradeable
...
         emit Withdrawn(msg.sender, amount);
     }
 
+    /**
+     * @notice Migrate a user's stake to the rewards distributor.
+     * @param addr The address of the user to migrate.
+     * @return The amount of tokens migrated.
+     */
+    function migrateStake(address addr) external whenNotPaused returns (uint256) {
+        require(msg.sender == rewardsDistributor, "GenesisStake: only rewards distributor can migrate stake");
```
kevinhalliday, line 187: I'm guessing we might get a "require statement too long" lint error on this

contracts/core/src/token/MerkleDistributorWithDeadline.sol
```diff
@@ -0,0 +1,129 @@
...
+        expenses[0] =
+            ISolverNet.TokenExpense({ token: address(0).toBytes32(), spender: address(0).toBytes32(), amount: amount });
+
+        ISolverNet.Call memory call = ISolverNet.Call({
+            chainId: OMNI_CHAINID,
+            target: account.toBytes32(),
+            value: amount,
+            data: "",
+            expenses: expenses
+        });
```
kevinhalliday, line 118: This is bridges stake to native omni.

I imaging we might want single click migration to native omni staking (in Staking.sol)

That would require letting users choose a validator to delegate too. Given we have allow list of validators, this is okay for now, as that set won't change. 

Check with Fab on UX here.

contracts/core/src/token/MerkleDistributorWithDeadline.sol
```diff
@@ -0,0 +1,129 @@
...
+    using AddrUtils for address;
+
+    error EndTimeInPast();
+    error ClaimWindowFinished();
+    error NoWithdrawDuringClaim();
+
+    bytes32 internal constant ORDER_DATA_TYPEHASH = keccak256(
+        "OrderData(address owner,Call call,Deposit[] deposits)Call(uint64 chainId,bytes32 target,uint256 value,bytes data,TokenExpense[] expenses)TokenExpense(bytes32 token,bytes32 spender,uint256 amount)Deposit(bytes32 token,uint256 amount)"
+    );
+    uint64 internal constant OMNI_CHAINID = 166;
```
kevinhalliday, line 26: Can read this from portal, so we can test out on omega

It will take some manual effort to set up, but it is worth is it IMO

contracts/core/src/token/MerkleDistributorWithDeadline.sol
Zodomo: I was under the impression that you are not supposed to do this, as the frontend needs to derive its own EIP-712 typehash of the given struct type to ensure compatibility.

Zodomo:
Now that #2954 introduced greatly optimized SolverNet contracts, I will refactor this to use those contracts and their order type.

contracts/core/src/token/GenesisStake.sol
Zodomo: Shortened the require error.

contracts/core/src/token/MerkleDistributorWithDeadline.sol
Zodomo: Switched to retrieving this value from the portal.

contracts/core/src/libraries/MerkleDistributor.sol
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity 0.8.24;
+
+import { IMerkleDistributor } from "../interfaces/IMerkleDistributor.sol";
+import { LibBitmap } from "solady/src/utils/LibBitmap.sol";
+import { MerkleProofLib } from "solady/src/utils/MerkleProofLib.sol";
+import { SafeTransferLib } from "solady/src/utils/SafeTransferLib.sol";
+
+contract MerkleDistributor is IMerkleDistributor {
```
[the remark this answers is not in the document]
Zodomo: Moved it to `/token`.

Issue #2963, opened 2025-02-04
gRPC CProvider doesn't support custom gogoproto types


corverroos:
#### Problem to Solve
The gRPC based CProvider doesn't support unmarshalling custom gogoprotobuf types, e.g. `mint-module.InflationReponse.Inflation` which is a `cosmos_math.LegacyDec` type).

```
// QueryInflationResponse is the response type for the Query/Inflation RPC
// method.
message QueryInflationResponse {
  // inflation is the current minting inflation value.
  bytes inflation = 1 [
    (cosmos_proto.scalar)  = "cosmos.Dec",
    (gogoproto.customtype) = "cosmossdk.io/math.LegacyDec",
    (gogoproto.nullable)   = false,
    (amino.dont_omitempty) = true
  ];
}
```

This means we cannot query these endpoint using the gRPC based client.

#### Proposed Solution
Add support for these custom types.

PR #2964, opened 2025-02-04
chore(halo/app): use depinject for encoding config

corverroos:
Provide a `halo_app.ClientEncodingConfig` function, similar to cosmos `MakeTestEncodingConfig` that uses depinject to load the codecs. Use this in the gRPC client so custom types can be decoded.

issue: #2963

### DIFF
--- a/halo/genutil/genutil.go
+++ b/halo/genutil/genutil.go
@@ -1,12 +1,13 @@
 package genutil
 
 import (
+	"context"
 	"encoding/json"
 	"os"
 	"sort"
 	"time"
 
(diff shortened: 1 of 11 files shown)


halo/genutil/genutil.go
```diff
@@ -283,34 +283,6 @@ func defaultAppState(
 	}
 }
 
-func getCodec() *codec.ProtoCodec {
-	// TODO(corver): Use depinject to get all of this.
```
corverroos, line 287: TODO done ✅

powerslider (approved):

PR #2967, opened 2025-02-04
feat(solver): v1 quote api endpoint

kevinhalliday:
Add v1 quotes api endpoint.

issue: #2921

### DIFF
--- a/solver/app/quote.go
+++ b/solver/app/quote.go
@@ -1,6 +1,7 @@
 package app
 
 import (
+	"context"
 	"encoding/json"
 	"math/big"
 	"net/http"
@@ -32,11 +33,19 @@ type QuoteRequest struct {
 
 // QuoteResponse is the response json for the /quote endpoint.
 type QuoteResponse struct {
-	Rejected          bool    `json:"rejected"`
-	RejectReason      string  `json:"rejectReason"`
-	RejectDescription string  `json:"rejectDescription"`
-	Deposit           Deposit `json:"deposit"`
-	Error             Error   `json:"error"`
+	Rejected          bool     `json:"rejected,omitempty"`
+	RejectReason      string   `json:"rejectReason,omitempty"`
+	RejectDescription string   `json:"rejectDescription,omitempty"`
+	Deposit           *Deposit `json:"deposit,omitempty"`
+	Error             *Error   `json:"error,omitempty"`
+}
+
+func (r QuoteResponse) StatusCode() int {
+	if r.Error != nil {
+		return r.Error.Code
+	}
+
+	return http.StatusOK
 }
 
 // Expense is a solver expense on the destination (matches bindings.SolverNetExpense).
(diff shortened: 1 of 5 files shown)


solver/app/quote.go
```diff
@@ -1,63 +1,217 @@
...
+	DepositToken       common.Address `json:"depositToken"`
+}
+
+// QuoteResponse is the response json for the /quote endpoint.
+type QuoteResponse struct {
+	Rejected          bool     `json:"rejected,omitempty"`
+	RejectReason      string   `json:"rejectReason,omitempty"`
+	RejectDescription string   `json:"rejectDescription,omitempty"`
+	Deposit           *Deposit `json:"deposit,omitempty"`
+	Error             *Error   `json:"error,omitempty"`
```
corverroos, line 40: made this pointers so `omitempty` actually works

solver/app/quote.go
```diff
@@ -1,63 +1,217 @@
...
+	Code    int    `json:"code"`
+	Status  string `json:"status"`
+	Message string `json:"message"`
+}
+
+type quoteFunc func(context.Context, QuoteRequest) (Deposit, error)
+
+// newQuoter returns a quoteFunc that can be used to quote deposits for expenses.
+// It is the logic behind the /quote endpoint.
+func newQuoter(backends ethbackend.Backends, solverAddr common.Address) quoteFunc {
```
corverroos, line 83: Split logic from http wrangling. This simplifies testing which then doesn't need to do http stuff and can focus on logic.

solver/app/quote.go
```diff
@@ -1,63 +1,217 @@
...
+	})
+}
+
 // getQuote returns payment in `depositTkns` required to pay for `expenses`.
 //
 // For now, this is a simple quote that requires a single expense, paid
 // for by an equal amount of an equivalent deposit token. Token equivalence is
 // determined by symbol (ex arbitrum "ETH" is equivalent to optimism "ETH").
-func getQuote(depositTkns []Token, expenses []Payment) ([]Payment, RejectOrErr) {
+func getQuote(depositTkns []Token, expenses []Payment) ([]Payment, error) {
```
corverroos, line 180: refactored `RejectOrError` into standard `error`:
 - Return `RejectionError` which is a rejection
 - Or return `errors.New` which is a error
 
 Handle using `errors.As`

solver/app/quote.go
```diff
@@ -1,63 +1,217 @@
...
-	}, RejectOrErr{}
+	}, nil
 }
 
 // coversQuote checks if `deposits` match or exceed a `quote` for expenses.
-func coversQuote(deposits, quote []Payment) RejectOrErr {
+func coversQuote(deposits, quote []Payment) error {
 	if len(quote) != len(deposits) {
-		return RejectOrErr{}
+		return errors.New("invalid params [BUG]")
```
corverroos, line 214: I assume this is unexpected and not actually valid.

corverroos (approved):

solver/app/quote.go
kevinhalliday: Actually can remove this

kevinhalliday: or maybe i lied

kevinhalliday: nvm unrelated test failed

PR #2983, opened 2025-02-06
chore(e2e/xbridge): refactor and test

kevinhalliday:
Refactor e2e/xbridge, test rlusd bridging.

issue: #2972


contracts/misc/src/bridge/Bridge.sol
```diff
@@ -26,7 +26,7 @@ contract Bridge is Initializable, AccessControlUpgradeable, PausableUpgradeable,
     /**
      * @dev Default gas limit for xcalls.
      */
-    uint64 internal constant DEFAULT_GAS_LIMIT = 140_000;
+    uint64 internal constant DEFAULT_GAS_LIMIT = 200_000;
```
kevinhalliday, line 29: Might be able to go lower than this.

This was required to cover gas limit of `receiveToken` for chains with a lockbox. Which means we over pay on chains without lockbox.

contracts/misc/src/bridge/Bridge.sol
```diff
@@ -220,13 +220,11 @@ contract Bridge is Initializable, AccessControlUpgradeable, PausableUpgradeable,
...
             ILockbox(_lockbox).token().safeTransferFrom(msg.sender, address(this), value);
-            ILockbox(_lockbox).deposit(value);
-        } else {
-            _token.safeTransferFrom(msg.sender, address(this), value);
+            ILockbox(_lockbox).depositTo(msg.sender, value);
         }
 
         // Burn the tokens.
-        ITokenOps(_token).burn(value);
+        ITokenOps(_token).clawback(msg.sender, value);
```
kevinhalliday, line 227: @Zodomo this was main change.

- using clawback means less token approvals needed
- RLUSD only has [clawbacker in initializer](https://github.com/ripple/RLUSD-Implementation/blob/07a27d74863a22f6ad403b5a9f8bb3aea76e05eb/contracts/StablecoinUpgradeable.sol#L44), so this way to we don't need another setup tx to assign BURNER role

e2e/app/run.go
```diff
@@ -174,15 +174,16 @@ func E2ETest(ctx context.Context, def Definition, cfg E2ETestConfig) error {
 		return err
 	}
 
+	stopReceiptMonitor := StartMonitoringReceipts(ctx, def)
+
 	var eg errgroup.Group
 	eg.Go(func() error { return testGasPumps(ctx, def) })
 	eg.Go(func() error { return testBridge(ctx, def) })
+	eg.Go(func() error { return xbridge.Test(ctx, NetworkFromDef(def), ExternalEndpoints(def)) })
```
kevinhalliday, line 182: Put here because it was helpful in debugging. Also, putting in `e2e/tests/` caused lot of nonce conflicts with solver tests. Something to think about. Solution can be part of the on going discussion on how we should structure e2e tests going forward

contracts/misc/src/bridge/Bridge.sol
kevinhalliday: @Zodomo

Zodomo: I'll add dynamic fees in a separate PR.

Zodomo (approved):

PR #2985, opened 2025-02-06
feat(contracts/misc): bridge dynamic gas

Zodomo:
More gas is needed to send tokens to bridges with a lockbox, so these changes allow for the gas calculation to reflect that so gas isn't needlessly wasted by defaulting to the upper bound.

issue: none


contracts/misc/src/bridge/Bridge.sol
```diff
@@ -24,9 +24,14 @@ contract Bridge is Initializable, AccessControlUpgradeable, PausableUpgradeable,
...
-     * @dev Default gas limit for xcalls.
+     * @dev Gas limit for xcalls to bridges without a lockbox.
      */
-    uint64 internal constant DEFAULT_GAS_LIMIT = 200_000;
+    uint64 internal constant DEFAULT_GAS_LIMIT = 125_000;
+
+    /**
+     * @dev Gas limit for xcalls to bridges with a lockbox.
+     */
+    uint64 internal constant LOCKBOX_GAS_LIMIT = 180_000;
```
kevinhalliday, line 34: maybe something in both these names to show they are `receieveToken` gas limitms

kevinhalliday (approved):

contracts/misc/src/bridge/Bridge.sol
Zodomo: I'll prepend `RECEIVE_`

Issue #2992, opened 2025-02-07
Support migrating directly to Omni Core Staking


Zodomo:
#### Problem to Solve
Right now the GenesisStaker migration only opens an order to send them their tokens on Omni Core. We want to be able to migrate the user's staked position in its entirety, so there needs to be a way to trigger a SolverNet order to deposit the user's stake into Staking on Omni Core, rather than just sending them their balance.

#### Proposed Solution
Implement a flow that allows the MerkleDistributor contract to create a SolverNet order for Omni Core Staking.

PR #2993, opened 2025-02-07
feat(contracts/core): support migration to omni staking

Zodomo:
Added support for migrating GenesisStake staked balance, rewards, and additional tokens from the user to Omni Core, and added an optional order flow where this balance can be delegated to a validator on behalf of the user via the Omni Core Staking contract. I deployed a test of this flow to staging and it was successful, the script details the process followed.

issue: #2992


Zodomo:
Closing as delegation on Omni Core will remain a separate action.

PR #2996, opened 2025-02-07
feat(sdk): order hooks

ga-reth:
Order hooks

issue: none

### DIFF
--- a/sdk/src/hooks/useOrderStatus.ts
+++ b/sdk/src/hooks/useOrderStatus.ts
@@ -1,0 +1,103 @@
+import { useMemo } from 'react'
+import { type Log, decodeEventLog } from 'viem'
+import {
+  type UseWaitForTransactionReceiptReturnType,
+  useReadContract,
+} from 'wagmi'
+import { inbox, outbox } from '../index.js'
+import type { OrderStatus } from '../types/orderStatus.js'
+
+type UseOrderStatusParams = {
+  txStatus?: UseWaitForTransactionReceiptReturnType['status']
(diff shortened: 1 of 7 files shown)


sdk/src/hooks/useOrderStatus.ts
```diff
@@ -0,0 +1,103 @@
...
+      abi: inbox.abi,
+      eventName: 'Open',
+      data: txLogs[txLogs.length - 1].data,
+      topics: txLogs[txLogs.length - 1].topics,
+    })
+    return {
+      id: openEvent.args.resolvedOrder.orderId,
+      originData: openEvent.args.resolvedOrder.fillInstructions[0].originData,
+    }
+  }, [txStatus, txLogs])
```
kevinhalliday, line 56: can be for later cause I know you're on a bit of a time crunch,  but IMO these should be a separate hook.  `useOrderStatus`should just take an `orderId`.

that way `userOrderStatus` can be used without an `open` tx preceding it

You can get the fill origin data you need for `didFill` in the same `getOrder` contract read we're already making

something like

```typescript
const useOpenOrder () => {
     // ...
     const { orderId } = useOrderFromOpenTx({ hash, logs })  // use tx hash or logs if you have htem
     const { status } = useOrderStatus({ orderId })
}
```

kevinhalliday (approved):

sdk/src/hooks/useOrderStatus.ts
ga-reth: nice idea, i've updated

PR #2998, opened 2025-02-07
feat(sdk): encode call data internally

ga-reth:
Handle call data encoding internally

issue: none

### DIFF
--- a/sdk/src/types/order.ts
+++ b/sdk/src/types/order.ts
@@ -60,6 +60,6 @@ type Call = {
   readonly target: Address
   readonly value: bigint
   // TODO: infer selector and args from abi
-  readonly selector: Hex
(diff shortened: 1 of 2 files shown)


sdk/src/types/order.ts
```diff
@@ -56,8 +56,10 @@ type Expense = {
...
 
 type Call = {
+  readonly abi: Abi
   readonly target: Address
-  readonly selector: Hex
   readonly value: bigint
-  readonly params: Hex
+  // TODO: infer selector and args from abi
+  readonly selector: Hex
+  readonly args?: unknown[]
```
kevinhalliday, line 64: let's match viem with `functionName`

kevinhalliday (approved):

PR #2999, opened 2025-02-07
feat(contracts/solve): delegateFor via SolverNet staging script

Zodomo:
Altered the `stagingaddrs` script so I could use its results in a foundry script via `ffi`. This allowed me to write a foundry script that always operates on the current staging contract addresses and uses them to issue SolverNet orders for a `delegateFor` call on the Omni Core `Staking` contract. We could eventually hook this up to github workflows as an easy way to use PRs as a way to generate test SolverNet activity.

issue: none


kevinhalliday (approved):

kevinhalliday (approved):
👁️ 👏

PR #3000, opened 2025-02-10
fix(octane/evmengine): fix proposing invalid blobs

corverroos:
Fixes issue where malicious proposers can trigger `retryForever` by proposing invalid `BlobCommitments`.

issue: none

### DIFF
--- a/octane/evmengine/keeper/msg_server.go
+++ b/octane/evmengine/keeper/msg_server.go
@@ -35,8 +35,13 @@ func (s msgServer) ExecutionPayload(ctx context.Context, msg *types.MsgExecution
 		return nil, err
 	}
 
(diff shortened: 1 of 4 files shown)


octane/evmengine/keeper/msg_server.go
```diff
@@ -35,8 +35,13 @@ func (s msgServer) ExecutionPayload(ctx context.Context, msg *types.MsgExecution
 		return nil, err
 	}
 
+	blockHashes, err := blobHashes(msg.BlobCommitments)
```
corverroos, line 38: do parsing here, instead of inside `retryForever`

powerslider (approved):

PR #3017, opened 2025-02-10
chore(halo/app): fix backwards e2e test

corverroos:
Fixes failing `backwards` e2e manifest by moving the new distribution module "burner" permission behind the `evm-staking-module` feature flag. Adding it changes app hash, so can only be done as part of network upgrade.

issue: none

### DIFF
--- a/halo/app/app_config.go
+++ b/halo/app/app_config.go
@@ -132,14 +132,27 @@ var (
 		minttypes.ModuleName,
 	}
 
-	moduleAccPerms = []*authmodulev1.ModuleAccountPermission{
-		{Account: authtypes.FeeCollectorName},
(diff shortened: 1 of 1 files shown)


halo/app/app_config.go
```diff
@@ -132,14 +132,27 @@ var (
 		minttypes.ModuleName,
 	}
 
-	moduleAccPerms = []*authmodulev1.ModuleAccountPermission{
-		{Account: authtypes.FeeCollectorName},
-		{Account: distrtypes.ModuleName, Permissions: []string{authtypes.Burner}},
```
corverroos, line 137: previously distribution didn't have burner permission, this was added earlier today. Revert that and put it behind the feature flag.

PR #3041, opened 2025-02-12
chore(solver): match quote with reject

kevinhalliday:
Match reject cases in quote. Particularly, dest call reverts.

This requires refactoring checkFill to not approve outbox spend on demand.
Instead, we set max approvals on startup (see #3039)

Quote / reject tests were refactored to share tests cases.

issue: #2965


solver/app/server.go
```diff
@@ -19,6 +19,9 @@ type JSONErrorResponse struct {
 	Message string `json:"message"`
 }
 
+// removeBUG removes [BUG] from the error messages, so they are not included in responses to users.
+func removeBUG(s string) string { return strings.ReplaceAll(s, "[BUG]", "") }
```
kevinhalliday, line 23: remove `[BUG]` from errors in api responses?

solver/app/approvals.go
```diff
@@ -133,3 +127,17 @@ func approveToken(ctx context.Context, backend *ethbackend.Backend, token Token,
...
+	if err != nil {
+		return false, errors.Wrap(err, "new token")
+	}
+
+	allowance, err := tkn.Allowance(&bind.CallOpts{Context: ctx}, solverAddr, outboxAddr)
+	if err != nil {
+		return false, errors.Wrap(err, "get allowance")
+	}
+
+	return new(big.Int).Sub(allowance, umath.MaxUint256).Sign() >= 0, nil
```
corverroos, line 142: not sure I get the math here? Isn't `x-umath.MaxUint256` always negative?

solver/app/reject.go
```diff
@@ -99,43 +108,28 @@ func newShouldRejector(
...
 		}
 
 		r := new(RejectionError)
 		if !errors.As(err, &r) { // Error, but no rejection
-			return rejectNone, false, err
+			return ShouldRejectResult{}, err
 		}
 
-		// TODO:  handle upstream, do not accept if already filled
 		e := new(AlreadyFilledError)
```
corverroos, line 127: Maybe not required now, but I would split this `AlreadyFill` out as high level check since itis something different from rejections
```
			switch event.Status {
			case statusPending:
				if ok, _ := alreadyFill(); ok {
					return errors.New("unexpected already filled [BUG]")
				}

				if didReject, err := maybeReject(); err != nil {
					return err
				} else if didReject {
					continue
				}

				log.Info(ctx, "Accepting order")
				if err := deps.Accept(ctx, chainID, order); err != nil {
					return errors.Wrap(err, "accept order")
				}
			case statusAccepted:
				if ok, _ := alreadyFill(); ok {
					// Nothing to do
					return nil
				}
				
				// check reject again, as liquidity (or other conditions) may have changed
				if didReject, err := maybeReject(); err != nil {
					return err
				} else if didReject {
					continue
				}

				log.Info(ctx, "Filling order")
				if err := deps.Fill(ctx, chainID, order); err != nil {
					return errors.Wrap(err, "fill order")
				}
```

corverroos (approved):

solver/app/reject.go
kevinhalliday: yeah was feeling similarily didn't love this part

PR #3050, opened 2025-02-12
chore(monitor): flowgen skeleton

kevinhalliday:
Add flowgen skeleton.

issue: #3006

### DIFF
--- a/monitor/flowgen/flowgen.go
+++ b/monitor/flowgen/flowgen.go
@@ -1,0 +1,26 @@
+package flowgen
+
+import (
+	"context"
+	"time"
+
+	"github.com/omni-network/omni/lib/tokens"
+)
+
+func Start(ctx context.Context) error {
+	jobs := []Job{
+		{
+			Name:    "Symbiotic",
+			Run:     newSymbioticRunner(),
+			Cadence: 1 * time.Hour,
+			Spend:   spend(tokens.WSTETH, ether1),
+		},
+	}
+
+	_ = jobs
+	_ = ctx
+
+	// TODO
+
+	return nil
+}
--- a/monitor/flowgen/job.go
+++ b/monitor/flowgen/job.go
@@ -1,0 +1,27 @@
+package flowgen
+
+import (
+	"context"
+	"math/big"
+	"time"
+
+	"github.com/omni-network/omni/lib/tokens"
+)
+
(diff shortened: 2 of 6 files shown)


monitor/flowgen/job.go
```diff
@@ -0,0 +1,40 @@
...
+
+	// Spend is spend this job uses on one run (exluding gas)
+	Spend Spend
+}
+
+// cronner returns a cron.Job for the given Job.
+func cronner(ctx context.Context, j Job) cron.Job {
+	return cron.FuncJob(func() {
+		if err := j.Run(ctx, j.Spend); err != nil {
+			log.Error(ctx, "Job failed", err, "name", j.Name)
```
corverroos, line 35: rather warn, since it retries, also add `(will retry)`.

Also add a return, since you will always log `log.Info(ctx, "Job finished", "name", j.Name)`

monitor/flowgen/util.go
```diff
@@ -0,0 +1,16 @@
...
+	"math/big"
+
+	"github.com/omni-network/omni/lib/tokens"
+)
+
+func spend(token tokens.Token, amt float64) Spend {
+	return Spend{token: unit(amt, token.Decimals)}
+}
+
+func unit(amt float64, decimals uint) *big.Int {
```
corverroos, line 13: please write some unit tests for this

corverroos: can also use `math.Pow10`

also, maybe panic if `g < 1`

monitor/flowgen/flowgen.go
```diff
@@ -0,0 +1,37 @@
...
+	c := cron.New(
+		cron.WithLocation(time.UTC),
+		cron.WithLogger(newCronLogger(ctx)),
+	)
+
+	jobs := []Job{
+		{
+			Name:  "Symbiotic",
+			Run:   newSymbioticRunner(),
+			Spec:  "0 0 * * *",
```
corverroos, line 23: do we really need this? why not just simple periodic loops like we normally do?

corverroos (approved):

monitor/flowgen/util.go
kevinhalliday: yeah will add

kevinhalliday: just skelotonning

Issue #3087, opened 2025-02-17
Remove accept from SolverNet flow


Zodomo:
#### Problem to Solve
`accept` costs our solver money, increases latency, and gets in the way of faster solving. Many implementations do not have this as a step.

#### Proposed Solution
Remove `accept` completely, refactor `cancel` to close orders after fillDeadline only (solver can still reject), and credit the order to the address the solver provides at fulfillment.

PR #3088, opened 2025-02-17
feat(contracts/solve): remove accept, replace cancel, enforce fillDeadline

Zodomo:
`accept` removed from Inbox, `cancel` refactored into `close`, `fillDeadline` must be provided in all cases, solver must provide an address during `fill`.

issue: #3087


contracts/solve/src/SolverNetInbox.sol
```diff
@@ -203,37 +183,36 @@ contract SolverNetInbox is OwnableRoles, ReentrancyGuard, Initializable, Deploye
...
-    function cancel(bytes32 id) external nonReentrant {
+    function close(bytes32 id) external nonReentrant {
         OrderState memory state = _orderState[id];
-        address user = _orderHeader[id].owner;
+        SolverNet.Header memory header = _orderHeader[id];
 
         if (state.status != Status.Pending) revert OrderNotPending();
-        if (user != msg.sender) revert Unauthorized();
+        if (header.owner != msg.sender) revert Unauthorized();
+        if (header.fillDeadline >= block.timestamp) revert OrderStillValid();
```
kevinhalliday, line 196: Add buffer 🤝 

Also some other maybe ideas for error name`OrderNotExpired()` ,  `OrderStillOpen()`? no hard preferences just spit balling

kevinhalliday (approved):

Issue #3089, opened 2025-02-18
Add close buffer and pausable logic


Zodomo:
#### Problem to Solve
Now that `accept` was removed, if the Omni Core relayer goes down and an order is filled, the user could close it after the `fillDeadline`. Additionally, we cannot pause the contracts in case there are cases where fulfillment attestation is held up.

#### Proposed Solution
Enforce a buffer after the `fillDeadline` before allowing the user to `close` their order. In the event the Core relayer goes down or something prevents fulfillment attestation from reaching the origin chain, implement pausable logic to halt `open` and/or `close` to prevent bad cancellations or orders being opened on bugged routes. Allow the solver to pause these functions.

PR #3090, opened 2025-02-18
feat(contracts/solve): add close buffer and pausable logic

Zodomo:
Now that `accept` has been removed, `close` needed to be refactored a bit to protect fillers from Omni Core relayer faults. A close buffer was added to give us/solver time to react to messaging shards having issues, and pausable logic enables us to halt `open` and/or `close` in the event anything is going wrong.

issue: #3089


contracts/solve/src/SolverNetInbox.sol
```diff
@@ -77,6 +103,19 @@ contract SolverNetInbox is OwnableRoles, ReentrancyGuard, Initializable, Deploye
      */
     mapping(Status => bytes32 id) internal _latestOrderIdByStatus;
 
+    /**
+     * @notice Modifier to ensure contract functions are not paused.
+     */
+    modifier whenNotPaused(bytes32 pauseKey) {
+        uint8 _pauseState = pauseState;
+        if (_pauseState != 0) {
```
kevinhalliday, line 111:
```suggestion
        if (_pauseState != NONE_PAUSED) {
```

kevinhalliday (approved):

PR #3094, opened 2025-02-18
chore(e2e/solve): better e2e solve tests

kevinhalliday:
Track orders by status, not balance checks. 

This lets us:
- define and tests more valid order types (different fill side effects)
- add more test rejects
- fails earlier when something fails

issue: none


corverroos (approved):

e2e/solve/testutil.go
```diff
@@ -0,0 +1,164 @@
...
+
+	if err := eg.Wait(); err != nil {
+		return errors.Wrap(err, "wait group")
+	}
+
+	return nil
+}
+
+func mintAndApprove(ctx context.Context, backends ethbackend.Backends, order TestOrder) error {
+	if (order.Deposit.Token == common.Address{}) {
```
powerslider, line 118: You can add `var zeroAddr common.Address` at the beginning of the file and compare against it. This way you will avoid inline struct creation and would be a cleaner comparison.
```
if order.Deposit.Token == zeroAddr {
           return nil
}
``` 

Also ditch the parenthesis in the `if` statement.

e2e/solve/test.go
```diff
@@ -2,35 +2,36 @@ package solve
```
powerslider, line 2: I would rename `test.go`  to `order_test.go` to better represent what is being tested.

e2e/solve/testutil.go
```diff
@@ -0,0 +1,164 @@
+package solve
```
powerslider, line 1: I would rename `testutil.go` to `order_tracker.go` to better represent what this file contains.

kevinhalliday: there is more in there than that

kevinhalliday: ill put order tarcket into a separate file

powerslider (approved):

Issue #3098, opened 2025-02-19
Withdrawals


chmllr:
#### Problem to Solve

Allow delegators to withdraw their stakes and validators to withdraw their commissions.

#### Proposed Solution

See ADR.

PR #3099, opened 2025-02-19
feat(octane/evmengine): fetch eligible withdrawals for block inclusion

chmllr:
Implements the querying of eligible withdrawals by height.

issue: #3098


powerslider (approved):

octane/evmengine/keeper/db.go
```diff
@@ -113,3 +114,34 @@ func (k *Keeper) listWithdrawalsByAddress(ctx context.Context, withdrawalAddr co
 
 	return withdrawals, nil
 }
+
+// ListWithdrawalsBelowHeight returns all withdrawals created below the specified height, sorted by the created height, limited by the provided count.
+func (k *Keeper) ListWithdrawalsBelowHeight(ctx context.Context, height uint64, limit int) ([]*Withdrawal, error) {
```
corverroos, line 119: why not `EligibleWithdrawals` that returns the thing we need exactly?

corverroos (changes requested):

octane/evmengine/keeper/db.go
chmllr: Sure, we can go with that name as well.

octane/evmengine/keeper/db.go
```diff
@@ -113,3 +113,33 @@ func (k *Keeper) listWithdrawalsByAddress(ctx context.Context, withdrawalAddr co
...
+			return nil, errors.Wrap(err, "get withdrawal")
+		}
+
+		if val.GetCreatedHeight() < uint64(height) {
+			withdrawals = append(withdrawals, val)
+
+			if uint64(len(withdrawals)) == k.maxWithdrawalsPerBlock {
+				break
+			}
+		}
```
corverroos, line 141: You only break out early when limit is reach, I would suggest to also break out if `height` is reached, otherwise you iterate over all withdrawals unnecasrily which increases load.

Also, best to avoid nesting if possible.

Also, can use the `umath.Len` function

```
if val.GetCreatedHeight() >= uint64(height) {
  // Withdrawals created in this block are not eligible
  break
} 

withdrawals = append(withdrawals, val)

if umath.Len(withdrawals) >= k.maxWithdrawalsPerBlock {
  break
}

```

corverroos (approved):

PR #3107, opened 2025-02-20
chore(scripts/join): add compose tear down logic

powerslider:
add docker compose down procedure to cleanup all containers after a test run

issue: none


scripts/join/join_test.go
```diff
@@ -357,3 +368,62 @@ func getContainerStats(ctx context.Context) (stats, error) {
 
 	return resp, nil
 }
+
+func shutdownContainers(home string, output *os.File, logsPath string) error {
+	// Create a context with timeout to ensure clean up doesn't hang indefinitely.
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
chmllr, line 374: We prob should reuse the `ctx` from the parent func instead of doung `Background()` — it also has a timeout.

chmllr (approved):

PR #3124, opened 2025-02-21
feat(sdk): refactor and api check call

ga-reth:
Refactor SDK interface and add /check api call

issue: none


sdk/src/hooks/useOrder.ts
```diff
@@ -0,0 +1,199 @@
...
+  validationError?: ValidationError
+  orderStatus?: OrderStatus
+  canOpen: boolean
+  isTxPending: boolean
+  isTxSubmitted: boolean
+  isValidated: boolean
+  isOpen: boolean
+  isRejected: boolean
+  isError: boolean
+  mutation: UseWriteContractReturnType<Config, unknown>
```
kevinhalliday, line 39: `txMutation`?

kevinhalliday (approved):

sdk/src/hooks/useOrder.ts
ga-reth: good catch - fixed

PR #3134, opened 2025-02-21
feat(octane/evmengine): withdrawals

chmllr:
WIP

issue: #3098

### DIFF
--- a/octane/evmengine/keeper/abci.go
+++ b/octane/evmengine/keeper/abci.go
@@ -285,11 +285,16 @@ func (k *Keeper) startBuild(ctx context.Context, appHash common.Hash, timestamp 
 		FinalizedBlockHash: headHash,
 	}
 
+	withdrawals, err := k.EligibleWithdrawals(ctx)
+	if err != nil {
+		return engine.ForkChoiceResponse{}, errors.Wrap(err, "eligible withdrawals")
+	}
+
 	attrs := &engine.PayloadAttributes{
 		Timestamp:             ts,
 		Random:                headHash, // We use head block hash as randao.
 		SuggestedFeeRecipient: k.feeRecProvider.LocalFeeRecipient(),
-		Withdrawals:           []*etypes.Withdrawal{}, // Withdrawals not supported yet.
+		Withdrawals:           withdrawals,
 		BeaconRoot:            &appHash,
 	}
 
--- a/octane/evmengine/keeper/db.go
+++ b/octane/evmengine/keeper/db.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/beacon/engine"
 	"github.com/ethereum/go-ethereum/common"
+	etypes "github.com/ethereum/go-ethereum/core/types"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
@@ -117,7 +118,7 @@ func (k *Keeper) listWithdrawalsByAddress(ctx context.Context, withdrawalAddr co
 
 // EligibleWithdrawals returns all withdrawals created below the specified height,
 // sorted by the id (oldest to newest), limited by the provided count.
-func (k *Keeper) EligibleWithdrawals(ctx context.Context) ([]*Withdrawal, error) {
+func (k *Keeper) EligibleWithdrawals(ctx context.Context) ([]*etypes.Withdrawal, error) {
 	height := sdk.UnwrapSDKContext(ctx).BlockHeight()
 	// Note: items are ordered by the id in ascending order (oldest to newest).
 	iter, err := k.withdrawalTable.List(ctx, WithdrawalPrimaryKey{})
@@ -146,5 +147,16 @@ func (k *Keeper) EligibleWithdrawals(ctx context.Context) ([]*Withdrawal, error)
 		}
 	}
 
-	return withdrawals, nil
+	var evmWithdrawals []*etypes.Withdrawal
+	for _, w := range withdrawals {
+		evmWithdrawals = append(evmWithdrawals, &etypes.Withdrawal{
+			Index:   w.GetId(),
+			Address: common.BytesToAddress(w.GetAddress()), //nolint:forbidigo // should be padded
+			Amount:  w.GetAmountGwei(),
+			// The validator index is not used for withdrawals.
+			Validator: 0,
+		})
+	}
+
+	return evmWithdrawals, nil
 }
--- a/octane/evmengine/keeper/db.go
+++ b/octane/evmengine/keeper/db.go
@@ -160,3 +160,15 @@ func (k *Keeper) EligibleWithdrawals(ctx context.Context) ([]*etypes.Withdrawal,
 
 	return evmWithdrawals, nil
 }
+
+// RemoveWithdrawals removes all passed withdrawals by the id.
+func (k *Keeper) RemoveWithdrawals(ctx context.Context, withdrawals []*etypes.Withdrawal) error {
+	for _, w := range withdrawals {
+		err := k.withdrawalTable.DeleteBy(ctx, WithdrawalIdIndexKey{}.WithId(w.Index))
+		if err != nil {
+			return errors.Wrap(err, "remowing withdrawal", "id", w.Index)
+		}
+	}
+
+	return nil
+}
--- a/octane/evmengine/keeper/keeper.go
+++ b/octane/evmengine/keeper/keeper.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/beacon/engine"
 	"github.com/ethereum/go-ethereum/common"
+	etypes "github.com/ethereum/go-ethereum/core/types"
 
 	ormv1alpha1 "cosmossdk.io/api/cosmos/orm/v1alpha1"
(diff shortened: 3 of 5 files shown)


corverroos:
Closing to reduce noise since we are shelving this in favour of solvernet push. Can re-open when re-prioritised

octane/evmengine/keeper/abci.go
```diff
@@ -300,7 +300,8 @@ func (k *Keeper) startBuild(ctx context.Context, appHash common.Hash, timestamp
 		FinalizedBlockHash: headHash,
 	}
 
-	withdrawals, err := k.EligibleWithdrawals(ctx)
+	callCtx := sdk.UnwrapSDKContext(ctx).WithBlockHeight(height)
+	withdrawals, err := k.EligibleWithdrawals(callCtx) //nolint:contextcheck // no new ctx
```
taeguk, line 304: FYI: This also worked for me:
https://github.com/mitosis-org/omni/commit/6764a63df55b485950470a049352b722971f9fa7

chmllr: Thanks @taeguk but using all available withdrawals is incorrect. Due to some octane-specific implementation details, when the proposal (containing the withdrawals) is created, we have no effects of BeginBlockers on the consensus state yet, but when we execute the proposal we do run  the BeginBlockers which might create new withdrawals which we didn't observe during the proposal validation. Hence, we have to skip the current block height entirely.

taeguk: Yeah you're right. I was wrong. Using all available withdrawals is incorrect in the manner of general usage.

Fortunately, I called `InsertWithdrawal()` only from `EndBlocker` and event processors, not from `BeginBlocker`, so this was fine for my limited use case. It worked fine for now, but it could be bug-prone in future code changes.

octane/evmengine/keeper/abci.go
```diff
@@ -300,11 +300,22 @@ func (k *Keeper) startBuild(ctx context.Context, appHash common.Hash, timestamp
 		FinalizedBlockHash: headHash,
 	}
 
+	intHeight, err := umath.ToInt64(height)
+	if err != nil {
+		return engine.ForkChoiceResponse{}, errors.Wrap(err, "height conversion")
+	}
+
+	callCtx := sdk.UnwrapSDKContext(ctx).WithBlockHeight(intHeight)
+	withdrawals, err := k.EligibleWithdrawals(callCtx) //nolint:contextcheck // no new ctx
```
corverroos, line 309: We used `ctx.Height` inside `EligibleWithdrawals`, assuming it would always be correct. However, we've found that this isn't always the case. Instead of still relying on `ctx.Height`, we should explicitly pass the desired height rather than providing it implicitly via a fake context height.

octane/evmengine/keeper/db.go
```diff
@@ -160,5 +161,28 @@ func (k *Keeper) EligibleWithdrawals(ctx context.Context) ([]*Withdrawal, error)
 		}
 	}
 
-	return withdrawals, nil
+	evmWithdrawals := []*etypes.Withdrawal{}
```
corverroos, line 164: add a comment that nil withdrawals isn't supported, so using empty slice by default

octane/evmengine/keeper/db.go
```diff
@@ -160,5 +161,28 @@ func (k *Keeper) EligibleWithdrawals(ctx context.Context) ([]*Withdrawal, error)
...
+	for _, w := range withdrawals {
+		evmWithdrawals = append(evmWithdrawals, &etypes.Withdrawal{
+			Index:   w.GetId(),
+			Address: common.BytesToAddress(w.GetAddress()), //nolint:forbidigo // should be padded
```
corverroos, line 168: use proper `cast.EthAddress`

octane/evmengine/keeper/db.go
```diff
@@ -127,7 +128,7 @@ func (k *Keeper) listWithdrawalsByAddress(ctx context.Context, withdrawalAddr co
 
 // EligibleWithdrawals returns all withdrawals created below the specified height,
 // sorted by the id (oldest to newest), limited by the provided count.
-func (k *Keeper) EligibleWithdrawals(ctx context.Context) ([]*Withdrawal, error) {
+func (k *Keeper) EligibleWithdrawals(ctx context.Context) ([]*etypes.Withdrawal, error) {
```
corverroos, line 131: also add a comment explaining why we exclude withdrawals before height.

octane/evmengine/keeper/keeper.go
```diff
@@ -270,3 +276,20 @@ func (k *Keeper) getOptimisticPayload() (engine.PayloadID, uint64, time.Time) {
 
 	return k.mutablePayload.ID, k.mutablePayload.Height, k.mutablePayload.UpdatedAt
 }
+
+func withdrawalsEqual(w1, w2 []*etypes.Withdrawal) bool {
```
corverroos, line 280: best do nil checks as well (for each element)

octane/evmengine/keeper/keeper.go
```diff
@@ -270,3 +276,20 @@ func (k *Keeper) getOptimisticPayload() (engine.PayloadID, uint64, time.Time) {
...
+	if len(w1) != len(w2) {
+		return false
+	}
+
+	for i, w := range w1 {
+		if w.Index != w2[i].Index ||
```
corverroos, line 286: the fields of the withdrawal are all comparable, so I think you can do:
```
if w1 == nil || w2 == nil {
  return false
} else if *w1 != *w2 {
  return false
}
```

Also remember to write a unit test for this function to ensure malicious proposer cant slip in invalid withdrawals.

octane/evmengine/keeper/db.go
corverroos: also, don't export stuff by defaut, rather keep things private by default

octane/evmengine/keeper/db.go
```diff
@@ -160,5 +161,28 @@ func (k *Keeper) EligibleWithdrawals(ctx context.Context) ([]*Withdrawal, error)
...
+			// The validator index is not used for withdrawals.
+			Validator: 0,
+		})
+	}
+
+	return evmWithdrawals, nil
+}
+
+// RemoveWithdrawals removes all passed withdrawals by the id.
+func (k *Keeper) RemoveWithdrawals(ctx context.Context, withdrawals []*etypes.Withdrawal) error {
```
corverroos, line 179: same here, don't export by default

corverroos: nit: maybe use a more "DB" term "deleteWithdrawals"

corverroos: we also need a unit test for this function (ensure it doesn't delete other withdrawals).

octane/evmengine/keeper/db.go
```diff
@@ -160,5 +161,28 @@ func (k *Keeper) EligibleWithdrawals(ctx context.Context) ([]*Withdrawal, error)
...
+	for _, w := range withdrawals {
+		err := k.withdrawalTable.DeleteBy(ctx, WithdrawalIdIndexKey{}.WithId(w.Index))
```
corverroos, line 181: note you can also use `.Delete(ctx, &Withdrawal{Id:w.Index})`

octane/evmengine/keeper/abci.go
chmllr: fully agree, but I thought you still want it in ctx 👍

octane/evmengine/keeper/db.go
chmllr: added

octane/evmengine/keeper/db.go
```diff
@@ -125,14 +126,12 @@ func (k *Keeper) listWithdrawalsByAddress(ctx context.Context, withdrawalAddr co
...
-// EligibleWithdrawals returns all withdrawals created below the specified height,
-// sorted by the id (oldest to newest), limited by the provided count.
-func (k *Keeper) EligibleWithdrawals(ctx context.Context) ([]*Withdrawal, error) {
-	height, err := umath.ToUint64(sdk.UnwrapSDKContext(ctx).BlockHeight())
-	if err != nil {
-		return nil, err
-	}
-
+// eligibleWithdrawals returns all withdrawals created below the specified height, sorted by the
+// id (oldest to newest), limited by the provided count.
```
corverroos, line 130: prefer technical DB terms like "ascending"

octane/evmengine/keeper/keeper.go
```diff
@@ -270,3 +281,21 @@ func (k *Keeper) getOptimisticPayload() (engine.PayloadID, uint64, time.Time) {
 
 	return k.mutablePayload.ID, k.mutablePayload.Height, k.mutablePayload.UpdatedAt
 }
+
+// withdrawalsEqual returns true only if two non-nil slices contain identical
+// withdrawals in identical order.
+func withdrawalsEqual(w1, w2 []*etypes.Withdrawal) bool {
+	if w1 == nil || w2 == nil || len(w1) != len(w2) {
```
corverroos, line 288: in general don't check slices for nil, always use len, since nil slice is equivalent to empty slice. Only check for nils of elements of a slice, not the slice itself.

corverroos (approved):

PR #3139, opened 2025-02-24
chore(sdk): quote refactor

kevinhalliday:
Refactor useQuote:

- use Quotable type, rather than Deposit / Expense
- explicit pending / loading / success 
- use request as query key (for caching)
- sepr async getter into own func

issue: none


sdk/src/hooks/useQuote.ts
```diff
@@ -1,132 +1,144 @@
 import { type UseQueryResult, useQuery } from '@tanstack/react-query'
 import { useMemo } from 'react'
-import { type Address, fromHex, toHex, zeroAddress } from 'viem'
-import type { Deposit, Expense } from '../types/order.js'
+import { type Address, fromHex, zeroAddress } from 'viem'
+import { toJSON } from './util.js'
+
+type Quoteable = { isNative: boolean; token?: Address; amount: bigint }
```
PaulLeCam, line 6: IIRC what we discussed last week, I think this type could be refined to ensure the `token` address is always provided when `isNative` is `false`
```suggestion
type Quoteable =
  | { isNative: true; token?: never; amount: bigint }
  | { isNative: false; token: Address; amount: bigint }
```

sdk/src/hooks/useQuote.ts
```diff
@@ -1,132 +1,144 @@
...
 
 type UseQuoteParams = {
-  srcChainId?: number
+  srcChainId: number
```
ga-reth, line 9: lets keep srcChainId as optional, since chainId isn't always available as it's often fetched via `useAccount` from wagmi which returns optional chainId, I have code that just disables the query if this value isn't defined

PR #3143, opened 2025-02-24
chore(sdk): hook tweaks

ga-reth:
- remove slicing as api expects complete call data
- check query key
- disable queries if amounts 0
- deposit and expense amounts should be optional, since only 1 of them should be defined - will add narrowing

issue: none

### DIFF
--- a/sdk/src/hooks/useQuote.ts
+++ b/sdk/src/hooks/useQuote.ts
@@ -4,8 +4,8 @@ import { type Address, fromHex, zeroAddress } from 'viem'
 import { toJSON } from './util.js'
 
 type Quoteable =
-  | { isNative: true; token?: never; amount: bigint }
-  | { isNative: false; token: Address; amount: bigint }
+  | { isNative: true; token?: never; amount?: bigint }
+  | { isNative: false; token: Address; amount?: bigint }
 
 type UseQuoteParams = {
   srcChainId?: number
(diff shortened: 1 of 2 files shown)


sdk/src/hooks/useQuote.ts
```diff
@@ -48,51 +49,41 @@ type Quote = {
...
-    p.deposit.amount,
-    p.deposit.isNative,
-    p.deposit.token,
-    p.expense.amount,
-    p.expense.isNative,
-    p.expense.token,
-  ])
-
-  const enabled = !p.srcChainId || p.enabled
+  }, [params])
```
kevinhalliday, line 63: i'd rather not enforce people memoize their params, when we can memoize request by literals for them

sdk/src/hooks/useQuote.ts
```diff
@@ -48,51 +49,41 @@ type Quote = {
...
+  const response = await fetch(`${apiBaseUrl} + /quote`, {
     method: 'POST',
     headers: { 'Content-Type': 'application/json' },
     body: request,
   })
 
-  if (!response.ok) {
-    const { error } = await response.json()
-    throw error as QuoteError
-  }
```
kevinhalliday, line 94: don't we want react query to handle erros for us?

kevinhalliday (approved):

sdk/src/hooks/useQuote.ts
ga-reth: react query will handle request errors for us

sdk/src/hooks/useQuote.ts
ga-reth: we can just memoize the params object itself, but added child objs explicitly here to account shallow compare

PR #3144, opened 2025-02-24
chore(sdk): get order and inbox status hooks

kevinhalliday:
Seperate useGetOrder and useInboxStatus hooks.

issue: none


sdk/src/hooks/useGetOrder.ts
```diff
@@ -0,0 +1,25 @@
+import type { Hex, Address } from 'viem'
+import { useReadContract } from 'wagmi'
+import { inbox } from '../constants/contracts.js'
+
+export function useGetOrder({
```
ga-reth, line 5: we already have this in `useOrderStatus` btw

ga-reth (approved):

PR #3149, opened 2025-02-24
feat(sdk): add omni context

kevinhalliday:
Add omni context:

- track api per env
- query contract addrs from api
- use defaults to avoid loading states

issue: none


sdk/src/constants/abis.ts
```diff
@@ -0,0 +1,189 @@
...
+          {
+            name: 'status',
+            internalType: 'enum ISolverNetInbox.Status',
+            type: 'uint8',
+          },
+          { name: 'timestamp', internalType: 'uint32', type: 'uint32' },
+          { name: 'claimant', internalType: 'address', type: 'address' },
+        ],
+      },
+    ],
```
kevinhalliday, line 91: Just copied existing abi, but this is out of date. once merged will update

sdk/src/context/omni.tsx
```diff
@@ -0,0 +1,132 @@
+import { createContext, useContext, useMemo } from 'react'
+import { useQuery } from '@tanstack/react-query'
+import { fetchJSON } from '../internal/api.js'
+import { throwingProxy } from '../internal/util.js'
+import type { Address } from 'viem'
+
+type Environment = 'devnet' | 'staging' | 'testnet' | 'mainnet'
```
ga-reth: can we change testnet to omega? is causing some confusion

ga-reth (approved):

PR #3165, opened 2025-02-25
ci(e2e): allow val3 and val4 on mainnet

corverroos:
Adds val3 and val4 to mainnet allow list.

issue: none


cli/cmd/staking.go
```diff
@@ -112,13 +112,13 @@ func (v EOAConfig) validate() error {
 	return nil
 }
 
-type createValConfig struct {
+type CreateValConfig struct {
```
corverroos, line 115: made public for scripting, similar to Delegate

Issue #3177, opened 2025-02-25
Add e2e test cases for different solvernet valid and invalid orders


powerslider:
#### Problem to Solve
We need more test coverage regarding types of SolverNet orders.

#### Proposed Solution
Add test cases covering different types of valid or invalid orders to check correct handling.

PR #3209, opened 2025-02-27
test(e2e/solve): add more reject reasons test cases

powerslider:
Covers following reject reasons:

- SameChain
- UnsupportedDeposit
- UnsupportedSrcChain

issue: #3177

### DIFF
--- a/e2e/solve/test.go
+++ b/e2e/solve/test.go
@@ -40,19 +40,19 @@ type TestOrder struct {
 	RejectReason  string
 }
 
-func (o TestOrder) IsDepositTokenInvalid() bool {
+func isDepositTokenInvalid(o TestOrder) bool {
 	return o.Deposit.Token == invalidTokenAddress
 }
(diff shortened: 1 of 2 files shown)


chmllr (approved):

e2e/solve/test.go
```diff
@@ -40,6 +40,22 @@ type TestOrder struct {
 	RejectReason  string
 }
 
+func (o TestOrder) IsDepositTokenInvalid() bool {
+	return o.Deposit.Token == invalidTokenAddress
```
corverroos, line 44: nit: this is very OOP style, rather make these normal functions. In general try to avoid linking state with logic. try decoupling state from logic.

e.g.
```
func isDepositTokenInvalid(o TestOrder) bool {
	return o.Deposit.Token == invalidTokenAddress
}
```

corverroos (approved):

PR #3214, opened 2025-02-27
chore(sdk): setup Vitest

PaulLeCam:
Basic setup of [Vitest](https://vitest.dev/) to run tests. I tried also bundling with Vite but it seems only meant for browser use, bundling all the dependencies rather than using package imports, so it's probably not ideal for most apps that bundle their code anyway and would benefit from tree-shaking.

issue: https://github.com/omni-network/ops/issues/951

### DIFF
--- a/sdk/biome.json
+++ b/sdk/biome.json
@@ -7,7 +7,7 @@
   },
   "files": {
     "ignoreUnknown": false,
-    "ignore": ["dist", "tsconfig.json", "tsconfig.*.json"]
+    "ignore": ["coverage", "dist", "tsconfig.json", "tsconfig.*.json"]
   },
   "formatter": {
     "enabled": true,
--- a/sdk/package.json
+++ b/sdk/package.json
@@ -43,7 +43,6 @@
   "devDependencies": {
     "@biomejs/biome": "1.9.4",
     "@types/react": "^19.0.8",
-    "@vitejs/plugin-react-swc": "^3.8.0",
     "@vitest/coverage-v8": "^3.0.7",
     "typescript": "^5.7.2",
     "vite": "^6.2.0",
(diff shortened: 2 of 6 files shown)


sdk/biome.json
```diff
@@ -7,7 +7,7 @@
   },
   "files": {
     "ignoreUnknown": false,
-    "ignore": ["dist", "tsconfig.json", "tsconfig.*.json"]
+    "ignore": ["coverage", "dist", "tsconfig.json", "tsconfig.*.json"]
```
ga-reth, line 10: do we need to ignore coverage from tsconfig.build.json also?

sdk/package.json
```diff
@@ -42,7 +43,11 @@
   "devDependencies": {
     "@biomejs/biome": "1.9.4",
     "@types/react": "^19.0.8",
-    "typescript": "5.7.2"
+    "@vitejs/plugin-react-swc": "^3.8.0",
```
ga-reth, line 46: curious, why do we need to use this plugin? have seen other uses of vitest w/ react without an addt'l plugin

PaulLeCam: AFAIK Vite needs a plugin to support React, it doesn't do it out of the box.

sdk/biome.json
PaulLeCam: Is it using the same folder? If so, maybe we can configure it differently, but do we need to format coverage files?

ga-reth: I meant excluding coverage from build output, but we're including from `"include": ["src/**/*.ts"],` so all good

sdk/package.json
PaulLeCam: Indeed it works without, I removed it

ga-reth (approved):

PR #3216, opened 2025-02-27
chore(sdk): rename hook and update pckg vers

ga-reth:
- rename hook
- match version in package.json to published

issue: none


PaulLeCam (approved):

sdk/src/hooks/useOrder.ts
```diff
@@ -51,7 +51,7 @@ export function useOrder<abis extends OptionalAbis>(
   const txMutation = useWriteContract()
   const wait = useWaitForTransactionReceipt({ hash: txMutation.data })
 
-  const { orderId, originData } = useGetOpenOrder({
+  const { orderId, originData } = useGetOrderData({
     status: wait.status,
     logs: wait.data?.logs,
   })
```
kevinhalliday, line 57: `OrderData` has some meaning in ERC7683 already. it's the order we encode to open the order. and is not actually what we are getting here.

here we are passing orderId and fillOriginData from an `Open` event.

Thoughts on this?

```typescript
// return full resolved order, rather then first element of fillInstructions (which we call fillOrginData)
cosnt { 
    orderId,    // bytes32 orderid 
    resolved,  // ERC7683.ResolvedCrossChainOrder
 } = useParseOpen({ status, logs })
 
 // then pass full resolved order into did fill
const { filled } =  useDidFill({ orderId, resolved })

// then in did fill we can parse out the info from resolved we need 
const fillOriginData = resolved.fillInstructions[0].originData
```

That way we interfaces to `useDidFill` and `useParseOpen` don't need to change if we add more / change resolved fill instructions in contracts 

other options for `useParseOpen`:

`useParseResolved`
`useOrderOpen`
`useParseOpenEvent`
`useParseOrderOpen`

something that says "we are taking logs from an open tx and getting order id / resolved order"

ga-reth: good point, it's specifically parsing the open event so I think some flavour of that sounds ideal

Issue #3217, opened 2025-02-27
Introduce basic error propagation to the SDK


ga-reth:
#### Problem to Solve

Right now there are some places in the hooks where we fail silently

#### Proposed Solution

Introduce some basic error propagation to be consumed

PR #3218, opened 2025-02-27
feat(sdk): introduce basic error propagation

ga-reth:
- introduce basic error propagation
- these can be improved to narrow reasons etc, but keeping it simple for now

issue: #3217

### DIFF
--- a/sdk/src/hooks/useDidFill.ts
+++ b/sdk/src/hooks/useDidFill.ts
@@ -24,5 +24,5 @@ export function useDidFill(params: UseDidFillParams) {
     },
   })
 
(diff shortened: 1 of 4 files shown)


sdk/src/hooks/useDidFill.ts
```diff
@@ -24,5 +24,5 @@ export function useDidFill(params: UseDidFillParams) {
     },
   })
 
-  return filled.data ?? false
+  return filled ?? false
```
kevinhalliday, line 27: probably just `return filled` here no?

kevinhalliday (approved):

PR #3219, opened 2025-02-27
chore(sdk): add readme

ga-reth:
- add readme

issue: none


sdk/README.md
```diff
@@ -1,17 +1,149 @@
...
+
+const queryClient = new QueryClient()   
+
+<WagmiProvider>
+  <QueryClientProvider client={queryClient}>
+    <OmniProvider env="testnet">
+      <App />
+    </OmniProvider>
+  </QueryClientProvider>
+</WagmiProvider>
```
PaulLeCam, line 41: People often simply copy and paste these kinds of code example so it would be safer to make sure they works as-is when possible, maybe just making sure these elements are wrapped in a component?

```suggestion
function Providers({ children }) {
  return (
    <WagmiProvider>
      <QueryClientProvider client={queryClient}>
        <OmniProvider env="testnet">
          {children}
        </OmniProvider>
      </QueryClientProvider>
    </WagmiProvider>
  )
}

File: sdk/README.md [hunk 2]
```diff
@@ -1,17 +1,149 @@
...
+- Deposit reflects the amount to be spent on the source chain.
+- Expenses reflects the amount to be used for the calls on the destination chain.
+- In this simple bridge, expense is simply the amount sent to the user (deposit - fees).
+- `useOrder` is used to create an order. An open method is returned to asynchronously open an order.
+- Validation reflects another call to our Solver API to validate the order, this can be helpful for verifying an order before opening it.
+- `error` will give you information about any potential issues.
+- `status` will give you the status of the order:
+
+```tsx
+export type OrderStatus =
```
PaulLeCam, line 135: Seems this type should be exported by the SDK, no?

sdk/README.md
```diff
@@ -1,17 +1,149 @@
...
-- [✅] Publish alpha package
-- [✅] Consume SolverNet API
-- [ ] Type improvements
-- [ ] Add unit tests
-- [ ] Add e2e tests
-- [ ] Zod schema validation
-- [ ] Write docs
-- [ ] Build package abstractions - `core` and `react`
-- [ ] Add examples
+You can reach out to us on telegram with any queries, feedback, or requests: [@omnidevsupport](https://t.me/omnidevsupport)
```
PaulLeCam, line 149: Wouldn't GitHub issues and possibly discussions be a simpler way for devs to interact?

PaulLeCam (approved):
LGTM, don't hesitate to ignore the comments

sdk/README.md
ga-reth: this is just a snippet so will leave it as is!

sdk/README.md
ga-reth: for sure, that can be inferred though right

sdk/README.md
ga-reth: and good point re testing it, have done so and all good

PR #3223, opened 2025-02-28
refactor(solver/app): refactor http handlers and types

corverroos:
Refactor solver API handlers:
 - single `handlerAdapator` that does all JSON (un)marshalling for all endpoints.
 - Decouple JSON errors from handler response types. Logic now returns struct and errors (standard go).

issue: none

### DIFF
--- a/solver/app/handler.go
+++ b/solver/app/handler.go
@@ -1,0 +1,153 @@
+package app
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/omni-network/omni/lib/contracts"
+	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/log"
+	"github.com/omni-network/omni/solver/types"
+)
+
+func newAPIError(err error, statusCode int) APIError {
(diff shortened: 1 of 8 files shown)


solver/app/handler.go
```diff
@@ -0,0 +1,153 @@
...
+type APIError struct {
+	Err        error
+	StatusCode int
+}
+
+func (e APIError) Error() string {
+	return e.Err.Error()
+}
+
+type Handler struct {
```
corverroos, line 30: handler abstracts an endpoint and logic

solver/app/quote.go
```diff
@@ -17,26 +17,20 @@ type (
...
 	QuoteUnit     = types.QuoteUnit
 
-	quoteFunc func(QuoteRequest) QuoteResponse
+	quoteFunc func(context.Context, QuoteRequest) (QuoteResponse, error)
 )
 
 // quoter is quoteFunc that can be used to quoter an expense or deposit.
 // It is the logic behind the /quoter endpoint.
-func quoter(req QuoteRequest) QuoteResponse {
+func quoter(_ context.Context, req QuoteRequest) (QuoteResponse, error) {
```
corverroos, line 25: logic now returns types and errors, like normal go logic.

PR #3225, opened 2025-02-28
chore(sdk): setup integration tests

PaulLeCam:
Basic setup for testing the SDK in browser, only works locally at the moment with the devnet running (not setup for CI yet).

issue: none

### DIFF
--- a/sdk/integration-tests/flows.test.ts
+++ b/sdk/integration-tests/flows.test.ts
@@ -1,0 +1,10 @@
+import { expect, test } from 'vitest'
+
+import {
+  ContextProvider,
+  MOCK_L1_ID,
+  MOCK_L2_ID,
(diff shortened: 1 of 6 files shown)


ga-reth:
thoughts on renaming `integration-tests` to `tests`? since we can share the utils / other setup with unit tests, so this directory will house integration tests as well as other logic for each flavour of tests?

ga-reth (approved):

PaulLeCam:
> thoughts on renaming `integration-tests` to `tests`? since we can share the utils / other setup with unit tests, so this directory will house integration tests as well as other logic for each flavour of tests?

I think the test utils can live in another folder if they need to be shared, but the test suites themselves should have a dedicated folder so we can scope the test runner, ex using `pnpm test -- integration-tests`

PR #3227, opened 2025-02-28
WIP: OpenOrder simulation

chmllr:
Body goes here...

issue: none or #12345

### DIFF
--- a/monitor/flowgen/bridging/bridging.go
+++ b/monitor/flowgen/bridging/bridging.go
@@ -1,6 +1,7 @@
 package bridging
 
 import (
+	"fmt"
 	"math/big"
 	"time"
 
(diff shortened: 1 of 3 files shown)


monitor/flowgen/bridging/bridging.go
```diff
@@ -31,7 +32,7 @@ func NewJob(
 	}
 
 	return types.Job{
-		Name:    "Bridging",
+		Name:    fmt.Sprintf("Bridging (%v->%v)", srcChain, dstChain),
```
corverroos, line 35: avoid using ints for chains, as humans struggle to read numbers, chain names make things a LOT easier to debug/read/understand and saves a lot of time.

monitor/flowgen/flowgen.go
```diff
@@ -46,22 +46,63 @@ func Start(
 
 	var jobs []types.Job
 
-	// Add jobs according to the current network ID
 	switch network.ID {
 	case netconf.Devnet:
-		// Instantiate bridging of native ETH on a devnet
+		// Bridging of native ETH on a devnet
+		amount := big.NewInt(0).Mul(util.MilliEther, big.NewInt(2)) // 0.002 ETH
```
corverroos, line 52: lets reduce the duplication, since we will be doing this for all networks?

PR #3231, opened 2025-02-28
test(e2e/solve): add InsufficientInventory test

powerslider:
- Add test order for InsufficientInventory test case.
- Introduce solver account balance drain/refund in order to simulate InsufficientInventory order rejection.

issue: #3177


e2e/solve/fund.go
```diff
@@ -16,39 +19,92 @@ import (
 	"cosmossdk.io/math"
 )
 
-func maybeFundSolver(ctx context.Context, network netconf.ID, backends ethbackend.Backends) error {
+// erc20 tokens to fund solver with on devnet. useful for solvernet development when forking public networks.
```
corverroos, line 22: golang comments start with function or variable name, e.g.: `// testTokenAddressesToFund are the tokens ...`

e2e/solve/fund.go
```diff
@@ -16,39 +19,92 @@ import (
...
+
 		// if devnet is not forking the public chain tkn is deployed on, this sets
 		// storage on an unused address, which is fine
-		err := anvil.FundERC20(ctx, ethCl, tkn.addr, eth1m, solver)
-		if err != nil {
+		if err := anvil.FundERC20(ctx, ethCl, tkn.addr, eth1m, solver); err != nil {
 			return errors.Wrap(err, "fund tkn failed", "chain_id", tkn.chainID, "addr", tkn.addr)
 		}
+
+		if err := checkAccountBalance(ctx, ethCl, solver, tkn.addr, tkn.chainID); err != nil {
```
corverroos, line 55: not sure why you check native balance after doing token funding (not native funding)?

e2e/solve/test.go
```diff
@@ -383,12 +399,36 @@ func openOrder(ctx context.Context, backends ethbackend.Backends, order TestOrde
...
 
-func testCheckAPI(ctx context.Context, orders []TestOrder) error {
+func testCheckAPI(ctx context.Context, backends ethbackend.Backends, orders []TestOrder) error {
 	const url = "http://localhost:26661/api/v1/check"
 
 	var eg errgroup.Group
-	for i, order := range orders {
+	var balanceChangingInProgress bool // Tracks if a solver account drain/refund cycle is running.
+	var refundMutex sync.Mutex
+	balanceRestored := sync.NewCond(&refundMutex)
```
corverroos, line 408: try to avoid complex async logic, this is very hard to follow and will probably flap. Rather make it synchronous and remove this.

e2e/solve/fund.go
```diff
@@ -16,39 +19,92 @@ import (
...
+			return errors.Wrap(err, "get solver account balance failed")
+		}
+	}
+
+	return nil
+}
+
+// maybeDrainSolverAccount drains solver account balance to 0.
+func maybeDrainSolverAccount(ctx context.Context, network netconf.ID, backends ethbackend.Backends) error {
```
corverroos, line 64: avoid functions that don't do much, especially if only used in one or two places. You can just call `maybeSetSolverAccountBalance` directly in the code below.

e2e/solve/fund.go
```diff
@@ -16,39 +19,92 @@ import (
...
+}
+
+// maybeFundSolverAccount funds/refunds solver account balance to 1_000_000 ETH.
+func maybeFundSolverAccount(ctx context.Context, network netconf.ID, backends ethbackend.Backends) error {
+	eth1m := math.NewInt(1_000_000).MulRaw(params.Ether).BigInt()
+	return maybeSetSolverAccountBalance(ctx, network, backends, eth1m)
+}
+
+// maybeSetSolverAccountBalance calls anvil_setBalance to set the solver account to a passed amount.
+func maybeSetSolverAccountBalance(ctx context.Context, network netconf.ID, backends ethbackend.Backends, amt *big.Int) error {
```
corverroos, line 75: not sure the network check (and therefore `maybe`)  is required. This code is only executed on e2e tests which is only devnet right?

e2e/solve/fund.go
corverroos: this type of refactor is subjective, please try to avoid subjective refactors.

corverroos (changes requested):

e2e/solve/test.go
```diff
@@ -384,71 +401,81 @@ func openOrder(ctx context.Context, backends ethbackend.Backends, order TestOrde
...
-				FillDeadline:       uint32(order.FillDeadline.Unix()), //nolint:gosec // this is fine for tests
-				SourceChainID:      order.SourceChainID,
-				DestinationChainID: order.DestChainID,
-				Expenses:           expensesFromBindings(order.Expenses),
-				Calls:              callsFromBindings(order.Calls),
-				Deposit:            addrAmtFromDeposit(order.Deposit),
+		// If this order requires balance draining, do it before test logic.
+		if isInsufficientInventory(order) {
+			// Drain solver native balance.
+			if err := setSolverAccountNativeBalance(ctx, backends, big.NewInt(0)); err != nil {
```
corverroos, line 411: I would also pass in the chain on which the balance should be set, in this case `DestinationChainID` (not hard coded like now)

e2e/solve/test.go
```diff
@@ -384,71 +401,81 @@ func openOrder(ctx context.Context, backends ethbackend.Backends, order TestOrde
...
 
-	if err := eg.Wait(); err != nil {
-		return errors.Wrap(err, "wait checks")
+		// Refund solver native balance after test logic.
+		if isInsufficientInventory(order) {
+			eth1m := math.NewInt(1_000_000).MulRaw(params.Ether).BigInt()
+			if err := setSolverAccountNativeBalance(ctx, backends, eth1m); err != nil {
+				return errors.Wrap(err, "refund solver account failed")
+			}
+			log.Debug(ctx, "Solver account native balance refunded successfully")
```
corverroos, line 477: I would remove log, it feels like debug logging only required for you adding the test.

e2e/solve/fund.go
```diff
@@ -52,3 +54,27 @@ func maybeFundSolver(ctx context.Context, network netconf.ID, backends ethbacken
...
+	if err := anvil.FundAccounts(ctx, ethCl, amt, solver); err != nil {
+		return errors.Wrap(err, "set solver account balance failed")
+	}
+
+	balance, err := ethCl.BalanceAt(ctx, solver, nil)
+	if err != nil {
+		return errors.Wrap(err, "get account balance failed", "chain_id", chainID)
+	}
+
+	log.Debug(ctx, "Current Solver Balance", "balance", balance.String(), "chain_id", chainID)
```
corverroos, line 77: I would also remove these logs, feels only required for you debugging the test

corverroos (approved):

chmllr (approved):

Issue #3170, opened 2025-02-25
Add unit tests to the SDK (vitest)


ga-reth:
#### Problem to Solve

We don't have any unit tests currently

#### Proposed Solution

Add unit tests where applicable using vitests

PR #3241, opened 2025-03-03
chore(sdk): initial mocks and unit setup file

ga-reth:
- optional mock /contracts call
- unit tests setup module

issue: #3170

### DIFF
--- a/sdk/test/react.ts
+++ b/sdk/test/react.ts
@@ -7,11 +7,12 @@ import {
 import { createElement } from 'react'
 import { WagmiProvider, type createConfig } from 'wagmi'
 import { OmniProvider } from '../src/context/omni.js'
+import { mockContractsQuery } from './mocks.js'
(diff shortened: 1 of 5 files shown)


sdk/test/react.ts
```diff
@@ -37,10 +38,21 @@ export function createWrapper<TComponent extends React.FunctionComponent<any>>(
 
 export function renderHook<Result, Props>(
   render: (props: Props) => Result,
+  options?: RenderHookOptions<Props> & {
+    mockContractsCall?: boolean
+  },
   env: 'devnet' | 'testnet' = 'devnet',
```
PaulLeCam, line 44: Maybe `env` could be part of the `options` object rather than a separate argument?

PaulLeCam (approved):

sdk/test/react.ts
ga-reth: good point, will update

PR #3253, opened 2025-03-03
feat(sdk): return complete resolved order

ga-reth:
- return complete resolved order

issue: none

### DIFF
--- a/sdk/src/hooks/useDidFill.ts
+++ b/sdk/src/hooks/useDidFill.ts
@@ -1,11 +1,11 @@
 import { useReadContract } from 'wagmi'
 import { outboxABI } from '../constants/abis.js'
 import { useOmniContext } from '../context/omni.js'
-import type { ResolvedOrder } from '../types/order.js'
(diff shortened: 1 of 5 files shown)


sdk/src/hooks/useDidFill.ts
```diff
@@ -1,25 +1,26 @@
...
+  const { resolvedOrder, destChainId } = params
   const { outbox } = useOmniContext()
   const filled = useReadContract({
     chainId: destChainId,
     address: outbox,
     abi: outboxABI,
     functionName: 'didFill',
-    args: orderId && originData ? [orderId, originData] : undefined,
+    args: resolvedOrder
+      ? [resolvedOrder.orderId, resolvedOrder.fillInstructions[0].originData]
```
kevinhalliday, line 20: Perhaps a sanity check on `fillInstructions[0]` being defined?

kevinhalliday (approved):

sdk/src/hooks/useDidFill.ts
ga-reth: good point, will ad

PR #3261, opened 2025-03-04
chore(docs): add order data field description

chmllr:
Describes the semantics of the order data fields.

issue: none


contracts/solve/src/lib/SolverNet.sol
```diff
@@ -2,11 +2,22 @@
 pragma solidity =0.8.24;
 
 library SolverNet {
+    // OrderData describes an intent for a cross-chain action between a source and a destination chain.
     struct OrderData {
         address owner;
         uint64 destChainId;
+        // deposit contains the token amount `X + fee` with `X` being the amount
```
corverroos, line 9: lets add the comments to the type definitions, not to field definitions.

chmllr: Done

kevinhalliday (approved):

Issue #3057, opened 2025-02-13
Add gasless order support


Zodomo:
#### Problem to Solve
ERC-7683 spec allows for orders to be opened gaslessly. We haven't implemented this yet.

#### Proposed Solution
Implement permit2 and gasless orders.

PR #3272, opened 2025-03-04
feat(contracts): initial gasless order support

Zodomo:
Added initial support for gasless SolverNet orders. This set of changes is just the MVP. It does not allocate any part of the user's assets to the relayer at open nor fill. This implementation doesn't reimburse the relayer at all, and only lets them open signed orders for users.

issue: #3057


Zodomo:
Closing this as I will need to redo this after the changes we've made to SolverNet.

Issue #3192, opened 2025-02-26
Unit test ensuring the solver status enums match go code


corverroos:
#### Problem to Solve
The solidity solver inbox status enum has implicit int values. In go, we manually assign those values. Everything will break if a regression is introduces by changing either the solidity or the goland.

#### Proposed Solution
Write a golang unit test that ensure the solidity enums match the go enums.

e.g.:
 - read the solidtyinbox.sol file
 - skip all lines until `enum Status {`
 - ensure subsequent lines match the go enums   
 - ensure `}` is next line.

PR #3276, opened 2025-03-05
chore(contracts/solvernet): generate status enum bindings

corverroos:
Introduce a naive tool `solenumgen` to generate golang enum bindings (which abigen doesn't do). Run this script for OrderStatus only for now.

issue: #3192


scripts/solenumgen/main.go
```diff
@@ -0,0 +1,180 @@
...
+	"golang.org/x/tools/imports"
+	"os"
+	"strings"
+	"text/template"
+)
+
+var (
+	flagInput   = flag.String("input", "", "input sol file path to generate the enums from")
+	flagPackage = flag.String("package", "", "package name to generate the enum in")
+	flagOutput  = flag.String("output", "", "output file path to write generated codes")
```
chmllr, line 20: Not sure if plural was intended
```suggestion
	flagOutput  = flag.String("output", "", "output file path to write generated code")
```

powerslider (approved):

chmllr (approved):
Bit: I would add at least one simple test just to capture the current functionality.

PR #3291, opened 2025-03-05
chore(sdk): units and require generic on mock

ga-reth:
- useInboxStatus units
- require generic on `createMockReadContractResult` to restrict input type of data to match expected

issue: #3170

### DIFF
--- a/sdk/.gitignore
+++ b/sdk/.gitignore
@@ -13,3 +13,6 @@ coverage
 
 # asdf
 .tool-versions
+
+# vitest
+tsconfig.vitest-temp.json
--- a/sdk/src/hooks/useInboxStatus.test.ts
+++ b/sdk/src/hooks/useInboxStatus.test.ts
@@ -1,0 +1,134 @@
+import { waitFor } from '@testing-library/react'
+import { beforeEach, expect, test, vi } from 'vitest'
+import { orderId, renderHook, resolvedOrder } from '../../test/index.js'
+import { createMockReadContractResult } from '../../test/mocks.js'
+import { useGetOrder } from './useGetOrder.js'
+import { useInboxStatus } from './useInboxStatus.js'
+
+const data = { status: 1, claimant: '0x123', timestamp: 1 } as const
+
+const { mockUseGetOrder } = vi.hoisted(() => {
+  return {
+    mockUseGetOrder: vi.fn().mockImplementation(() => {
+      return createMockReadContractResult()
+    }),
+  }
+})
+
+vi.mock('./useGetOrder.js', async () => {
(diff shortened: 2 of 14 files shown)


sdk/.gitignore
```diff
@@ -13,3 +13,6 @@ coverage
 
 # asdf
 .tool-versions
+
+# vitest
+tsconfig.vitest-temp.json
```
PaulLeCam, line 18: Do we need this?

sdk/src/hooks/useInboxStatus.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    ...actual,
+    useGetOrder: mockUseGetOrder,
+  }
+})
+
+beforeEach(() => {
+  mockUseGetOrder.mockReturnValue(createMockReadContractResult())
+})
+
+test('default', async () => {
```
PaulLeCam, line 30: What does "default" mean here please? Any way we could make the test name more explicit?

sdk/src/hooks/useInboxStatus.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    mockContractsCall: true,
+  })
+
+  expect(result.current).toBe('open')
+})
+
+test('parameters: status rejected', () => {
+  mockUseGetOrder.mockReturnValue(
+    createMockReadContractResult<ReturnType<typeof useGetOrder>>({
+      data: [resolvedOrder, { ...data, status: 2 }],
```
PaulLeCam, line 83: Could we use named constants for the status rather than magic numbers please?

sdk/src/hooks/useInboxStatus.test.ts
ga-reth: default - basically just happy or expected case

but open to suggestions!

sdk/src/hooks/useInboxStatus.test.ts
ga-reth: the enum from the contract is returned by an integer, we can improve that in a follow up

sdk/src/hooks/useInboxStatus.test.ts
PaulLeCam: Maybe we could simply describe what is the expected default behavior, ex "returns the inbox status"?

PaulLeCam (approved):

sdk/src/hooks/useInboxStatus.test.ts
ga-reth: sounds good, let's do that going forward

Issue #3193, opened 2025-02-26
SolverNet audit


fabtreb:
commit [3ef37ae](https://github.com/omni-network/omni/commit/3ef37ae2704db4e1dcb8fc3f7acbb8babb543e8c)

PR #3296, opened 2025-03-05
feat(contracts): applied feedback from solvernet audit

Zodomo:
Applied feedback from the SolverNet audit. Addressed vulnerabilities and changes such as the reentrancy bug on `markFulfilled()`, gas estimation not accounting for memory expansion, OmniPortal pause status checks, two step token approvals, unbounded arrays, state bloat, revoke approval revert case, unnecessary initializer in middleman, and possible out of bounds array reads. Some audit findings were not included due to not being applicable.

issue: #3193


contracts/solve/src/SolverNetInbox.sol
```diff
@@ -541,6 +556,25 @@ contract SolverNetInbox is OwnableRoles, ReentrancyGuard, Initializable, Deploye
...
+        if (status == Status.Pending) {
+            return;
+        } else {
+            if (status != Status.Filled) delete _orderDeposit[id];
+            if (status != Status.Claimed) {
+                delete _orderHeader[id];
+                delete _orderCalls[id];
+                delete _orderExpenses[id];
+            }
+        }
```
kevinhalliday, line 577: no need for top level if else block
```solidity
    function _purgeState(bytes32 id, Status status) internal {
        if (status == Status.Pending) return;
        if (status != Status.Filled) delete _orderDeposit[id];
        if (status != Status.Claimed) {
            delete _orderHeader[id];
            delete _orderCalls[id];
            delete _orderExpenses[id];
        }
    }
```

kevinhalliday (approved):
👏

PR #3313, opened 2025-03-06
chore(e2e): test various solvernetspec configs

Zodomo:
Testing various `solvernetspec` configs to ensure the pause mechanism works.

issue: none

### DIFF
--- a/e2e/app/admin/solvernetspec.go
+++ b/e2e/app/admin/solvernetspec.go
@@ -20,9 +20,8 @@ import (
 // To update SolverNet spec, update this map.
 // Then run `ensure-solvernet-spec` to apply the changes.
 var solverNetSpec = map[netconf.ID]NetworkSolverNetSpec{
-	netconf.Devnet:  {Global: DefaultSolverNetSpec()},
-	netconf.Staging: {Global: DefaultSolverNetSpec()},
-	netconf.Omega: {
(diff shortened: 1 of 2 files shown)


e2e/app/admin/solvernetspec.go
```diff
@@ -21,7 +22,24 @@ import (
...
+		ChainOverrides: map[uint64]*SolverNetSpec{
+			evmchain.IDBaseSepolia: {
+				PauseAll:   false,
+				PauseOpen:  false,
+				PauseClose: true,
+			},
+			evmchain.IDOpSepolia: func() *SolverNetSpec {
+				spec := DefaultSolverNetSpec()
+				return &spec
+			}(),
```
kevinhalliday, line 40: we don't need if we're just returning default

kevinhalliday (commented):
let's test on staging instead

e2e/app/admin/solvernetspec.go
kevinhalliday: oh i see cause global isn't default

kevinhalliday (approved):

e2e/app/admin/solvernetspec.go
Zodomo: I was just trying to test overriding the global spec with the default. was getting linter errors otherwise.

image

PR #3327, opened 2025-03-08
chore(lib/xchain): add support for head-min-1 conf level

corverroos:
Add support for `xchain.ConfMin1` which lags behind head by 1 block.

issue: none

### DIFF
--- a/solver/app/app.go
+++ b/solver/app/app.go
@@ -32,14 +32,13 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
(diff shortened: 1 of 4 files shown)


solver/app/app.go
```diff
@@ -239,6 +238,20 @@ func startEventStreams(
 			return err
 		}
 
+		// Logic to query block timestamp for a given height
+		inboxTimestamps[chain] = func(height uint64) time.Time {
```
corverroos, line 242: not directly related, but fixing in any case, this must be moved up.

PR #3331, opened 2025-03-10
chore(sdk): add useOmniContracts hook

PaulLeCam:
Extracts the contract addresses from the Omni context to ensure they are always loaded from the API.
This affects other hooks and APIs that need to handle loading the contracts addresses.

issue: https://github.com/omni-network/omni/issues/3332

### DIFF
--- a/sdk/src/hooks/useOmniContracts.ts
+++ b/sdk/src/hooks/useOmniContracts.ts
@@ -1,38 +1,12 @@
 import { type UseQueryResult, useQuery } from '@tanstack/react-query'
-import type { Address } from 'viem'
+
 import { useOmniContext } from '../context/omni.js'
-import { fetchJSON } from '../internal/api.js'
-
-export type OmniContracts = {
-  inbox: Address
-  outbox: Address
-  middleman: Address
-}
-
-async function getContracts(apiBaseUrl: string) {
-  const json = await fetchJSON(`${apiBaseUrl}/contracts`)
-
-  if (!isContracts(json)) throw new Error('Unexpected /contracts response')
-
-  return json
-}
-
-function isContracts(json: unknown): json is OmniContracts {
-  const contracts = json as OmniContracts
-  return (
-    contracts != null &&
-    typeof contracts.inbox === 'string' &&
-    typeof contracts.outbox === 'string' &&
-    typeof contracts.middleman === 'string'
-  )
-}
+import type { OmniContracts } from '../types/contracts.js'
+import { getOmniContractsQueryOptions } from '../utils/getContracts.js'
 
 export type UseOmniContractsResult = UseQueryResult<OmniContracts>
 
 export function useOmniContracts(): UseOmniContractsResult {
-  const { apiBaseUrl, env } = useOmniContext()
-  return useQuery({
-    queryKey: ['contracts', env],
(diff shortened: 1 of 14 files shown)


sdk/src/utils/withExecAndTransfer.ts
```diff
@@ -39,14 +39,13 @@ type WithExecAndTransferParams = {
...
 ): Call<typeof middlemanABI> {
-  const { middleman } = useOmniContext()
   const { call, transfer } = params
   const _callData = encodeFunctionData({ ...call })
 
   return {
     ...call,
     abi: middlemanABI,
-    target: middleman,
+    target: params.middlemanAddress,
```
kevinhalliday, line 48: thoughts on still using `useOmniContracts` here, so users don't need to pass in middleman address

sdk/src/hooks/useOmniContracts.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export type UseOmniContractsResult = UseQueryResult<OmniContracts>
+
+export function useOmniContracts(): UseOmniContractsResult {
+  const { apiBaseUrl, env } = useOmniContext()
+  return useQuery({
+    queryKey: ['contracts', env],
+    queryFn: () => getContracts(apiBaseUrl),
+  })
```
kevinhalliday, line 37: Can we [prefetch](https://tanstack.com/query/v4/docs/framework/react/guides/prefetching) this query in omni context

sdk/src/hooks/useOrder.ts
```diff
@@ -139,9 +156,17 @@ export function useOrder<abis extends OptionalAbis>(
     txMutation,
     waitForTx: wait,
   }
+  return status === 'initializing' || contractsResult.isError
+    ? {
+        ...orderReturn,
+        status: status === 'initializing' ? 'initializing' : 'error',
+        isReady: false,
+      }
```
kevinhalliday, line 164: We'll need to update docs in README with new behaviour, require that users check `isReady`

Other option is to be a little more forgiving in interface, and always return an `open` func, and just have that `open` func error (as it does now). Benefit here is that in most all cases, the contracts query will succeed before a user tries to call open. So we don't impose a ux on devs to handle that case explicitly. And instead, just error when it happens. Thoughts?

PaulLeCam: Yes the idea here is that consumers need to explicitly check for `isReady` so TS can statically check that `open` can be called.
As you described, we can alternatively always return the `open` function and in most cases it should work, but consumers that don't check for `isReady` will for sure have an error at runtime when `isReady` is `false`, and TS checks wouldn't help catch these.
I don't mind strongly either way, as you wrote the main thing here is having good docs on how to use the hook.

sdk/src/utils/withExecAndTransfer.ts
kevinhalliday: oh yeah it's a util function good call. keep it as is then yeah

sdk/src/hooks/useOrder.ts
PaulLeCam: Updated

kevinhalliday (approved):

PR #3337, opened 2025-03-10
refactor(solver): multiple devnet streamers

corverroos:
Add a second solver event processor to all devnet chains. 

This ensures two things:
1. Processor is indeed idempotent (since each block processed twice explicitly).
2. That we are ready to enable this in public chains if required. 

issue: none

### DIFF
--- a/solver/app/processor.go
+++ b/solver/app/processor.go
@@ -107,7 +107,7 @@ func newEventProcessor(deps procDeps, chainID uint64) xchain.EventLogsCallback {
 			switch event.Status {
 			case statusPending:
 				if alreadyFilled() {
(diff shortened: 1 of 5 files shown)


solver/app/processor.go
```diff
@@ -107,7 +107,7 @@ func newEventProcessor(deps procDeps, chainID uint64) xchain.EventLogsCallback {
 			switch event.Status {
 			case statusPending:
 				if alreadyFilled() {
-					return errors.New("unexpected already filled [BUG]")
```
corverroos, line 110: this is expected when reprocessing the same block

kevinhalliday (approved):

Issue #3279, opened 2025-03-05
Staking instrumentation service


chmllr:
#### Problem to Solve

Once we let users stake, we need some basic KPIs to analyze the staking behaviour.

#### Proposed Solution

Implement a new simple monitor module that would periodically query the stats from the consensus and execution chains and export them to Grafana.

PR #3341, opened 2025-03-10
fix(monitor): staking and supply metrics

chmllr:
Adds staking and supply metrics

issue: #3279


lib/cchain/queryutil/staking.go
```diff
@@ -99,7 +100,7 @@ func DelegatorInflationRates(ctx context.Context, cprov cchain.Provider, delegat
 // DelegationBalance represents the total delegation balance of a delegator.
 type DelegationBalance struct {
 	DelegatorAddress sdk.AccAddress
-	Balance          sdk.Coin
+	Balance          big.Int
```
kevinhalliday, line 103: i know it's a little weird but with `big.Int` you generally just work with pointers.

So this should be `Balance *big.Int`.

monitor/supply/supply.go
```diff
@@ -0,0 +1,93 @@
...
+	response, err := cprov.QueryClients().Bank.TotalSupply(ctx, &types.QueryTotalSupplyRequest{})
+	if err != nil {
+		return errors.Wrap(err, "total supply query")
+	}
+	cosmosSupplyWei := math.NewInt(0)
+	for _, coin := range response.Supply {
+		if coin.Denom == sdk.DefaultBondDenom {
+			cosmosSupplyWei = cosmosSupplyWei.Add(coin.Amount)
+		}
+	}
```
corverroos, line 54: nit: although this works, I'd reads like "supply" may contain multiple "bond" denoms which we need to sum, which isn't correct.

```
func stakeAmount(coins sdk.Coins) *big.Int {
	if len(coins) != 1 {
		return errors.new("unexpected number of coins")
	}
	ok, coin := resp.Supply.Find(sdk.DefaultBondDenom)
	if !ok {
		return errors.new("missing default bond denom")
	}
	return coin.Amount.BigInt()
}
```

monitor/supply/supply.go
```diff
@@ -0,0 +1,93 @@
...
+
+	cChainSupply.Set(toEtherF64(cosmosSupplyWei.BigInt()))
+
+	addrs, err := contracts.GetAddresses(ctx, network.ID)
+	if err != nil {
+		panic(err)
```
corverroos, line 60: avoid panics

monitor/supply/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+
+var (
+	cChainSupply = promauto.NewGauge(prometheus.GaugeOpts{
+		Namespace: "monitor",
+		Subsystem: "supply",
+		Name:      "cchain_total_ether",
+		Help:      "Token supply on the consensus chain in ether",
+	})
+
+	eChainSupply = promauto.NewGauge(prometheus.GaugeOpts{
```
corverroos, line 16: technically, we only call the `omni_evm` the "echain". We don't call l1 the echain. Suggest rename to `l1erc20Supply`

monitor/supply/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+		Namespace: "monitor",
+		Subsystem: "supply",
+		Name:      "l1_erc20_total_ether",
+		Help:      "ERC20 token supply on Ethereum chain in ether",
```
corverroos, line 20: `OMNI ERC20 ...`

corverroos (approved):

Issue #3342, opened 2025-03-10
Implement unique order IDs


Zodomo:
#### Problem to Solve
Sequential order IDs pose reorg risks as open transactions can be flipped on a reorg, rendering solver claims invalid. Switching to origin-unique order IDs prevents this from occurring.

#### Proposed Solution
Replace sequential order ID with a sequential order offset value for metrics and reporting, but replace order ID generation with a hash of user address, user nonce, and source chain ID.

PR #3343, opened 2025-03-10
feat(contracts/solve): switch to unique order IDs

Zodomo:
Replace sequential order IDs with sequential order offsets and a unique order ID (hash of user address, user nonce, and origin chain ID).

issue: #3342


contracts/solve/src/SolverNetInbox.sol
```diff
@@ -186,23 +196,48 @@ contract SolverNetInbox is OwnableRoles, ReentrancyGuard, Initializable, Deploye
...
+     * @param user Address of the user.
+     */
+    function getUserNonce(address user) external view returns (uint256) {
+        return _userNonce[user];
+    }
+
+    /**
+     * @notice Returns the next order offset.
+     */
+    function getOffset() external view returns (bytes32) {
```
kevinhalliday, line 239: `getNextOrderOffset()`? matches order id getter

tho as it stands this returns current this returns the last order offset (or zero). should update natspact

`getLastOrderOffset()`?

contracts/solve/src/SolverNetInbox.sol
```diff
@@ -109,6 +109,16 @@ contract SolverNetInbox is OwnableRoles, ReentrancyGuard, Initializable, Deploye
      */
     mapping(bytes32 id => OrderState) internal _orderState;
 
+    /**
+     * @notice Map order ID to order offset.
+     */
+    mapping(bytes32 id => bytes32) internal _orderOffset;
```
kevinhalliday, line 115: since this is now clearly an offset, I think we can just store as uint

contracts/solve/src/SolverNetInbox.sol
```diff
@@ -514,7 +550,8 @@ contract SolverNetInbox is OwnableRoles, ReentrancyGuard, Initializable, Deploye
      * @param orderData Order data to open.
      */
     function _openOrder(SolverNet.Order memory orderData) internal returns (ResolvedCrossChainOrder memory resolved) {
-        bytes32 id = _incrementId();
+        address user = orderData.header.owner;
```
kevinhalliday, line 553: why not just call it `owner`? and use `owner` in other getters / state funcs. rather than introducing a new character "user"

i don't feel strongly tho so don't view this as blocking

contracts/solve/src/SolverNetInbox.sol
```diff
@@ -560,7 +597,9 @@ contract SolverNetInbox is OwnableRoles, ReentrancyGuard, Initializable, Deploye
             timestamp: uint32(block.timestamp),
             updatedBy: updatedBy
         });
+
         _latestOrderIdByStatus[status] = id;
+        if (status == Status.Pending) _orderOffset[id] = _incrementOffset();
```
kevinhalliday, line 602: why the check on status here? Or I see this is in _upsertOrder

why not just put this in the only place it happens, `open`?

kevinhalliday: did this one

contracts/solve/src/SolverNetInbox.sol
kevinhalliday: did this one too

contracts/solve/src/SolverNetInbox.sol
kevinhalliday: did this one

Zodomo: `getLastOrderOffset()` works too!

contracts/solve/src/SolverNetInbox.sol
Zodomo: `owner` is a reserved name in `OwnableRoles` due to the `owner()` function. That's why I opted for `user`.

contracts/solve/src/SolverNetInbox.sol
Zodomo: Yeah it works well there too!

kevinhalliday (approved):
Will need to bump omega versions and redeploy

PR #3348, opened 2025-03-10
feat(solver): restrict call targets

kevinhalliday:
Restrict call targets

Allow
- symbiotic wsteth vaults on sepolia
- staking contract on omni evm
- flowgen bridging
- calls through middleman to allowed targets

Middleman is part of the "product" for devs, so we need to support it. 

issue: #3349

### DIFF
--- a/solver/app/reject.go
+++ b/solver/app/reject.go
@@ -315,7 +315,7 @@ func checkOrderCalls(order Order, isAllowed callAllowFunc) error {
 func checkCalls(destChainID uint64, calls []types.Call, isAllowed callAllowFunc) error {
 	for _, call := range calls {
 		if !isAllowed(destChainID, call.Target, call.Data) {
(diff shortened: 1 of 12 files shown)


solver/app/reject.go
```diff
@@ -305,6 +311,36 @@ func checkLiquidity(ctx context.Context, expenses []TokenAmt, backend *ethbacken
...
+// checkCalls checks if all calls to destChainID are allowed.
+func checkCalls(destChainID uint64, calls []types.Call, isAllowed callAllowFunc) error {
+	for _, call := range calls {
+		if !isAllowed(destChainID, call.Target, call.Data) {
+			return newRejection(rejectCallNotAllowed, errors.New("call not allowed", "target", call.Target.Hex(), "data", hexutil.Encode(call.Data)))
+		}
+	}
+
+	return nil
+}
```
kevinhalliday, line 342: Should we add an error here (or catch and change before sending api response) that says something like "if you are a developer looking to use omni in your app, reach out to the team at some@email"?

corverroos: nah

Issue #3353, opened 2025-03-10
Migrate stake to Staking delegations directly


Zodomo:
#### Problem to Solve
Instead of bridging OMNI to the user on Omni EVM, we need to stake them as well.

#### Proposed Solution
Replace the transfer call in the SolverNet order with a call to Staking's `delegateFor` function.

PR #3354, opened 2025-03-10
feat(contracts/core): migrate to staking delegation on staging

Zodomo:
Replaced SolverNet order for a token transfer with staking via the `delegateFor` function in our `Staking` contract.

issue: #3353


contracts/core/src/token/distributor/MainnetMerkleDistributorWithDeadline.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity 0.8.24;
+
+import { MerkleDistributorWithDeadline } from "./MerkleDistributorWithDeadline.sol";
+
+contract OmegaMerkleDistributorWithDeadline is MerkleDistributorWithDeadline {
```
kevinhalliday, line 6: This is `Omega` but also see another omega below

contracts/core/src/token/distributor/MainnetMerkleDistributorWithDeadline.sol
```diff
@@ -0,0 +1,29 @@
...
+        address token_,
+        bytes32 merkleRoot_,
+        uint256 endTime_,
+        address omniPortal_,
+        address genesisStaking_,
+        address solverNetInbox_
+    ) MerkleDistributorWithDeadline(token_, merkleRoot_, endTime_, omniPortal_, genesisStaking_, solverNetInbox_) { }
+
+    function _getValidator() internal override returns (address) {
+        uint256 selection = ++_delegationCount % 4;
```
kevinhalliday, line 22: could do something like `uint256(delegator) % 4` if we want?

contracts/core/src/token/distributor/MainnetMerkleDistributorWithDeadline.sol
Zodomo: Fixed

contracts/core/src/token/distributor/MainnetMerkleDistributorWithDeadline.sol
Zodomo: Would be much preferred to eliminate the SLOAD/SSTORE. Is random distribution okay?

Zodomo: Replaced with random distribution to save the SLOAD/SSTORE.

PR #3357, opened 2025-03-11
chore(solver): add holesky on omega

kevinhalliday:
Add back holesky on omega.

Looks like holesky might be back, so let's try it. 

issue: none


Zodomo (approved):

kevinhalliday:
should not need with https://github.com/omni-network/omni/pull/3363

PR #3359, opened 2025-03-11
ci(e2e/anvil): infer upstream image platform

corverroos:
When building `anvilproxy`, don't hardcode upstream forge docker image platform. Rather infer from environment. 

On latest docker and MacOS, this resulted in the forge binary being amd64, while the anvil proxy image was arm64, so it crashed.

issue: none


e2e/anvilproxy/Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG FOUNDRY_VERSION
-FROM --platform=linux/amd64 ghcr.io/foundry-rs/foundry:${FOUNDRY_VERSION}
```
corverroos, line 2: foundary now support multi-arch images, so auto-detect

chmllr (approved):

PR #3360, opened 2025-03-11
feat(monitor): symbiotic deposit flowgen

chmllr:
Symbiotic deposit flowgen.

issue: #3006


monitor/flowgen/flowgen.go
```diff
@@ -48,10 +49,31 @@ func Start(
...
 	result, err := bridgeJobs(network.ID)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "bridge jobs")
 	}
 	jobs = append(jobs, result...)
 
+	if network.ID == netconf.Omega {
+		deposit := big.NewInt(0).Mul(util.MilliEther, big.NewInt(20)) // 0.02 ETH
+		wstETHBaseSepolia := common.HexToAddress("0x6319df7c227e34B967C1903A08a698A3cC43492B")
```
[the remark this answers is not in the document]
chmllr: How about just using the list stored in solver? I added `FindBySymbol`, please TAL.

monitor/flowgen/util/util.go
```diff
@@ -26,3 +35,51 @@ func dec(amt float64, decimals int) *big.Int {
 
 	return new(big.Int).SetUint64(uint64(p))
 }
+
+// ApproveToken gives `user` max allowance for `token`.
+func ApproveToken(ctx context.Context, backend *ethbackend.Backend, token, user, contract common.Address) error {
```
corverroos, line 40: this is duplicated from `solver/app`, at least add a TODO to remove duplication.

Also, I think the comment can be improved: `ApproveToken gives the 'contract' max allowance to spend the 'user's 'tokens'.`

monitor/flowgen/symbiotic/config.go
```diff
@@ -0,0 +1,28 @@
...
+	expenseToken tokens.Token
+	vaultAddr    common.Address
+}
+
+var config = map[netconf.ID]flowConfig{
+	netconf.Omega: {
+		srcChain:     evmchain.IDBaseSepolia,
+		dstChain:     evmchain.IDHolesky,
+		depositToken: tokens.WSTETH,
+		expenseToken: tokens.WSTETH,
```
corverroos, line 25: is this really required? Isn't symbiotic always `WSETH` on all chains and all networks?

monitor/flowgen/flowgen.go
```diff
@@ -45,14 +45,40 @@ func Start(
...
-	result, err := bridgeJobs(network.ID)
+	result, err := bridgeJobs(ctx, network.ID, owner)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "bridge jobs")
 	}
 	jobs = append(jobs, result...)
 
+	if network.ID == netconf.Omega {
+		deposit := big.NewInt(0).Mul(util.MilliEther, big.NewInt(20)) // 0.02 ETH
```
corverroos, line 68: I'd follow your existing `bridgeJobs` patter here, so wrap this in a `symbioticJobs` function

monitor/flowgen/flowgen.go
```diff
@@ -64,7 +90,8 @@ func Start(
 					return
 				case <-timer.C:
 					jobsTotal.Inc()
-					if err := run(log.WithCtx(ctx, "job", job.Name), backends, job); err != nil {
+					ctx := log.WithCtx(ctx, "job", job.Name)
```
corverroos, line 93: assigning ctx inside the loop is risky, are you sure you are not overriding the ctx for other loop iterations?

monitor/flowgen/symbiotic/symbiotic.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return types.Job{}, errors.Wrap(err, "get backend")
+	}
+
+	srcChainTkn, ok := solver.AllTokens().FindBySymbol(conf.srcChain, conf.depositToken.Symbol)
+	if !ok {
+		return types.Job{}, errors.Wrap(err, "src token not found")
+	}
+
+	dstChainTkn, ok := solver.AllTokens().FindBySymbol(conf.dstChain, conf.expenseToken.Symbol)
```
corverroos, line 55: please add a TODO to solver/app.Tokens to move it to a separate package. It isn't good for other services (like monitor) to import the `app` package from another service. That is a very large dependency.

monitor/flowgen/symbiotic/symbiotic.go
```diff
@@ -0,0 +1,148 @@
...
+// orderData returns the order data required to do the job.
+func orderData(
+	owner common.Address,
+	srcChain, dstChain uint64,
+	srcChainToken common.Address,
+	dstChainToken common.Address,
+	symbioticContractAddress common.Address,
+	amount *big.Int,
+) (bindings.SolverNetOrderData, error) {
+	srcToken, ok := solver.AllTokens().Find(srcChain, srcChainToken)
```
corverroos, line 96: you already found the token above, why not just pass that in?

solver/app/tokens.go
```diff
@@ -91,9 +91,10 @@ var tokens = append(Tokens{
 	omniERC20(netconf.Staging),
 	omniERC20(netconf.Devnet),
 
-	// wtSETH
+	// wstETH
 	wstETH(evmchain.IDHolesky, common.HexToAddress("0x8d09a4502cc8cf1547ad300e066060d043f6982d")),
 	wstETH(evmchain.IDSepolia, common.HexToAddress("0xB82381A3fBD3FaFA77B3a7bE693342618240067b")),
+	wstETH(evmchain.IDBaseSepolia, common.HexToAddress("0x6319df7c227e34B967C1903A08a698A3cC43492B")),
```
corverroos, line 97: you are duplicating this, see golden file, it always contains the address. Please remove as it is already present.

monitor/flowgen/symbiotic/config.go
```diff
@@ -0,0 +1,28 @@
...
+type flowConfig struct {
+	srcChain     uint64
+	dstChain     uint64
+	depositToken tokens.Token
+	expenseToken tokens.Token
+	vaultAddr    common.Address
+}
+
+var config = map[netconf.ID]flowConfig{
+	netconf.Omega: {
```
corverroos, line 21: we should test everything in devnet/CI first.

monitor/flowgen/types/types.go
```diff
@@ -16,12 +16,14 @@ type Job struct {
...
 
-	Network netconf.ID
+	NetworkID netconf.ID
 
 	SrcChain uint64
 	DstChain uint64
 
 	Owner common.Address
 
+	InboxAddr common.Address
```
corverroos, line 26: this job struct is getting very cluttered with all kinds of data, many of which are constant for the network, so never change. The Job struct should only contain data unique to the job. But I can also take a go at cleaning this up later, no worries.

corverroos (approved):

monitor/flowgen/symbiotic/config.go
chmllr: Good question. Theoretically, Symbiotic supports a lot of tokens (see [existing vaults](https://app.symbiotic.fi/deposit?sort=tvl_desc&filter_option=collateral)), but if we can assume that flowgen will never work with other tokens, we can hardcode wstETH. WDYT?

monitor/flowgen/flowgen.go
chmllr: Will do! Context: I absolutely planned this to do this, but without introducing abstractions for just one job in the same PR.

PR #3367, opened 2025-03-11
fix(monitor): fix avg staking rewards query

chmllr:
Fixes bugs:

- exclude jailed validators
- reset timer on errors
- don't exit on success

issue: none


lib/cchain/queryutil/staking.go
```diff
@@ -165,7 +165,7 @@ func waitUntil(ctx context.Context, cprov cchain.Provider, target uint64) error
 			return nil
 		}
 
-		time.Sleep(2 * time.Second)
+		time.Sleep(10 * time.Second)
```
chmllr, line 168: For our purpose, we don't need to check every block.

Issue #3199, opened 2025-02-26
Improve solver metrics and dashboards


corverroos:
#### Problem to Solve
Identify key metrics for solver golang service

#### Proposed Solution
Implement them and build dashboards that allow tracking/monitoring/debugging all important solver metrics via grafana

PR #3378, opened 2025-03-12
refactor(solver/app): explicit pending data

corverroos:
Split `solver/app.Order`into multiple types, since not all fields are always available. `PendingData` is only available in `pending` state, similarly for `FilledData`.

Had to refactor monitoring since the data isn't always available.

issue: #3199


kevinhalliday (approved):
Nice. Good catch

PR #3387, opened 2025-03-12
feat(sdk): allow get status by just id

kevinhalliday:
Allow useGetOrderStatus with just order id.

issue: none


PaulLeCam (approved):

PaulLeCam:
We can remove the `sdk/src/index.test.ts` suite if it's too annoying, I just added it to have a simple test running to check the setup initially.

PR #3396, opened 2025-03-13
chore(contracts/core): change migration to upgrade

Zodomo:
Changed migration typehash to say `Upgrade` rather than `Migration`.

issue: none


ttarsi (approved):

contracts/core/src/token/MerkleDistributorWithDeadline.sol
```diff
@@ -30,7 +30,7 @@ contract MerkleDistributorWithDeadline is MerkleDistributor, Ownable, EIP712 {
         "OrderData(address owner,uint64 destChainId,Deposit deposit,Call[] calls,TokenExpense[] expenses)Deposit(address token,uint96 amount)Call(address target,bytes4 selector,uint256 value,bytes params)TokenExpense(address spender,address token,uint96 amount)"
     );
     bytes32 internal constant MIGRATION_TYPEHASH =
```
kevinhalliday, line 32: change constant name?

kevinhalliday (approved):

contracts/core/src/token/MerkleDistributorWithDeadline.sol
Zodomo: Should I also update `getMigrationDigest` to `getUpgradeDigest`?

Zodomo: Also, should I change `migrateToOmni` and `migrateUserToOmni` to `upgradeStake` and `upgradeUserStake`?

kevinhalliday: yeah can rename functions. don't need to but doesn't hurt

PR #3409, opened 2025-03-14
refactor(lib/umath): add ethereum unit conversion

corverroos:
Add convenience functions to `umath` package for converting between wei/gwei/ether.

issue: none


chmllr (approved):
Thanks!

PR #3443, opened 2025-03-18
chore(sdk): enable mainnet

ga-reth:
- enable mainnet

issue: none


sdk/src/context/omni.tsx
```diff
@@ -10,6 +10,8 @@ function apiUrl(env: Environment): string {
       return 'http://localhost:26661/api/v1'
     case 'testnet':
       return 'https://solver.omega.omni.network/api/v1'
+    case 'mainnet':
+      return 'https://solver.omni.network/api/v1'
```
kevinhalliday, line 14: Needs to be `https://solver.mainnet.omni.network/api/v1`

kevinhalliday (approved):

PR #3446, opened 2025-03-18
chore(e2e): update mainnet solver pin

Zodomo:
Updating mainnet to the solver version that has my address whitelisted so I can generate mainnet solver order activity.

issue: none


Zodomo:
Closed as Kevin opened the same PR literally minutes before mine 😂

PR #3455, opened 2025-03-19
feat(monitor/flowgen): dynamic order size for native ETH

chmllr:
Introduces dynamic order sizes that will maximize the flow volume automatically.

issue: #3006

### DIFF
--- a/monitor/flowgen/bridging/bridging.go
+++ b/monitor/flowgen/bridging/bridging.go
@@ -67,11 +67,11 @@ func openOrder(
 		return solvernet.OrderID{}, false, errors.New("dst token not found")
 	}
 
-	orderSize, err := estimateOrderSize(ctx, networkID, backend.Client, owner, conf)
+	orderSize, ok, err := estimateOrderSize(ctx, networkID, backend.Client, owner, conf, srcToken.Token)
 	if err != nil {
 		return solvernet.OrderID{}, false, errors.Wrap(err, "estimate order size")
 	}
-	if orderSize.Cmp(bi.Zero()) == 0 {
+	if !ok {
 		return solvernet.OrderID{}, false, nil
 	}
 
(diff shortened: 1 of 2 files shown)


monitor/flowgen/bridging/bridging.go
```diff
@@ -61,7 +67,15 @@ func openOrder(
 		return solvernet.OrderID{}, false, errors.New("dst token not found")
 	}
 
-	expense := app.TokenAmt{Token: dstToken, Amount: conf.orderSize}
+	orderSize, err := estimateOrderSize(ctx, networkID, backend.Client, owner, conf)
+	if err != nil {
+		return solvernet.OrderID{}, false, errors.Wrap(err, "estimate order size")
+	}
+	if orderSize.Cmp(bi.Zero()) == 0 {
```
corverroos, line 74: no need to use `Cmp` anymore. See `bi` package, you can use `bi.IsZero` in this case.

monitor/flowgen/bridging/bridging.go
```diff
@@ -111,3 +125,39 @@ func Jobs(networkID netconf.ID, backends ethbackend.Backends, owner common.Addre
...
+	client ethclient.Client,
+	owner common.Address,
+	conf flowConfig,
+) (*big.Int, error) {
+	balance, err := client.BalanceAt(ctx, owner, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	thresholds, ok := eoa.GetFundThresholds(tokens.ETH, networkID, eoa.RoleFlowgen)
```
corverroos, line 143: you are hardcoding `ETH` token, but the `openOrder` logic supports dynamic native tokens, you call `app.AllTokens().Find(`. I'd suggest being more consistent, and either use hardcoded values or dynamic native chain tokens, not both. Suggest passing in `srcToken.Token` rather.

monitor/flowgen/bridging/bridging.go
```diff
@@ -111,3 +125,39 @@ func Jobs(networkID netconf.ID, backends ethbackend.Backends, owner common.Addre
...
+	if !ok {
+		// Skip accounts without thresholds
+		return bi.Zero(), nil
```
corverroos, line 146: this is unexpected for flowgen, rather error

monitor/flowgen/bridging/bridging.go
```diff
@@ -111,3 +125,39 @@ func Jobs(networkID netconf.ID, backends ethbackend.Backends, owner common.Addre
...
+
+	// if order size is too small, do nothing
+	if orderSize.Cmp(conf.minOrderSize) < 0 {
+		return bi.Zero(), nil
+	}
+
+	// cap the order if necessary
+	if orderSize.Cmp(conf.maxOrderSize) > 0 {
+		orderSize = conf.maxOrderSize
+	}
```
corverroos, line 160: please use the new `bi` package for all `bigInt` related logic.

monitor/flowgen/bridging/bridging.go
```diff
@@ -111,3 +125,39 @@ func Jobs(networkID netconf.ID, backends ethbackend.Backends, owner common.Addre
 
 	return []types.Job{job1, job2}, nil
 }
+
+// estimateOrderSize checks the current balance of the flowgen EOA and returns
+// the maximal possible order size or nil if the minimal balance is reached.
+func estimateOrderSize(
```
corverroos, line 131: nit: rather return `*big.Int, bool, error`, the bool makes result valid status explicit, rather than implicitly encoding it in the value. This is slightly easier to read.

corverroos (approved):

PR #3456, opened 2025-03-19
chore(solver): stream conf min2 for l1 chains

kevinhalliday:
Stream conf min2 for l1 chains in solver event proc.

issue: none


solver/app/app.go
```diff
@@ -38,7 +38,14 @@ func chainVersFromID(network netconf.ID, chainID uint64) []xchain.ChainVersion {
 	if network == netconf.Devnet {
 		return []xchain.ChainVersion{
 			xchain.NewChainVersion(chainID, xchain.ConfLatest),
-			xchain.NewChainVersion(chainID, xchain.ConfMin1),
+			xchain.NewChainVersion(chainID, xchain.ConfMin2),
```
kevinhalliday, line 41: Switched this do `ConfMin2` so we run this conf level in e2e

Zodomo (approved):

PR #3462, opened 2025-03-20
fix(monitor/flowgen): fix reverse route when bridging

chmllr:
Fixes the reverse routing and adds debug logs.

issue: #3006

### DIFF
--- a/monitor/flowgen/bridging/bridging.go
+++ b/monitor/flowgen/bridging/bridging.go
@@ -106,7 +106,7 @@ func openOrder(
 	return orderID, true, nil
 }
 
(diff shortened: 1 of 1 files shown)


monitor/flowgen/bridging/bridging.go
```diff
@@ -118,7 +119,9 @@ func Jobs(networkID netconf.ID, backends ethbackend.Backends, owner common.Addre
 		return nil, err
 	}
 
-	job2, err := newJob(networkID, backends, conf, owner)
+	conf2 := conf
```
corverroos, line 122: nit, add comment `# Clone the job and flip the chains`

PR #3468, opened 2025-03-20
chore(halo): prevent overflowing

chmllr:
Prevent overflowing when computing withdrawal rewards (it would only overflow if rewards are far larger than total $OMNI supply, but it's still better to catch this in case of bugs).

issue: none


halo/withdraw/wrapper_internal_test.go
```diff
@@ -82,11 +82,16 @@ func TestToGwei(t *testing.T) {
...
 			t.Parallel()
-			gwei, wei := toGwei(tt.arg)
+			gwei, wei, err := toGwei(tt.arg)
+			require.NoError(t, err)
 			require.Equal(t, tt.gwei, gwei)
 			require.Equal(t, tt.wei, wei)
 		})
 	}
+
+	const maxUint64 = ^uint64(0)
```
corverroos, line 92: can use `math.MaxUint64`

corverroos (approved):

PR #3471, opened 2025-03-20
chore(solver/app): reset oldest order

corverroos:
Reset `oldestOrder` gauge to remove stale data for chains without any orders anymore.

issue: none


corverroos:
The goal is to remove these "flat" sections in the graph, rather resulting in "no data" for those sections.
image

chmllr (approved):

PR #3484, opened 2025-03-22
bug: 🐛 Fix bn.js and node error

KhairallahA:
While trying to run the SDK using Node v22.11.0 and pnpm v10.6.5, I encountered the following error:

```bash
error TS2688: Cannot find type definition file for 'bn.js'.
  The file is in the program because:
    Entry point for implicit type library 'bn.js'

error TS2688: Cannot find type definition file for 'node'.
  The file is in the program because:
    Entry point for implicit type library 'node'


Found 2 errors.

 ELIFECYCLE  Command failed with exit code 2.
 ELIFECYCLE  Command failed with exit code 2.
 ELIFECYCLE  Command failed with exit code 2.
```

After installing `@types/bn.js` and `@types/node`, the issue was resolved.

### DIFF
--- a/sdk/packages/react/package.json
+++ b/sdk/packages/react/package.json
@@ -43,6 +43,8 @@
   "devDependencies": {
     "@testing-library/dom": "^10.4.0",
     "@testing-library/react": "^16.2.0",
+    "@types/bn.js": "^5.1.6",
+    "@types/node": "^22.13.10",
     "@types/react": "^19.0.8",
     "@types/react-dom": "^19.0.4",
     "@vitest/coverage-v8": "^3.0.7",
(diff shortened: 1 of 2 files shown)


KhairallahA:
Hi @PaulLeCam!

I did that like you see here:
image

PaulLeCam:
Thanks for the details.

I'd like to figure out what is the cause of the issue, considering I can't reproduce it locally and it doesn't appear in our CI environment either, I'm wondering if there is anything else with your setup that might affect it please? Have you made any change to the code or dependencies that could explain the added requirement on these types?

KhairallahA:
Hi my bro @PaulLeCam, got the problem!

I was experiencing this issue just by running `pnpm run build` without any modifications or additions to the code.

I encountered the same issue with the test-utils folder in the latest update. To solve it, I needed to install `@types/bin.js`.

I fixed it by replacing my global tsc version (`5.8.2`) with the same version as the one expected in the project (`5.7.2`), and now everything works — I think because the stricter TS rules from `5.8.2` are gone.

PaulLeCam:
Thanks for the details!

PR #3502, opened 2025-03-25
chore(lib/ethclient): wire headerdb

corverroos:
Replace headercache internals with headerdb to cache by height as well instead of only hash.

issue: none


lib/ethclient/headercache.go
```diff
@@ -3,97 +3,104 @@ package ethclient
...
 
-func (c *headerCache) HeaderByHash(ctx context.Context, hash common.Hash) (*types.Header, error) {
-	if header, ok := c.get(hash); ok {
+func (c headerCache) HeaderByHash(ctx context.Context, hash common.Hash) (*types.Header, error) {
+	if header, ok, err := c.db.ByHash(ctx, hash); err != nil {
+		return nil, err
+	} else if ok {
 		return header, nil
 	}
+	cacheMisses.WithLabelValues(c.Name()).Inc()
```
chmllr, line 79: Aren't we missing `cacheHits.WithLabelValues(c.Name()).Inc()` above?

chmllr (approved):

lib/ethclient/headercache.go
corverroos: yes good catch

PR #3535, opened 2025-03-27
fix(sdk): patch api error handling

kevinhalliday:
Patch json error handling.

issue: none

### DIFF
--- a/sdk/packages/react/src/internal/api.ts
+++ b/sdk/packages/react/src/internal/api.ts
@@ -22,8 +22,9 @@ export async function fetchJSON(
   const json = await res.json()
 
   if (!res.ok) {
-    if (!isJSONError(json.error)) new Error(`${res.status} ${res.statusText}`)
-    throw json.error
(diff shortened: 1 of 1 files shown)


PaulLeCam (approved):

sdk/packages/react/src/internal/api.ts
```diff
@@ -22,8 +22,9 @@ export async function fetchJSON(
   const json = await res.json()
 
   if (!res.ok) {
-    if (!isJSONError(json.error)) new Error(`${res.status} ${res.statusText}`)
-    throw json.error
+    if (isJSONError(json)) throw json
+    if (isJSONError(json.error)) throw json.error
```
corverroos, line 26: shouldn't we rather just fix this in the server, so return the previous `.error` field? then no need to patch the SDK.

corverroos:
Closed in favour of https://github.com/omni-network/omni/pull/3537

PR #3537, opened 2025-03-27
fix(solver/app): revert to nested API error json

corverroos:
Revert solver API errors to nested `{"error":{...}}` which was introduced in #3223.

Also add some golden files to prevent accidental issues like this.

issue: none


solver/app/testdata/TestQuote/no_deposit_of_expense_amount_specified/resp_body.json
```diff
@@ -0,0 +1,7 @@
+{
+  "error": {
```
corverroos, line 2: example error message

PR #3538, opened 2025-03-27
fix(halo/app): verify tx signing info empty

corverroos:
Ensure that tx signing info is empty. This prevents malicious proposers from adding arbitrary data to the chain, which negatively affects chain operation.

This issue was reported on Immunify

issue: none

### DIFF
--- a/solver/app/metrics.go
+++ b/solver/app/metrics.go
@@ -84,6 +84,7 @@ var (
 		Subsystem: "worker",
 		Name:      "job_duration_seconds",
 		Help:      "Job duration in seconds by chain and event status",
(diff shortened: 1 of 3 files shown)


solver/app/metrics.go
```diff
@@ -84,6 +84,7 @@ var (
 		Subsystem: "worker",
 		Name:      "job_duration_seconds",
 		Help:      "Job duration in seconds by chain and event status",
+		Buckets:   prometheus.ExponentialBucketsRange(0.1, 60, 10),
```
corverroos, line 87: not related, but required for metrics, slipped in.

chmllr (approved):

Issue #3534, opened 2025-03-27
Handle invalid checksum addr


ttarsi:
#### Problem to Solve

when users provide an invalid checksum addr for a token address, we get the following unhandled error.
![Image](https://github.com/user-attachments/assets/87721b14-4fb0-4539-b007-18cddae1eac9)

#### Proposed Solution

handle the error and return properly. perhaps convert the address to proper checksum to see if it is a valid address.

PR #3545, opened 2025-03-27
fix(sdk): return an error when encoding order

PaulLeCam:
Wraps the order encoding so we can check and carry the error as needed.

issue: https://github.com/omni-network/omni/issues/3534

### DIFF
--- a/sdk/packages/react/src/hooks/useValidateOrder.ts
+++ b/sdk/packages/react/src/hooks/useValidateOrder.ts
@@ -63,9 +63,31 @@ export function useValidateOrder<abis extends OptionalAbis>({
   enabled,
 }: UseValidateOrderParams<abis>): UseValidateOrderResult {
   const { apiBaseUrl } = useOmniContext()
+  const encoded = encodeOrder(order)
 
-  function _encodeCalls() {
-    return order.calls.map((call) => {
+  const query = useQuery<ValidationResponse, FetchJSONError>({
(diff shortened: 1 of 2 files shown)


sdk/packages/react/src/hooks/useValidateOrder.ts
```diff
@@ -146,9 +167,12 @@ const isValidateRes = (json: unknown): json is ValidationResponse => {
 }
 
 const useResult = (
+  encoded: EncodeOrderResult,
   query: UseQueryResult<ValidationResponse, FetchJSONError>,
 ): UseValidateOrderResult =>
   useMemo(() => {
+    if (!encoded.ok) return { status: 'error', error: encoded.error }
+
```
kevinhalliday, line 175: This pattern makes sense - quicker feedback of encoding errors, and don't need to deal with react-query retrying on errors. So let's keep it 

but wonder it it would work if we just let the async func throw

kevinhalliday (approved):

sdk/packages/react/src/hooks/useValidateOrder.ts
PaulLeCam: Yes good question, I went with trying to bypass react-query but there might be better ways to handle this

kevinhalliday: react query right now retries on error (4 times) so better to bypass

PR #3550, opened 2025-03-28
fix(monitor/flowgen): check order before creation

chmllr:
Before opening an order, use the solver's check API. If the solver returns the error "insufficient inventory" don't create an order. Otherwise, propagate the error.

issue: none

### DIFF
--- a/e2e/app/setup.go
+++ b/e2e/app/setup.go
@@ -626,16 +626,11 @@ func writeMonitorConfig(ctx context.Context, def Definition, logCfg log.Config, 
 		return errors.New("monitor must use archive node, no archive node found")
 	}
 
-	solverData, ok := def.Infra.GetInfrastructureData().Instances["solver"]
-	if !ok {
-		return errors.New("no solver instance found")
-	}
-
 	cfg := monapp.DefaultConfig()
 	cfg.PrivateKey = privKeyFile
 	cfg.Network = def.Testnet.Network
 	cfg.HaloCometURL = archiveNode.AddressRPC()
-	cfg.SolverIPAddress = solverData.IPAddress.String()
+	cfg.SolverAddress = def.Testnet.SolverInternalAddr
 	cfg.HaloGRPCURL = haloGRPCAddress(archiveNode)
 	cfg.RouteScanAPIKey = def.Cfg.RouteScanAPIKey
 	cfg.LoadGen.ValidatorKeysGlob = validatorKeyGlob
--- a/e2e/app/setup.go
+++ b/e2e/app/setup.go
@@ -626,10 +626,16 @@ func writeMonitorConfig(ctx context.Context, def Definition, logCfg log.Config, 
 		return errors.New("monitor must use archive node, no archive node found")
 	}
 
+	solverData, ok := def.Infra.GetInfrastructureData().Instances["solver"]
+	if !ok {
+		return errors.New("no solver instance found")
+	}
+
 	cfg := monapp.DefaultConfig()
 	cfg.PrivateKey = privKeyFile
 	cfg.Network = def.Testnet.Network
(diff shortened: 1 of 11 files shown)


monitor/flowgen/bridging/bridging.go
```diff
@@ -96,48 +117,61 @@ func openOrder(
...
-		},
+		Calls:    []bindings.SolverNetCall{call.ToBinding()},
 	}
 
 	orderID, err := solvernet.OpenOrder(ctx, networkID, conf.srcChain, backends, owner, orderData)
 	if err != nil {
 		return types.Result{}, false, errors.Wrap(err, "open order")
 	}
 
+	if err = checker(ctx, stypes.CheckRequest{
```
corverroos, line 128: avoid inlining such complex things, it is hard to read, break this into multiple steps

monitor/flowgen/bridging/bridging.go
```diff
@@ -37,6 +39,19 @@ func newJob(networkID netconf.ID, backends ethbackend.Backends, conf flowConfig,
 		return types.Job{}, errors.Wrap(err, "src chain backend")
 	}
 
+	addrs, err := contracts.GetAddresses(ctx, networkID)
+	if err != nil {
+		return types.Job{}, errors.New("contract addresses")
+	}
+
+	solverAddr := eoa.MustAddress(networkID, eoa.RoleSolver)
+	checker := solver.NewChecker(
```
corverroos, line 48: not a fan of simply extracting the logic from solver into monitor. Encapsulation of logic and ownership is important. Why not call the actual API? Then we test test infra at the same time.

e2e/app/setup.go
```diff
@@ -626,10 +626,16 @@ func writeMonitorConfig(ctx context.Context, def Definition, logCfg log.Config,
...
+	solverData, ok := def.Infra.GetInfrastructureData().Instances["solver"]
+	if !ok {
+		return errors.New("no solver instance found")
+	}
+
 	cfg := monapp.DefaultConfig()
 	cfg.PrivateKey = privKeyFile
 	cfg.Network = def.Testnet.Network
 	cfg.HaloCometURL = archiveNode.AddressRPC()
+	cfg.SolverIPAddress = solverData.IPAddress.String()
```
corverroos, line 638: rather make this: `cfg.SolverAddr = def.Testnet.SolverAddr` which includes the port.

monitor/flowgen/bridging/bridging.go
```diff
@@ -96,48 +119,68 @@ func openOrder(
...
 
 	orderID, err := solvernet.OpenOrder(ctx, networkID, conf.srcChain, backends, owner, orderData)
 	if err != nil {
 		return types.Result{}, false, errors.Wrap(err, "open order")
 	}
 
+	req := stypes.CheckRequest{
+		SourceChainID:      conf.srcChain,
+		DestinationChainID: conf.dstChain,
+		FillDeadline:       uint32(time.Now().Add(time.Hour).Unix()), //nolint:gosec // unix timestamp won't overflow
```
corverroos, line 133: rather just do proper `umath.ToUint32` in production code

monitor/flowgen/bridging/bridging.go
```diff
@@ -96,48 +119,68 @@ func openOrder(
...
+			Amount: call.Value,
+		}},
+		Calls: []stypes.Call{stypes.Call(call)},
+	}
+	resp, err := solverClient.Check(ctx, req)
+	if err != nil {
+		return types.Result{}, false, errors.Wrap(err, "check solving")
+	}
+
+	if resp.Rejected {
```
corverroos, line 145: I suggest sticking to PR description which states: "only skip on insufficient inventory", that implies erroring on others which are unexpected.

corverroos (approved):

PR #3558, opened 2025-03-31
chore(solver/client): add error test

corverroos:
Add a unit test for sovler client json errors. Also do some small cleanups.

issue: none

### DIFF
--- a/solver/client/client.go
+++ b/solver/client/client.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"io"
 	"net/http"
 	"net/url"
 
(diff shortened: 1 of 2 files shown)


solver/client/client.go
```diff
@@ -51,18 +50,12 @@ func (c Client) Check(ctx context.Context, req types.CheckRequest) (types.CheckR
...
 	if resp.StatusCode != http.StatusOK {
-		var jsonError *types.JSONErrorResponse
-
-		bodyBytes, err := io.ReadAll(resp.Body)
-		if err != nil {
-			return types.CheckResponse{}, errors.Wrap(err, "response body")
-		}
-
-		if err := json.Unmarshal(bodyBytes, &jsonError); err != nil {
-			return types.CheckResponse{}, errors.Wrap(err, "unmarshal")
```
corverroos, line 62: rather return original error and status code instead of json unmarshal error

solver/client/client.go
```diff
@@ -73,14 +66,10 @@ func (c Client) Check(ctx context.Context, req types.CheckRequest) (types.CheckR
...
 }
 
-func (c Client) uri(path string) (*url.URL, error) {
-	absURL, err := url.JoinPath(c.host, path)
-	if err != nil {
-		return nil, errors.Wrap(err, "join path", "base", c.host, "path", path)
-	}
-	uri, err := url.Parse(absURL)
+func (c Client) uri(path string) (string, error) {
+	uri, err := url.JoinPath(c.host, path)
```
corverroos, line 70: JoinPath already doesn't Parse inside it

PR #3569, opened 2025-03-31
chore(solver): open mainnet targets

kevinhalliday:
Open mainnet targets.

issue: none


Zodomo (approved):
Why not just comment out the test cases instead of removing them?

kevinhalliday:
> Why not just comment out the test cases instead of removing them?

no real reason tbh, just nice to remove when you can

corverroos (approved):

PR #3571, opened 2025-03-31
feat(contracts): emit OriginData event

Zodomo:
SolverNetInbox now emits an `OriginData` event that emits a typed `FillOriginData` struct so we can analyze it in Dune. Added validation for this to post-upgrade tests.

issue: none


contracts/solve/src/SolverNetInbox.sol
```diff
@@ -265,6 +265,9 @@ contract SolverNetInbox is OwnableRoles, ReentrancyGuard, Initializable, Deploye
         ResolvedCrossChainOrder memory resolved = _openOrder(orderData);
 
         emit Open(resolved.orderId, resolved);
+        emit OriginData(
+            resolved.orderId, abi.decode(resolved.fillInstructions[0].originData, (SolverNet.FillOriginData))
+        );
```
kevinhalliday, line 270: Maybe let's call it `FillOriginData`? matches our typename

Zodomo: I originally thought this would cause compiler issues so I avoided doing so, but it turns out I was wrong! Applying this change.

PR #3572, opened 2025-04-01
ci(e2e): add solver address to testnet

corverroos:
Adds `SolverAddr` to `Testnet` and use it in e2e/tests.

Populate this from `InfraData`, which is populated in docker and vmcompose respectively.

issue: none


chmllr (approved):
Thanks a lot!

PR #3584, opened 2025-04-01
chore(halo): fix pending blocks attest metric

kevinhalliday:
Do not set pending blocks gauge if attestation has already been approved.

issue: none


corverroos:
As discussed, the metric is correct

PR #3585, opened 2025-04-02
feat(halo/evmstaking): undelegate stake

chmllr:
Implements undelegation.

Issue: #3596


halo/withdraw/wrapper.go
```diff
@@ -45,12 +45,26 @@ func (w *BankWrapper) SendCoinsFromModuleToAccountNoWithdrawal(ctx context.Conte
...
 // creates EVM withdrawal to the user account.
-// TODO(corver): This is unexpected in magellan upgrade, must be implemented in drake.
-func (w *BankWrapper) UndelegateCoinsFromModuleToAccount(ctx context.Context, senderModule string, recipientAddr sdk.AccAddress, amt sdk.Coins) error {
-	log.Error(ctx, "Unexpected call to DelegateCoinsFromAccountToModule [BUG]", nil, "sender", senderModule, "recipient", recipientAddr, "amt", amt)
+func (w *BankWrapper) UndelegateCoinsFromModuleToAccount(ctx context.Context, senderModule string, recipientAddr sdk.AccAddress, coins sdk.Coins) error {
+	log.Debug(ctx, "Undelegating coins from module to account", "sender", senderModule, "recipient", recipientAddr, "coins", coins)
 
-	if err := w.Keeper.UndelegateCoinsFromModuleToAccount(ctx, senderModule, recipientAddr, amt); err != nil {
-		return errors.Wrap(err, "undelegate coins from module to account")
+	if err := w.Keeper.UndelegateCoinsFromModuleToAccount(ctx, senderModule, recipientAddr, coins); err != nil {
```
corverroos, line 51: I think we should align `UndelegateCoinsFromModuleToAccount` with `SendCoinsFromAccountToModule`, so follow the same pattern, not different patterns. 

I think we can just follow the `SendCoinsFromAccountToModule` logic for now. But also add a check for VestingAccount and error if that is the case with comment: `Current implementation doesn't support vesting accounts or locked coins". Then simply burning and creating the withdrawal in both cases is sufficient.

halo/evmstaking/keeper/keeper.go
```diff
@@ -145,12 +147,15 @@ func (Keeper) Name() string {
...
+	return []common.Address{k.address}, [][]common.Hash{{
+		createValidatorEvent.ID, delegateEvent.ID, undelegateEvent.ID, editValidatorEvent.ID,
+	}}
 }
 
 // Deliver processes a omni deposit log event, which must be one of:
-// - CreateValidator
-// - Delegate.
+// - CreateValidator,
+// - Delegate,
```
corverroos, line 157: also EditValidator

halo/evmstaking/keeper/keeper.go
```diff
@@ -335,6 +382,10 @@ func (k Keeper) createAccIfNone(ctx context.Context, addr sdk.AccAddress) {
 	}
 }
 
+func (k Keeper) hasAccount(ctx context.Context, addr sdk.AccAddress) bool {
+	return k.aKeeper.HasAccount(ctx, addr)
+}
+
```
corverroos, line 388: not a fan of functions that just call another function, it just obfuscates what is actually happening. Why not just call the actual function directly?

halo/evmstaking/keeper/keeper.go
```diff
@@ -283,6 +297,39 @@ func (k Keeper) deliverDelegate(ctx context.Context, ev *bindings.StakingDelegat
...
+	if _, err := k.sKeeper.GetValidator(ctx, valAddr); err != nil {
+		return errors.New("validator does not exist", "validator", valAddr.String())
+	}
+
+	amountCoin, _ := omniToBondCoin(ev.Amount)
+
+	if !k.hasAccount(ctx, delAddr) {
+		return errors.New("no account", "delegator", delAddr.String())
+	}
```
corverroos, line 317: nit: do we really need these two checks? I'm pretty sure `sServer.Undelegate` will fail if the validator or delegator doesn't exist. Isn't that sufficient?

halo/withdraw/wrapper.go
```diff
@@ -45,15 +48,24 @@ func (w *BankWrapper) SendCoinsFromModuleToAccountNoWithdrawal(ctx context.Conte
...
-func (w *BankWrapper) UndelegateCoinsFromModuleToAccount(ctx context.Context, senderModule string, recipientAddr sdk.AccAddress, amt sdk.Coins) error {
-	log.Error(ctx, "Unexpected call to DelegateCoinsFromAccountToModule [BUG]", nil, "sender", senderModule, "recipient", recipientAddr, "amt", amt)
+func (w *BankWrapper) UndelegateCoinsFromModuleToAccount(ctx context.Context, senderModule string, recipientAddr sdk.AccAddress, coins sdk.Coins) error {
+	log.Debug(ctx, "Undelegating coins from module to account", "sender", senderModule, "recipient", recipientAddr, "coins", coins)
 
-	if err := w.Keeper.UndelegateCoinsFromModuleToAccount(ctx, senderModule, recipientAddr, amt); err != nil {
-		return errors.Wrap(err, "undelegate coins from module to account")
+	acc := w.ak.GetModuleAccount(ctx, senderModule)
+	if acc == nil {
+		return errors.New("module account does not exist", "module_name", senderModule)
```
corverroos, line 56: can add `[BUG]` since this is unexpected

halo/withdraw/wrapper.go
```diff
@@ -45,15 +48,24 @@ func (w *BankWrapper) SendCoinsFromModuleToAccountNoWithdrawal(ctx context.Conte
...
 	}
 
-	return nil
+	_, ok := acc.(banktypes.VestingAccount)
+	if ok {
+		return errors.New("vesting accounts are not supported")
```
corverroos, line 61: same here, also a BUG for us.

corverroos: this check must (also) be done in `SendCoinsFromModuleToAccount`

corverroos: oh, and the account to check for "Vesting or not" isn't the module (since modules cannot be vesting accounts), but rather the `recipient` account

lib/ethclient/enginemock.go
```diff
@@ -129,7 +130,45 @@ func WithMockEditValidator(pubkey crypto.PubKey, params *bindings.StakingEditVal
...
+		contractAddr := common.HexToAddress(predeploys.Staking)
+		eventLog := types.Log{
+			Address: contractAddr,
+			Topics: []common.Hash{
+				undelegateEvent.ID,
+				common.HexToHash(delegatorAddr.Hex()), // delegator
+				common.HexToHash(valAddr.Hex()),       // validator
+			},
+			Data:  data,
+			Index: 300,
```
corverroos, line 162: best not use the same index

lib/netconf/static.go
```diff
@@ -219,6 +223,7 @@ var statics = map[ID]Static{
 		AVSContractAddress:   omegaAVS,
 		OmniExecutionChainID: evmchain.IDOmniOmega,
 		MaxValidators:        maxValidators,
+		UnbondingTime:        1814400 * time.Second,
```
corverroos, line 226: rather make it easy for humans to read: `time.Hour * 24 * 7 * 3 // 3 weeks`

lib/netconf/static.go
```diff
@@ -239,6 +244,7 @@ var statics = map[ID]Static{
 		AVSContractAddress:   mainnetAVS,
 		OmniExecutionChainID: evmchain.IDOmniMainnet,
 		MaxValidators:        maxValidators,
+		UnbondingTime:        1814400 * time.Second,
```
corverroos, line 247: also here time.Hour * 24 * 7 * 3, // 3 weeks

corverroos (approved):

PR #3586, opened 2025-04-02
feat(contracts/core): add undelegate to staking contract

chmllr:
Implements the undelegation function + event in the staking contract.

issue: none


contracts/core/src/octane/Staking.sol
```diff
@@ -299,6 +321,18 @@ contract Staking is OwnableUpgradeable, EIP712Upgradeable {
         emit Delegate(delegator, validator, msg.value);
     }
 
+    /**
+     * @notice Undelegate tokens from a validator
+     * @param delegator The address of the delegator
+     * @param validator The address of the validator to undelegate from
+     */
+    function _undelegate(address delegator, address validator) internal {
+        require(!isAllowlistEnabled || isAllowedValidator[validator], "Staking: not allowed val");
```
chmllr, line 330: Not sure if this check makes sense, maybe I should remove it?

corverroos: we need to charge a fee to avoid spam attacks

corverroos (changes requested):
Just blocking since it was approved but still isn't correct, so want to avoid accidental merge

contracts/core/src/octane/Staking.sol
```diff
@@ -230,6 +238,18 @@ contract Staking is OwnableUpgradeable, EIP712Upgradeable {
         _delegate(delegator, validator);
     }
 
+    /**
+     * @notice Undelegate tokens from a validator
+     * @dev Proxies x/staking.MsgUndelegate
+     * @param validator The address of the validator to undelegate from
+     * @param undelegationAmount The amount of ether tokens to undelegate
+     */
+    function undelegate(address validator, uint256 undelegationAmount) external payable {
```
corverroos, line 247: nit: rename to `amount` since `undelegationAmount` stutters a bit, it is the only amount and the context is clear. This also aligns with your emitted event nameing.

corverroos (approved):

PR #3590, opened 2025-04-03
chore(lib/ethclient): close idle connections periodically

corverroos:
Add `CloseIdleConnectionsForever` to ethclient. Call it in production use-cases.

This aims to mitigate the problem of TCP-keep-alive maintaining long-lived connections to stalled (but alive) RCP servers. This sometimes occurs with QuickNode, when one of the load-balanced upstream servers/nodes stalled, but we remain connected to it, requiring a restart of the affected service to reconnect to a healthy upstream. 

By periodically closing idle connections, we get best of both worlds, TCP-keep-alive improves performance of most queries since TCP connection doesn't need to be reestablished for each query. But we still reconnect every 5min ensuring good load balancing over time.

This also required some cleanup:
 - Pass in context in production uses of `ethclient.DialContext`.
 - Remove `ethclient` duplication in `monitor`.

issue: none

### DIFF
--- a/monitor/loadgen/loadgen.go
+++ b/monitor/loadgen/loadgen.go
@@ -13,7 +13,6 @@ import (
 	"github.com/omni-network/omni/lib/ethclient"
 	"github.com/omni-network/omni/lib/ethclient/ethbackend"
 	"github.com/omni-network/omni/lib/netconf"
-	"github.com/omni-network/omni/lib/xchain"
 
 	"github.com/ethereum/go-ethereum/common"
 	ethcrypto "github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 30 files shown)


monitor/xfeemngr/xfeemngr.go
```diff
@@ -74,11 +80,6 @@ func Start(ctx context.Context, network netconf.Network, cfg Config, privKeyPath
 		return err
 	}
 
-	ethClients, err := makeEthClients(toSync, cfg.RPCEndpoints)
```
corverroos, line 77: remove duplicate ethclients in monitor

monitor/loadgen/loadgen.go
```diff
@@ -124,7 +123,7 @@ func startXCaller(ctx context.Context, network netconf.Network, rpcEndpoints xch
 		return errors.Wrap(err, "load xcaller key", "path", keyPath)
 	}
 
-	backends, err := ethbackend.BackendsFromNetwork(network, rpcEndpoints, privKey)
```
corverroos, line 127: remove duplicate ethclients in monitor

chmllr (approved):

PR #3606, opened 2025-04-04
chore(hale): adds empty drake upgrade

chmllr:
Create empty drake upgrade.

issue: #3098


halo/app/upgrades/drake/upgrade.go
```diff
@@ -0,0 +1,66 @@
...
+		if err := staking.SetParams(ctx, StakingParams); err != nil {
+			return nil, errors.Wrap(err, "set staking params")
+		}
+
+		return mm.RunMigrations(ctx, configurator, fromVM)
+	}
+}
+
+func GenesisState(codec.JSONCodec) (map[string]json.RawMessage, error) {
+	return map[string]json.RawMessage{}, nil
```
corverroos, line 65: `nil` is valid zero map (for reading)

corverroos (approved):

halo/app/upgrades/drake/upgrade.go
corverroos: also, please see how magellan and uluwatu handle module parameter changes wrt genesis state

PR #3610, opened 2025-04-04
chore(sdk): prefer return error over throw

kevinhalliday:
Prefer returning error over throwing.

issue: none


kevinhalliday:
actually, going to handle differently

PR #3612, opened 2025-04-05
feat(contracts/solve): add onchain order fee sponsorship

Zodomo:
We had a [discussion](https://omni-chain.slack.com/archives/C06S1D1S37B/p1743803019108719) where apps sponsoring fees for their users came up. I believe that simply allowing apps to sponsor orders via an approval would be the best way to actually implement this, instead of relying on vault designs. Doing so also allows them to custody their funds however they'd like.

This functionality was introduced through adding a new order type where a `sponsor` address can be defined, and a `signature` from the sponsor to validate that they are indeed paying the fee for the order. Sponsors must pay the fee in the same token as the order's deposit. Allowing them to select which token they pay in introduces many problems to solve downstream, largely involving trust in the API (to quote conversion rates) or being limited to onchain oracles for specific tokens.

issue: none


Zodomo:
Closing as more discussion is needed regarding how we actually want to do this and this was primarily explorative.

PR #3615, opened 2025-04-07
fix(ethclient/headerdb): prevent deadlock

corverroos:
Fixes deadlock in `headercachedb` since `cosmos-db.MemDB` deosn't support iterating and writing. It can deadlock when iteration is large.

issue: none


monitor/xmonitor/monitor.go
```diff
@@ -21,18 +20,6 @@ func Start(
 	cprovider cchain.Provider,
 	rpcClients map[uint64]ethclient.Client,
 ) error {
-	{
-		meta, _ := evmchain.MetadataByID(evmchain.IDHolesky)
```
corverroos, line 25: not used anymore

chmllr (approved):

PR #3633, opened 2025-04-08
test(sdk): initial useOrder tests and useQuote fixes

ga-reth:
- initial useOrder tests - more coming
- useQuote - ensure we use a throwing condition within waitFor call
- will add back more useQuote checks in a follow up (need to resolve issue with query data not being mocked)

issue: #3170

### DIFF
--- a/sdk/packages/react/src/hooks/useOrder.test.ts
+++ b/sdk/packages/react/src/hooks/useOrder.test.ts
@@ -1,0 +1,130 @@
+import { waitFor } from '@testing-library/react'
+import { expect, test, vi } from 'vitest'
+import {
+  createMockWaitForTransactionReceiptResult,
+  createMockWriteContractResult,
+} from '../../test/mocks.js'
+import { renderHook } from '../../test/react.js'
+import { order } from '../../test/shared.js'
+import { useOrder } from './useOrder.js'
+
+const {
+  mockUseValidateOrder,
+  mockUseGetOrderStatus,
+  mockUseWriteContract,
+  mockUseWaitForTransactionReceipt,
+} = vi.hoisted(() => {
+  return {
+    mockUseValidateOrder: vi.fn(),
+    mockUseGetOrderStatus: vi.fn(),
+    mockUseWaitForTransactionReceipt: vi.fn().mockImplementation(() => {
+      return createMockWaitForTransactionReceiptResult({
+        isPending: true,
+        isSuccess: false,
+        data: undefined,
+        status: 'pending',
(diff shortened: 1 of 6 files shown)


sdk/packages/react/src/hooks/useOrder.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    waitFor(() => expect(result.current.isValidated).toBeFalsy()),
+    waitFor(() => expect(result.current.isError).toBeFalsy()),
+    waitFor(() => expect(result.current.isTxPending).toBeTruthy()),
+    waitFor(() => expect(result.current.isTxSubmitted).toBeFalsy()),
+    waitFor(() => expect(result.current.txMutation.data).toBeUndefined()),
+    waitFor(() => expect(result.current.isOpen).toBeFalsy()),
+    waitFor(() => expect(result.current.orderId).toBeUndefined()),
+    waitFor(() => expect(result.current.txHash).toBeUndefined()),
+    waitFor(() => expect(result.current.error).toBeUndefined()),
+  ])
```
PaulLeCam, line 88: Do we need to use a `Promise.all()` with multiple `waitFor()` here please? Can't all the assertions be part of a single `waitFor()` call?

sdk/packages/react/src/hooks/useOrder.test.ts
```diff
@@ -0,0 +1,130 @@
...
+  const res = await result.current.open()
+
+  await Promise.all([
+    waitFor(() => expect(result.current.isOpen).toBeTruthy()),
+    waitFor(() => expect(result.current.isTxPending).toBeFalsy()),
+    waitFor(() => expect(result.current.isTxSubmitted).toBeTruthy()),
+    waitFor(() => expect(result.current.txMutation.data).toBe('0xTxHash')),
+    waitFor(() => expect(result.current.txMutation.isSuccess).toBeTruthy()),
+    waitFor(() => expect(res).toBe('0xTxHash')),
+  ])
```
PaulLeCam, line 121: Same question as above

PaulLeCam (approved):

sdk/packages/react/src/hooks/useOrder.test.ts
ga-reth: in this instance they could be yeah, but if the values may update at different times it's needed

i'll update before merge

ga-reth: that also works! will update before merge

PR #3646, opened 2025-04-08
feat(contracts/solve): separated changes and improved permit2

Zodomo:
Separated the gasless order changes into a series of V2 contracts so the old contracts could remain intact in the event of new deployments being necessary. I also completely overhauled the permit2 integration, which now utilizes permit2 witness data so we can have a user sign once for the entire order (including transfers), rather than signing for transfers separately.

issue: #3057


contracts/solve/src/erc7683/IOriginSettler.sol
```diff
@@ -27,10 +26,7 @@ interface IOriginSettler is IERC7683 {
     /// @param order The GaslessCrossChainOrder definition
     /// @param originFillerData Any filler-defined data required by the settler
     /// @return ResolvedCrossChainOrder hydrated order data including the inputs and outputs of the order
-    function resolveFor(GaslessCrossChainOrder calldata order, bytes calldata originFillerData)
-        external
-        view
-        returns (ResolvedCrossChainOrder memory);
+    // function resolveFor(GaslessCrossChainOrder calldata order, bytes calldata originFillerData) external view returns (ResolvedCrossChainOrder memory);
```
kevinhalliday, line 29: Can delete

contracts/solve/src/SolverNetInboxV2.sol
```diff
@@ -0,0 +1,825 @@
...
+     * @param gasless Whether the order is gasless.
+     */
+    function _processDeposit(SolverNet.Deposit memory deposit, address from, bool gasless) internal {
+        if (deposit.token == address(0)) {
+            if (msg.value != deposit.amount) revert InvalidNativeDeposit();
+        } else {
+            if (msg.value > 0) revert InvalidNativeDeposit();
+            if (!gasless) deposit.token.safeTransferFrom2(from, address(this), deposit.amount);
+        }
+    }
```
kevinhalliday, line 649: So now we have two separate places where we process deposits? This function, for native and on-chain erc20 deposits. Then somewhere else (i think the _permit2 util?) for gasless erc20 deposits.

Can we clear that up. Maybe just rename this function. And instead of passing a boolean flag `gasless` to change behavior. Let's just not call this func if orders gasless.

contracts/solve/src/lib/Permit2Lib.sol
```diff
@@ -0,0 +1,214 @@
...
+    bytes internal constant GASLESS_ORDER_TYPE = abi.encodePacked(
+        "GaslessCrossChainOrder(address originSettler,address user,uint256 nonce,uint256 originChainId,uint32 openDeadline,uint32 fillDeadline,bytes32 orderDataType,OmniOrderData orderData)"
+    );
+
+    /**
+     * @notice Full EIP-712 type for the GaslessCrossChainOrder struct.
+     * @dev "Omni" prefix is added so wallets can show that the signature is for an Omni order.
+     */
+    bytes internal constant EIP712_GASLESS_ORDER_TYPE =
+        abi.encodePacked(GASLESS_ORDER_TYPE, CALL_TYPE, DEPOSIT_TYPE, "Omni", ORDERDATA_TYPE, TOKENEXPENSE_TYPE);
```
kevinhalliday, line 84: Should we just prefix the struct name to `OmniOrderData` like we discussed?

contracts/solve/src/lib/Permit2Lib.sol
```diff
@@ -0,0 +1,214 @@
...
+        return keccak256(abi.encode(DEPOSIT_TYPEHASH, deposit.token, deposit.amount));
+    }
+
+    /**
+     * @dev Hashes a call.
+     * @param call Call to hash.
+     * @return _ Hashed call.
+     */
+    function _hashCall(SolverNet.Call memory call) internal pure returns (bytes32) {
+        return keccak256(abi.encode(CALL_TYPEHASH, call.target, call.selector, call.value, keccak256(call.params)));
```
kevinhalliday, line 133: why are we keccaking call params?

contracts/solve/src/erc7683/IOriginSettlerV2.sol
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity =0.8.24;
+
+import { IERC7683 } from "./IERC7683.sol";
+
+/// @title IOriginSettler
+/// @notice Standard interface for settlement contracts on the origin chain
+/// @dev See https://github.com/ethereum/ERCs/blob/master/ERCS/erc-7683.md
+interface IOriginSettlerV2 is IERC7683 {
```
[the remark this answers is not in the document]
Zodomo: I guess a V2 interface isn't totally necessary, as the compiler wont complain about commented out functions that are implemented. I can remove the additional interface.

contracts/solve/src/SolverNetInboxV2.sol
```diff
@@ -0,0 +1,825 @@
...
+     * @notice Open an order to execute a call on another chain, backed by deposits.
+     * @dev Token deposits are transferred from msg.sender to this inbox.
+     * @param order OnchainCrossChainOrder to open.
+     */
+    function open(OnchainCrossChainOrder calldata order) external payable whenNotPaused(OPEN) nonReentrant {
+        SolverNet.Order memory _order = _validate(order);
+        address user = _order.header.owner;
+        bytes32 id = _getOrderId(false, user, _onchainUserNonce[user]++);
+
+        _open(_order, id, msg.sender, 0, false);
```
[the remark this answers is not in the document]
Zodomo: Added keyword args to calls with unclear values. However, for `open`, `user` in the `_open` call is msg.sender, as we take payment from the user opening the order, even if the order is assigned to someone else.

contracts/solve/src/lib/Permit2Lib.sol
```diff
@@ -0,0 +1,214 @@
...
+pragma solidity ^0.8.24;
+
+import { IERC7683 } from "../erc7683/IERC7683.sol";
+import { SolverNet } from "./SolverNet.sol";
+
+library Permit2Lib {
+    /**
+     * @notice Type for the Deposit struct.
+     */
+    bytes internal constant DEPOSIT_TYPE = abi.encodePacked("Deposit(address token,uint96 amount)");
```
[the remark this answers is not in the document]
Zodomo: You always use `abi.encodePacked` when converting strings to bytes. Use of `abi.encode` will lead to padding being introduced, which will break the EIP-712 types.
image

contracts/solve/src/SolverNetInboxV2.sol
```diff
@@ -0,0 +1,825 @@
...
+    /**
+     * @dev Validate OnchainCrossChainOrder parameters
+     * @param order OnchainCrossChainOrder to validate
+     */
+    function _validateOnchainOrder(OnchainCrossChainOrder calldata order) internal view {
+        if (order.fillDeadline <= block.timestamp) revert InvalidFillDeadline();
+        if (
+            order.orderDataType != Permit2Lib.OLD_ORDERDATA_TYPEHASH
+                && order.orderDataType != Permit2Lib.FULL_ORDERDATA_TYPEHASH
+        ) {
```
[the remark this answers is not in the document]
Zodomo: Changed to `HashLib` as that makes more sense.

contracts/solve/src/SolverNetInboxV2.sol
Zodomo: Refactored all of the deposit logic and cleaned things up a bit. Deposits take place in the `open` and `openFor` methods rather than `_open`, as different methods are used. I also renamed `_processDeposit` to `_onchainDeposit` and `_permit2` to `_gaslessDeposit`.

contracts/solve/src/lib/Permit2Lib.sol
Zodomo: Are we sure this change won't have major implications downstream in the API and/or solver?

contracts/solve/src/lib/Permit2Lib.sol
Zodomo: EIP-712 states that dynamic values (such as `bytes` and `string`) must be hashed.

contracts/solve/src/lib/Permit2Lib.sol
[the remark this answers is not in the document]
kevinhalliday: yeah I guess question was more about why using bytes instead of strings

contracts/solve/src/lib/Permit2Lib.sol
kevinhalliday: We are using new liib and new v2 contracts, so it's fine

contracts/solve/src/lib/Permit2Lib.sol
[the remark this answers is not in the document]
kevinhalliday: if that's how it works than all good

contracts/solve/src/lib/Permit2Lib.sol
kevinhalliday: tho guess we are still using same lib. we don't need to change struct name, can just use name in 

tho not a blocker. we can see what it looks like when doing this from a wallet

kevinhalliday (approved):

contracts/solve/src/lib/Permit2Lib.sol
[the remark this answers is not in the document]
Zodomo: `keccak256` expects bytes, it cannot hash a string directly. I stored them in the types they need to be used as.

contracts/solve/src/lib/Permit2Lib.sol
Zodomo: I will need to either duplicate the struct with the proper name, or create `SolverNetV2.sol`. As this is the only change, I'll just duplicate the struct for now.

PR #3649, opened 2025-04-09
refactor(lib/tokens): merge tokenmeta with tokens as Assets

corverroos:
Merges `tokenmeta` package with `tokens` and rename `Meta` to `Asset`. 

This simplifies things a lot. But had to move defining of tokens to test package and embed into `tokens` package via generated JSON file to remove all dependencies of `tokens` package to ensure it is self-contained and can therefore be imported by all other packages.

issue: none


Jojad9 (commented):
merge tokenname as Assets

PR #3666, opened 2025-04-10
chore(sdk): setup tests for API calls

PaulLeCam:
Adds unit tests to core package

issue: none


sdk/packages/core/src/api/getQuote.ts
```diff
@@ -0,0 +1,79 @@
...
+    expense: { ...expense, amount: fromHex(expense.amount, 'bigint') },
+  } as Quote
+}
+
+export async function getQuote(
+  apiBaseUrl: string,
+  params: GetQuoteParams,
+): Promise<Quote> {
+  const encoded = encodeQuote(params)
+  return await getQuoteWithEncoded(apiBaseUrl, encoded)
```
kevinhalliday, line 57: why do we need this `getQuoteWithEncoded` function? why not just encode and fetch

sdk/packages/core/src/api/getQuote.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    }),
+  ).resolves.toBeDefined()
+
+  await expect(
+    getQuote('http://localhost', {
+      ...params,
+      expense: { isNative: true },
+    }),
+  ).resolves.toBeDefined()
+})
```
kevinhalliday, line 48: let's test return values, no?

sdk/packages/core/src/contracts/didFillOutbox.ts
```diff
@@ -0,0 +1,25 @@
+import type { Address, Chain, Client, Transport } from 'viem'
+import { readContract } from 'viem/actions'
+import { outboxABI } from '../constants/abis.js'
+import type { ParseOpenEventReturn } from './parseOpenEvent.js'
+
+export type DidFillParams<chain extends Chain> = {
+  client: Client<Transport, chain>
+  outboxAddress: Address
+  resolvedOrder: ParseOpenEventReturn
```
kevinhalliday, line 9: Feels like this should be `ResolvedOrder` type

sdk/packages/core/src/contracts/parseOpenEvent.ts
```diff
@@ -0,0 +1,42 @@
...
+} from 'viem'
+import { inboxABI } from '../constants/abis.js'
+import { ParseOpenEventError } from '../errors/base.js'
+
+export type ParseOpenEventReturn = DecodeEventLogReturnType<
+  typeof inboxABI,
+  'Open'
+>['args']['resolvedOrder']
+
+export function parseOpenEvent(logs: Log[]): ParseOpenEventReturn {
```
kevinhalliday, line 15: Can we just have the return a `ResolvedOder`

sdk/packages/react-next/src/hooks/useDidFillOutbox.ts
```diff
@@ -0,0 +1,34 @@
...
+  destChainId,
+}: UseDidFillOutboxParams): UseDidFillOutboxReturn {
+  const client = useClient({ chainId: destChainId })
+  const { data: contracts } = useOmniContracts()
+  return useQuery({
+    queryKey: ['didFill', destChainId, resolvedOrder?.orderId],
+    queryFn: async () => {
+      if (!client || !contracts || !resolvedOrder) {
+        throw new Error('Invalid query parameters')
+      }
```
kevinhalliday, line 24: Can we have more granular / useful error mesages

sdk/packages/test-utils/src/constants.ts
```diff
@@ -31,8 +31,6 @@ export const TOKEN_ADDRESS =
   '0x73cc960fb6705e9a6a3d9eaf4de94a828cfa6d2a' as const
 export const INVALID_TOKEN_ADDRESS =
   '0x1234000000000000000000000000000000000000' as const
-export const ZERO_ADDRESS =
-  '0x0000000000000000000000000000000000000000' as const
 
 export const MOCK_L1_CHAIN: Chain = {
```
kevinhalliday, line 35: General nit: I don't think we need to do capital snake case here. We can just have all our vars / exports be camel case (similar viem / wagmi)

sdk/packages/core/src/api/validateOrder.ts
```diff
@@ -83,6 +86,16 @@ export async function validateOrder(apiBaseUrl: string, encodedOrder: string) {
   return json as ValidationResponse
 }
 
+export async function validateOrder<abis extends OptionalAbis>(
+  apiBaseUrl: string,
+  order: Order<abis>,
+) {
+  return await validateOrderWithEncoded(
+    apiBaseUrl,
+    encodeOrderForValidation(order),
```
kevinhalliday, line 95: why not just `encodeOrder`? matched `encodeQuote` pattern. do we have a different variation of order encoding

sdk/packages/core/src/api/validateOrder.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  const response = {
+    error: {
+      code: 1,
+      message: 'an error',
+    },
+  }
+  vi.spyOn(api, 'fetchJSON').mockResolvedValue(response)
+  await expect(validateOrder('http://localhost', testOrder)).resolves.toBe(
+    response,
+  )
```
kevinhalliday, line 60: do we want to resolve on supported error? or throw on supported error.

my understanding is that react query pattern is to throw on errors, so you can use their error states

sdk/packages/react-next/src/hooks/useParseOpenEvent.ts
```diff
@@ -0,0 +1,37 @@
...
+export function useParseOpenEvent(
+  params: UseParseOpenEventParams,
+): UseParseOpenEventReturn {
+  const { status, logs } = params
+  const eventData = useMemo(() => {
+    if (!logs || status !== 'success') return
+    try {
+      return { resolvedOrder: parseOpenEvent(logs) }
+    } catch (error) {
+      return { error: error as ParseOpenEventError }
```
kevinhalliday, line 29: should we confirm error type here

sdk/packages/react-next/src/hooks/useParseOpenEvent.ts
```diff
@@ -0,0 +1,37 @@
...
+
+type UseParseOpenEventParams = {
+  status: UseWaitForTransactionReceiptReturnType['status']
+  logs?: Log[]
+}
+
+type UseParseOpenEventReturn = {
+  resolvedOrder: ParseOpenEventReturn | undefined
+  error: ParseOpenEventError | undefined
+}
```
kevinhalliday, line 18: type narrowing?

sdk/packages/core/src/api/validateOrder.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  { rejected: true },
+  { rejected: true, rejectReason: 'a reason' },
+  { rejecetd: true, rejectDescription: 'a description' },
+])('behaviour: throws if response is not valid: %s', async (mockReturn) => {
+  vi.spyOn(api, 'fetchJSON').mockResolvedValue(mockReturn)
+
+  const expectRejection = expect(async () => {
+    await validateOrder('http://localhost', testOrder)
+  }).rejects
+  await expectRejection.toBeInstanceOf(Error)
```
kevinhalliday, line 87: do we want to test our specific error types?

sdk/packages/core/src/contracts/parseOpenEvent.ts
PaulLeCam: 👍 updated

sdk/packages/react-next/src/hooks/useDidFillOutbox.ts
PaulLeCam: This error should never throw, it's only a check for TS to assert the parameters needed for the call are set.
I updated it to make it more explicit it's related to the `enabled` flag, what do you think please?

sdk/packages/react-next/src/hooks/useQuote.ts
```diff
@@ -0,0 +1,80 @@
...
+  const { apiBaseUrl } = useOmniContext()
+  const { enabled, ...quoteParams } = params
+
+  const request = encodeQuote(quoteParams)
+  const query = useQuery<Quote, QuoteError>({
+    queryKey: ['quote', request],
+    queryFn: async () => await getQuoteWithEncoded(apiBaseUrl, request),
+    enabled,
+  })
+
```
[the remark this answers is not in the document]
PaulLeCam: Yes I agree, I don't like it either. To me it's really the react-specific implementation details that are leaking to core.
I think the better solution here would be to have the `queryKey` defined based on the relevant hook parameters without needing specific encoding, so we wouldn't need this 2-step process.
That would likely be a larger refactoring though, so for a first version I mostly wanted to make the core package expose everything needed by the react one.

I removed the `With` from the call functions names and added the `Request` suffix to the encoding functions, what do you think please for a first version until we refactor the react logic further?

sdk/packages/test-utils/src/constants.ts
PaulLeCam: Good idea, I renamed them

sdk/packages/core/src/api/validateOrder.ts
PaulLeCam: `encodeOrder` was already used to encode the order data for the contract, I renamed it to `encodeOrderData` and `encodeOrderForValidation` to `encodeOrderRequest` so hopefully it's more explicit?

sdk/packages/core/src/api/validateOrder.test.ts
PaulLeCam: Ideally yes, but again I wanted to keep the core aligned to the react package's expectations, which handles the `error` object from the response. I think we need a bigger refactoring of error handling, especially having clear differences between input errors, known server errors and possible exceptions.

sdk/packages/react-next/src/hooks/useParseOpenEvent.ts
PaulLeCam: I reused the type as currently used by the react package, but yes I can change it if you want?
To me for this kind of pattern it would be best to use something like [typescript-result](https://github.com/everweij/typescript-result) for example, rather than ad-hoc types every time.

sdk/packages/core/src/api/validateOrder.test.ts
PaulLeCam: Yes eventually I think we should be able to differentiate between input errors (can't encode the order) and API errors or exceptions, I think we should consider it as part of a larger refactoring though.

sdk/packages/react-next/src/hooks/useDidFillOutbox.ts
ga-reth: could we instead use an `invariant`? something like:

```
const invariant(
  condition: any, 
  message: string = "Invariant violation: condition not met"
) => asserts condition {
  if (!condition) {
    throw new Error(message);
  }
}
````

then we can use it like `invariant(client && contracts && resolvedOrder, 'Invalid query params')`

Comment (ga-reth) [reply sdk/packages/react-next/src/hooks/useDidFillOutbox.ts]: or rename invariant to assertExists or something similar to be more explicit

**Comment by ga-reth** [comment]
noticed you didn't include the barrel file from `/react`, these are helpful to simplify imports for consumers 
```
import { useOrder } from '@omni-network/react/useOrder';
// with barrel
import { useOrder } from '@omni-network/react';
```
i'll add since you're OOO

sdk/packages/react-next/src/hooks/useQuote.ts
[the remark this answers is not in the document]
kevinhalliday: I like that

ga-reth (approved):

PR #3687, opened 2025-04-14
fix(ethclient/headerdb): prevent deadlock

chmllr:
Fixes deadlock in `headercachedb` since `cosmos-db.MemDB` deosn't support iterating and writing. It can deadlock when iteration is large.

issue: #3615


monitor/xmonitor/monitor.go
```diff
@@ -21,18 +20,6 @@ func Start(
 	cprovider cchain.Provider,
 	rpcClients map[uint64]ethclient.Client,
 ) error {
-	{
```
corverroos, line 24: nit: can remove this change as this only relates to monitor which isn't applicable to releases

corverroos (approved):

PR #3688, opened 2025-04-14
chore(lib/ethclient): add http timeout

corverroos:
Add 30s timeout to all ethclient http requests. This should fix production issue where clients are stuck waiting for response from server.

issue: none


corverroos:
Stuck thread dump
image

PR #3692, opened 2025-04-14
fix(halo): relax block validation around upgrades

chmllr:
Introduces the `strictMode` flag to the block payload validation which defines whether the payload should be rejected if it contains no withdrawals, even though we have some eligible ones. This is required, because after we upgrade the network to a new binary, the first block it executes is built by the previous binary, which cannot include the withdrawals.

This change should be deleted after drake is rolled out.

issue: none

### DIFF
--- a/octane/evmengine/keeper/keeper.go
+++ b/octane/evmengine/keeper/keeper.go
@@ -183,8 +183,8 @@ func (k *Keeper) parseAndVerifyProposedPayload(ctx context.Context, msg *types.M
 		return engine.ExecutableData{}, errors.Wrap(err, "eligible withdrawals")
 	}
 
-	// Allow 0 payload withdrawals for first block after drake upgrade.
-	// Since block built and verified by magellen, but executed by drake.
+	// Allow 0 payload withdrawals until then ext release after drake.
+	// Since block built and verified by magellan, but executed by drake.
 	// So do strict validation unless in FinalizeBlock and payload is empty.
 	strictWithdrawals := sdk.UnwrapSDKContext(ctx).ExecMode() == sdk.ExecModeProcessProposal || len(payload.Withdrawals) > 0
 
--- a/octane/evmengine/keeper/keeper.go
+++ b/octane/evmengine/keeper/keeper.go
@@ -183,9 +183,13 @@ func (k *Keeper) parseAndVerifyProposedPayload(ctx context.Context, msg *types.M
 		return engine.ExecutableData{}, errors.Wrap(err, "eligible withdrawals")
 	}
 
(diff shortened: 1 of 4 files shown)


octane/evmengine/keeper/keeper.go
```diff
@@ -183,7 +188,10 @@ func (k *Keeper) parseAndVerifyProposedPayload(ctx context.Context, msg *types.M
 		return engine.ExecutableData{}, errors.Wrap(err, "eligible withdrawals")
 	}
 
-	if !withdrawalsEqual(payload.Withdrawals, eligibleWithdrawals) {
+	// If eligible withdrawals don't match with what's in the payload,
+	// only throw an error in strict mode or if the payload is non-empty
+	if !withdrawalsEqual(payload.Withdrawals, eligibleWithdrawals) &&
```
corverroos, line 193: I'd make this slightly more readable (also no need to pass in the flag, can infer from context):
```
// Allow 0 payload withdrawals for first block after drake upgrade.
	// Since block built and verified by magellen, but executed by drake.
	// So do strict validation unless in FinalizeBlock and payload is empty.
	strictWithdrawals := sdk.UnwrapSDKContext(ctx).ExecMode() == sdk.ExecModeProcessProposal || len(payload.Withdrawals) > 0

	if !withdrawalsEqual(payload.Withdrawals, eligibleWithdrawals) && strictWithdrawals {
		return engine.ExecutableData{}, errors.New("mismatch with eligible withdrawals")
	}
```

corverroos (commented):
lets add a unit test for this please

octane/evmengine/keeper/keeper.go
```diff
@@ -183,9 +183,13 @@ func (k *Keeper) parseAndVerifyProposedPayload(ctx context.Context, msg *types.M
 		return engine.ExecutableData{}, errors.Wrap(err, "eligible withdrawals")
 	}
 
-	if !withdrawalsEqual(payload.Withdrawals, eligibleWithdrawals) {
-		height := sdk.UnwrapSDKContext(ctx).BlockHeight()
-		return engine.ExecutableData{}, errors.New("mismatch with eligible withdrawals", "height", height)
+	// Allow 0 payload withdrawals for first block after drake upgrade.
+	// Since block built and verified by magellen, but executed by drake.
```
fabtreb, line 187: Magellan

Also check overall sentence structure, something like

"Allow 0 payload withdrawals for first block after the Drake upgrade, since blocks are built and verified by Magellan, but executed by Drake." ?

chmllr: I'll adjust the comment: it's fine to do this until the next release after drake.

corverroos (approved):

Issue #3618, opened 2025-04-07
Review and enhance integration tests to ensure we have full coverage against breaking changes


ga-reth:
#### Problem to Solve

- integration tests didn't catch a breaking change in the contract

#### Proposed Solution

- review existing tests and ensure we have enough coverage to catch these in the short 
term

PR #3704, opened 2025-04-15
test(sdk): increase integration test coverage

ga-reth:
- move to camel case naming
- add balance checks to filled lifecycle test
- add more assertions to existing tests
- replace some if(condition) checks in tests with throwing condition, otherwise the tests may be false positive
- tests for `useParseOpenEvent`, `useGetOrder`, `useOmniContracts` - remaining hooks are tested implicitly already (as are these, but wanted to add explicit testing of return types)
- assert data shape of returned value from various hooks to bolster defence against contract changes

issue: #3618


kevinhalliday (approved):
👏

PR #3710, opened 2025-04-15
feat(contracts/solve): integrate hyperlane

Zodomo:
Integrated Hyperlane as an alternative messaging provider to help us expand to more chains quickly in the short term. The `SolverNetInbox` can receive fulfillment attestations regardless of if they are routed via Omni Core or Hyperlane. The `SolverNetOutbox` can be configured to use either Omni Core or Hyperlane to route fulfillment attestations to an Inbox on any given chain. There are no breaking changes for the solver, this is a backwards-compatible additive change. We use a customized Hyperlane `MailboxClient` contract suited for our use case, and also use refactored mock contracts to test against to eliminate dependency issues.

issue: #3741


contracts/solve/src/SolverNetOutbox.sol
```diff
@@ -240,6 +255,41 @@ contract SolverNetOutbox is OwnableRoles, ReentrancyGuard, Initializable, Deploy
...
 
+    /**
+     * @notice Route a message to the inbox.
+     * @param orderId  ID of the order.
+     * @param fillHash Hash of the fill instructions origin data.
+     * @param claimant Address specified by the filler to claim the order (msg.sender if none specified).
+     * @param fillData ABI decoded fill originData.
+     * @return fee     Fee amount in native currency.
+     */
+    function _routeMsg(bytes32 orderId, bytes32 fillHash, address claimant, SolverNet.FillOriginData memory fillData)
```
kevinhalliday, line 266: Maybe `_routeMarkFilled`, because that's the only msg type we are routing

kevinhalliday (approved):
This is great

PR #3723, opened 2025-04-16
Add Replay test

dapslegend:
issue: none or #12345


dapslegend:
@corverroos i want to draw your attention here to my test

corverroos (changes requested):
Thanks for the test, but empty proposals are allowed by design, so this isn't a "replay attack" but simply "two consecutive empty proposals" which is allowed.

corverroos:
Oh, btw, you are welcome to submit any possible bugs via our bug bounty program: https://immunefi.com/bug-bounty/omni-network/information/

PR #3727, opened 2025-04-16
feat(contracts/solve): exploratory layerzero integration

Zodomo:
Similar to #3710 with Hyperlane, I explored integrating LayerZero into the SolverNet contracts. Here the integration is structurally similar to Hyperlane, as would be other messaging protocol integrations. See `SolverNetInboxLZ` and `SolverNetOutboxLZ` for revised contracts that collectively support sending/receiving fulfillment settlements via either Omni Core or LayerZero.

issue: none


Zodomo:
Closing PR as we are moving forward with Hyperlane for now. Branch will remain.

PR #3738, opened 2025-04-17
chore(lib/log): add test logger for e2e tests

corverroos:
Adds a `log.WihtTestLogger` function that tests can use to group test logs.

issue: none


kevinhalliday (approved):
🙏 🙏

PR #3748, opened 2025-04-21
fix(solver/app): retry on invalid transitions

corverroos:
Detect "invalid transitions" and retry, assuming that this is due to RPC providers providing stale data, so retries should fix it.

Also write a `TestManualClaim` utility to manually claim orders.

Also improve flowgen logging and decrease flowgen role min balance. 

issue: none

### DIFF
--- a/solver/app/processor.go
+++ b/solver/app/processor.go
@@ -43,7 +43,10 @@ func newEventProcFunc(deps procDeps, chainID uint64) eventProcFunc {
 			"status", order.Status,
 		)
 
-		if event.Status != order.Status {
+		if !event.Status.ValidTarget(order.Status) {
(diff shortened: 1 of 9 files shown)


solver/app/processor.go
```diff
@@ -43,7 +43,10 @@ func newEventProcFunc(deps procDeps, chainID uint64) eventProcFunc {
 			"status", order.Status,
 		)
 
-		if event.Status != order.Status {
+		if !event.Status.ValidTarget(order.Status) {
+			// Invalid order transition can occur when RPCs return stale data, so just retry for now.
+			return errors.New("invalid order transition", "event_status", event.Status.String())
+		} else if event.Status != order.Status {
```
corverroos, line 49: main change

kevinhalliday (approved):

PR #3756, opened 2025-04-22
chore(lib/uniswap): add usdt swap route

kevinhalliday:
Add USDT -> USDC route.

issue: none

### DIFF
--- a/lib/uniswap/swap_test.go
+++ b/lib/uniswap/swap_test.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"github.com/omni-network/omni/e2e/app/eoa"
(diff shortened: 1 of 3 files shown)


lib/uniswap/swap_test.go
```diff
@@ -77,6 +81,11 @@ func TestSwapToUSDC(t *testing.T) {
 			asset:    tokens.WSTETH,
 			amountIn: bi.Ether(1),
 		},
+		{
+			name:     "USDT to USDC",
+			asset:    tokens.USDT,
+			amountIn: bi.Ether(1),
+		},
```
kevinhalliday, line 88: This isn't working yet. Need to get the slot index FundERC20 option from other PR in and use that

PR #3757, opened 2025-04-22
chore(sdk): encode bigint prior to stringify

ga-reth:
- encode bigints prior to JSON.stringify
- this fix is needed since we found browser extensions (e.g. initia wallet) may introduce a toJSON method to BigInt
- by doing so, the replacer passed to JSON.stringify to convert big ints to hex strings is bypassed

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#:~:text=Attempting%20to%20serialize,by%20the%20user.

issue: none

### DIFF
--- a/sdk/packages/core/src/utils/toJSON.ts
+++ b/sdk/packages/core/src/utils/toJSON.ts
@@ -1,6 +1,33 @@
 import { toHex } from 'viem'
 
-export const toJSON = (value: unknown) =>
-  JSON.stringify(value, (_, value) =>
-    typeof value === 'bigint' ? toHex(value) : value,
-  )
+const convertBigIntsToHex = (value: unknown): unknown => {
+  if (typeof value === 'bigint') {
+    return toHex(value)
+  }
+
+  if (Array.isArray(value)) {
+    return value.map(convertBigIntsToHex)
+  }
+
+  if (typeof value === 'object' && value !== null) {
+    const newValue: Record<string, unknown> = {}
+    for (const key in value) {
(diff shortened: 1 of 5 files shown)


sdk/packages/core/src/utils/toJSON.ts
```diff
@@ -1,6 +1,33 @@
...
+  return value
+}
+
+export const toJSON = (value: unknown) => {
+  const processedValue = convertBigIntsToHex(value)
+  // stringify w/out the replacer callback since we've seen browser
+  // extensions bypass the replacer by attaching a toJSON method to BigInt
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#:~:text=Attempting%20to%20serialize,by%20the%20user
+  return JSON.stringify(processedValue)
+}
```
kevinhalliday, line 33:
```suggestion
export const toJSON = (value: unknown) => {
  // stringify w/out the replacer callback since we've seen browser
  // extensions bypass the replacer by attaching a toJSON method to BigInt
  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#:~:text=Attempting%20to%20serialize,by%20the%20user
  return JSON.stringify(convertBigIntsToHex(value))
}
```

? 😬

kevinhalliday (approved):

sdk/packages/core/src/utils/toJSON.ts
ga-reth: haha good point

PR #3778, opened 2025-04-23
chore(lib/txmgr): double gas limits but cap

corverroos:
Double gas limits (instead of adding 20%), but cap increase to 500K. 

This aims to fix continued symbiotic tx reverts on omega due to out-of-gas.

issue: none

### DIFF
--- a/lib/txmgr/txmgr.go
+++ b/lib/txmgr/txmgr.go
@@ -33,10 +33,12 @@ const (
 	// submissions reserve nonces and reach upstream geth nodes out of order.
 	retryNonceTooHigh = 3
 
-	// gasFactorNum is the numerator by which to increase the gas limit for transactions.
-	gasFactorNum uint64 = 12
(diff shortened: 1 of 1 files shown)


lib/txmgr/txmgr.go
```diff
@@ -33,10 +33,12 @@ const (
...
 	retryNonceTooHigh = 3
 
-	// gasFactorNum is the numerator by which to increase the gas limit for transactions.
-	gasFactorNum uint64 = 12
-	// gasFactorDenom is the denominator by which to increase the gas limit for transactions.
-	gasFactorDenom uint64 = 10
+	// gasBumpNum is the numerator by which to increase the gas limit.
+	gasBumpNum uint64 = 1
+	// gasBumpDenom is the denominator by which to increase the gas limit.
+	gasBumpDenom uint64 = 1
```
kevinhalliday, line 39: should be `2/1`?

kevinhalliday (approved):

PR #3800, opened 2025-04-24
feat(halo): reward withdrawals

chmllr:
- Adds rewards withdrawals functionality to halo
- Adds a new predeploy contract Distribute.sol
- Implements e2e tests for reward withdrawals

issue: #80


corverroos (approved):

kevinhalliday (approved):

contracts/core/src/octane/Distribution.sol
```diff
@@ -0,0 +1,55 @@
...
+        _burnFee();
+        emit Withdraw(msg.sender, validator);
+    }
+
+    /**
+     * @notice Burn the fee, requiring it be sent with the call
+     */
+    function _burnFee() internal {
+        require(msg.value >= Fee, "Distribution: insufficient fee");
+        payable(BurnAddr).transfer(msg.value);
```
Zodomo, line 53: Best practice is to use call for this:
`payable(BurnAddr).call{value: msg.value}("")`

Or inherit `SafeTransferLib` from `solady` and use that.

kevinhalliday: transfer fine here because we're sending to a known empty addr

PR #3801, opened 2025-04-24
chore(sdk): optional queryOpts param

ga-reth:
- optional queryOpts param for quote and validate

issue: none

### DIFF
--- a/sdk/packages/react/src/hooks/useQuote.ts
+++ b/sdk/packages/react/src/hooks/useQuote.ts
@@ -1,4 +1,4 @@
-import { type UseQueryResult, useQuery } from '@tanstack/react-query'
+import { type UseQueryOptions, type UseQueryResult, useQuery } from '@tanstack/react-query'
 import { useMemo } from 'react'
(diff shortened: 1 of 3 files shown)


sdk/packages/react/src/hooks/useQuote.ts
```diff
@@ -62,6 +70,7 @@ export function useQuote(params: UseQuoteParams): UseQuoteResult {
     queryKey: ['quote', request],
     queryFn: async () => doQuote(apiBaseUrl, request),
     enabled,
+    ...params.queryOpts,
   })
```
PaulLeCam, line 74: Considering `queryOpts` is optional, I think we need to ensure it's an object before using the spread, ex `...(params.queryOpts ?? {})`.
Also, I'd put these options spread before the other parameters in the object to ensure they don't accidentally override the other parameters, ex `useQuery({ ...opts, queryFn: ... })`

PaulLeCam (dismissed):

kevinhalliday (approved):

PR #3812, opened 2025-04-25
chore(sdk): add unit tests to core package

PaulLeCam:
Unit tests for contract interactions

issue: none

### DIFF
--- a/sdk/packages/core/src/contracts/openOrder.test.ts
+++ b/sdk/packages/core/src/contracts/openOrder.test.ts
@@ -1,4 +1,4 @@
-import { testOrder } from '@omni-network/test-utils'
+import { mockL1Client, testOrder } from '@omni-network/test-utils'
 import { type Client, zeroAddress } from 'viem'
 import { expect, test, vi } from 'vitest'
 import { inboxABI } from '../constants/abis.js'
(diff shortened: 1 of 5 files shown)


sdk/packages/core/src/contracts/openOrder.test.ts
```diff
@@ -0,0 +1,151 @@
+import { testOrder } from '@omni-network/test-utils'
```
ga-reth, line 1: mind adding a test around the encodeOrder behaviour? aka what happens if bad inputs are provided

sdk/packages/core/src/contracts/openOrder.test.ts
```diff
@@ -0,0 +1,151 @@
...
+})
+
+test('behaviour: sets the order value when the deposit token address is zero', async () => {
+  writeContract.mockResolvedValueOnce('0xtxHash')
+  vi.spyOn(encode, 'encodeOrderData').mockReturnValueOnce('0xencodedOrder')
+
+  const client = {
+    account: {},
+    chain: {},
+  } as Client
```
ga-reth, line 94: want to put this in test-utils?

ga-reth (commented):
we should probably add the same assertion to verify the client exists that we have in openOrder to the other actions I guess? 

maybe we can pull that assertion into a shared util for all actions? 

then add unit tests for that where applicable

sdk/packages/core/src/contracts/openOrder.test.ts
PaulLeCam: :+1: I replaced by the `mockL1Client` already defined there

PaulLeCam:
> we should probably add the same assertion to verify the client exists that we have in openOrder to the other actions I guess?

We don't need to check that the client exists as it's already specified in the params type, the check in `openOrder` is that the client contains an `account`.
I agree if we needed this check in multiple places, would be good to extract as a dedicated assertion maybe, but at the moment we only need it in `openOrder` so it seems premature to extract the logic, what do you think?

ga-reth:
@PaulLeCam okay nice, missed the requirement for an account, sounds good then!

ga-reth (approved):

PR #3819, opened 2025-04-28
fix(halo): add noop migration handlers

chmllr:
Add noop migration handlers.

issue: none


corverroos (changes requested):
Have a look at the blog, I don't think this is the way

corverroos (approved):

PR #3821, opened 2025-04-28
chore(sdk): unit tests for parsing logic in core

PaulLeCam:
Brings logic from the react package to core

issue: none

### DIFF
--- a/sdk/packages/core/src/contracts/parseInboxStatus.test.ts
+++ b/sdk/packages/core/src/contracts/parseInboxStatus.test.ts
@@ -2,15 +2,18 @@ import { expect, test } from 'vitest'
 import type { GetOrderReturn } from './getOrder.js'
 import { parseInboxStatus, status } from './parseInboxStatus.js'
 
-test.each(Object.entries(status))('code %i returns "%s"', (status, text) => {
-  expect(
-    parseInboxStatus({
-      order: [{}, { status }, {}] as unknown as GetOrderReturn,
-    }),
-  ).toBe(text)
-})
+test.each(Object.entries(status))(
+  'default: code %i returns "%s"',
+  (status, text) => {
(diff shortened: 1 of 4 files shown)


sdk/packages/core/src/contracts/parseInboxStatus.test.ts
```diff
@@ -0,0 +1,24 @@
+import { expect, test } from 'vitest'
+import type { GetOrderReturn } from './getOrder.js'
+import { parseInboxStatus, status } from './parseInboxStatus.js'
+
+test.each(Object.entries(status))('code %i returns "%s"', (status, text) => {
```
ga-reth, line 5: mind adding `default:`?

sdk/packages/core/src/contracts/parseOpenEvent.ts
```diff
@@ -37,6 +37,9 @@ export function parseOpenEvent(logs: Log[]): ResolvedOrder {
 
     return openEvent.args.resolvedOrder
   } catch (error) {
+    if (error instanceof Error) {
+      throw error
+    }
```
ga-reth, line 42: mind explaining why we need this? previous logic makes sure we throw with an expected error type

PaulLeCam: If the branch at line 23 throws a `ParseOpenEventError`, this logic would create a new `ParseOpenEventError` wrapping the existing one.
I updated the logic to check for `ParseOpenEventError` specifically rather than a generic error but I still think this error management logic is fragile, I hope we can work on a better system soon.

ga-reth: nice thanks

ga-reth (approved):

PR #3852, opened 2025-04-29
fix(sdk): integration tests concurrency

PaulLeCam:
Ensures concurrent tests use different accounts

issue: none


sdk/integration-tests/suites/flows.tsx
```diff
@@ -62,32 +56,49 @@ describe.concurrent('ERC20 OMNI to native OMNI transfer orders', () => {
...
 
-  describe.concurrent('default: succeeds with native deposit', () => {
+  describe.sequential('default: succeeds with native deposit', () => {
+    const account = getNextAccount()
+    const srcClient = createClient({ account, chain: mockL1Chain })
+    const destClient = createClient({ chain: omniDevnetChain })
     const amount = parseEther('10')
     const order: AnyOrder = {
-      owner: account.address,
+      owner: testAccount.address,
```
ga-reth, line 81: you're still creating account but using testAccount here btw

ga-reth (approved):

sdk/integration-tests/suites/flows.tsx
PaulLeCam: good catch, thanks! I fixed it

PR #3870, opened 2025-04-30
feat(sdk): higher-level core APIs

PaulLeCam:
Opening as draft to gather feedback on APIs, here are the main changes:

- rename `openOrder` to `sendOrderTransaction` to be able to use `openOrder` for a higher-level flow
- add `waitForOrderOpen(): Promise<ResolvedOrder>` to watch for the order `open` event given the transaction hash returned by `sendOrderTransaction`
- new `openOrder` function implementing the `validateOrder` + `sendOrderTransaction` + `waitForOrderOpen` flow
- add `waitForOrderStatus` function to track reaching the wanted `InboxStatus` for a given order
- add `generateOrder(): AsyncGenerator<OrderState>` to iteratively track the order lifecycle: validate -> send transaction -> wait for open event -> wait for settled status

issue: none


sdk/packages/core/src/contracts/waitForOrderStatus.ts
```diff
@@ -0,0 +1,34 @@
...
+import { type InboxStatus, parseInboxStatus } from './parseInboxStatus.js'
+import { watchOrder } from './watchOrder.js'
+
+export type WaitForOrderStatusParameters<Status extends InboxStatus> =
+  GetOrderParameters & {
+    status: Status | Status[]
+    pollingInterval?: number
+  }
+
+export function waitForOrderStatus<Status extends InboxStatus>(
```
ga-reth, line 11: nice 👌

what do you think about - instead of watching for a given status/statuses, we instead just `watchOrder` or `waitOrder` and resolve upon reaching a terminal state ( basically one of `['closed', 'filled', 'rejected']` ) ?

reduces leaking impl detail to users and simplifies the API

ga-reth: we could also then emit status changes which would be cool

sdk/packages/core/src/contracts/watchOrder.ts
```diff
@@ -0,0 +1,37 @@
...
+  onOrder: (order: GetOrderReturn) => void
+  onError?: (error: Error) => void
+  pollingInterval?: number
+}
+
+export type WatchOrderReturn = () => void
+
+export function watchOrder(params: WatchOrderParameters): WatchOrderReturn {
+  const { onError, onOrder, pollingInterval, ...getOrderParams } = params
+  return watchBlockNumber(params.client, {
```
ga-reth, line 19: instead of watching blocks, could we watch events instead (`watchEvent`)? that way we can scope the query a decent amount by only watching the inbox contract events

get a nice benefit of also being able to batch the events between the intervals

sdk/packages/core/src/flows/generateOrder.ts
```diff
@@ -0,0 +1,53 @@
...
+import { waitForOrderStatus } from '../contracts/waitForOrderStatus.js'
+import type { OptionalAbis } from '../types/abi.js'
+
+export type GenerateOrderParameters<abis extends OptionalAbis> =
+  SendOrderTransactionParameters<abis> & {
+    apiBaseUrl: string
+    pollingInterval?: number
+  }
+
+export type SettledStatus = 'rejected' | 'closed' | 'filled'
```
ga-reth, line 18: small nit - could we use `TerminalStatus` or something similar? settled typically refers to a success completion

sdk/packages/core/src/flows/generateOrder.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export type OrderState =
+  | { status: 'valid' }
+  | { status: 'sent'; txHash: Hex }
+  | { status: 'open'; txHash: Hex; order: ResolvedOrder }
+  | { status: SettledStatus; txHash: Hex; order: ResolvedOrder }
+
+export async function* generateOrder<abis extends OptionalAbis>(
```
ga-reth, line 26: 🤌 nice, i love this generator pattern

sdk/packages/core/src/flows/generateOrder.ts
```diff
@@ -0,0 +1,53 @@
...
+  type SendOrderTransactionParameters,
+  sendOrderTransaction,
+} from '../contracts/sendOrderTransaction.js'
+import { waitForOrderOpen } from '../contracts/waitForOrderOpen.js'
+import { waitForOrderStatus } from '../contracts/waitForOrderStatus.js'
+import type { OptionalAbis } from '../types/abi.js'
+
+export type GenerateOrderParameters<abis extends OptionalAbis> =
+  SendOrderTransactionParameters<abis> & {
+    apiBaseUrl: string
```
ga-reth, line 14: `apiBaseUrl` here

sdk/packages/core/src/contracts/waitForOrderStatus.ts
```diff
@@ -0,0 +1,34 @@
...
+    Array.isArray(status) ? status : [status]
+  ) as Array<InboxStatus>
+  return new Promise((resolve, reject) => {
+    const stop = watchOrder({
+      ...watchParams,
+      onOrder: (order) => {
+        const status = parseInboxStatus({ order })
+        if (expectedStatus.includes(status)) {
+          stop()
+          resolve(status as Status)
```
ga-reth, line 25: we need to fix this type assertion, I think it was warning you because `not-found` is returned from `parseInboxStatus` but not included in `InboxStatus`

ga-reth: this was added in case an order id is passed in that isn't valid, can either allow that status or update the underlying logic

sdk/packages/core/src/contracts/waitForOrderStatus.ts
ga-reth: should we also add the ability abort the polling? wouldn't work with this current design, but maybe with AbortController or just returning a callback to terminate polling

sdk/packages/core/src/api/validateOrder.ts
```diff
@@ -97,3 +98,27 @@ const isValidateRes = (json: unknown): json is ValidationResponse => {
...
+
+export type AcceptedResult = {
+  accepted: true
+  rejected?: false
+  error?: never
+  rejectReason?: never
+  rejectDescription?: never
+}
+
+export function assertAcceptedResult(
```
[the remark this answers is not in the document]
PaulLeCam: The `validateOrder()`function doesn't use this function, it's only used in the higher-level functions in the core package that aren't used by the react package

sdk/packages/core/src/flows/generateOrder.ts
PaulLeCam: I used the JS promises terminology here where "settled" stands for either "resolved" or "rejected" but no problem, I made the change

sdk/packages/core/src/contracts/waitForOrderStatus.ts
PaulLeCam: Is `not-found` a terminal state please? If so, I think we should add it to the list, but otherwise the behaviour might be confusing, no?

sdk/packages/core/src/contracts/watchOrder.ts
PaulLeCam: Yeah I think we could improve, but do you mind if we implement it like this for now to simplify the logic please? I think for now the best would be to settle on the public APIs so we can make a release, and then we can improve the internals, what do you think?

sdk/packages/core/src/contracts/waitForOrderStatus.ts
PaulLeCam: I updated the logic to be just `waitForOrderClose` as parallel to the `waitForOrderOpen` function, hopefully that's enough to get started and then we can add more granular functions as needed, what do you think?

sdk/packages/core/src/contracts/watchOrder.ts
ga-reth: Yeah no worries! sgtm :)

sdk/packages/core/src/contracts/waitForOrderStatus.ts
ga-reth: yeah i agree its a bit confusing, we could just remove the not-found thing I guess

sdk/packages/core/src/contracts/waitForOrderStatus.ts
ga-reth: yeah parallel naming sgtm!

ga-reth (approved):

PR #3876, opened 2025-04-30
chore(ci): skip rebalance tests if no secret access

kevinhalliday:
Skip rebalance tests if runner does not have access to secrets

issue: none


kevinhalliday:
Must have invalid workflow because these things keep hanging. will try something else

PR #3886, opened 2025-05-01
chore(sdk): pnpm catalog to unify deps

ga-reth:
- needed to bump viem and wagmi version to resolve issues with `Client` type and using a catalog is a lot easier to unify deps across the workspace - open to feedback here though!
- fixed viem and wagmi versions so we can manually bump
- also scoped build command to only run for /packages

issue: none

### DIFF
--- a/sdk/packages/react/package.json
+++ b/sdk/packages/react/package.json
@@ -38,20 +38,22 @@
   "peerDependencies": {
     "@tanstack/react-query": "^5.64.2",
     "react": ">=18",
-    "viem": ">=2.22.10",
-    "wagmi": ">=2.0.0"
+    "viem": ">=2.28.1",
+    "wagmi": ">=2.15.1"
   },
   "devDependencies": {
     "@omni-network/test-utils": "workspace:^",
-    "@testing-library/dom": "^10.4.0",
-    "@testing-library/react": "^16.2.0",
-    "@types/react": "^19.0.8",
-    "@types/react-dom": "^19.0.4",
+    "@testing-library/dom": "catalog:",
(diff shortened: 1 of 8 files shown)


sdk/pnpm-workspace.yaml
```diff
@@ -1,3 +1,15 @@
...
   - packages/*
+
+catalog:
+  "@tanstack/react-query": "^5.69.0"
+  "@testing-library/dom": "^10.4.0"
+  "@testing-library/react": "^16.2.0"
+  "@types/react": "^18.3.1"
+  "@types/react-dom": "^18.3.0"
+  "wagmi": "2.15.1"
+  "viem": "2.28.1"
```
PaulLeCam, line 12: Should use `^` to ensure any v2.x versions are supported
```suggestion
  "wagmi": "^2.15.1"
  "viem": "^2.28.1"
```

sdk/packages/react/package.json
```diff
@@ -38,20 +38,22 @@
   "peerDependencies": {
     "@tanstack/react-query": "^5.64.2",
     "react": ">=18",
-    "viem": ">=2.22.10",
-    "wagmi": ">=2.0.0"
+    "viem": ">=2.28.1",
+    "wagmi": ">=2.15.1"
```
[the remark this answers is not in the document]
ga-reth: good point, will update

sdk/pnpm-workspace.yaml
ga-reth: will do

PaulLeCam (approved):

PR #3887, opened 2025-05-01
chore(sdk): use dedicated test accounts for integration tests

PaulLeCam:
Skipping tests requiring to mint OMNI as the minting function fails with these new accounts

issue: none


sdk/integration-tests/suites/flows.tsx
```diff
@@ -144,6 +154,11 @@ describe.concurrent('ETH transfer orders', () => {
 
       test('using core APIs', async () => {
         const account = getNextAccount()
+        await createAnvilClient(mockL1Chain).setBalance({
+          address: account.address,
+          value: parseEther('10'),
+        })
```
ga-reth, line 160: could we just set balance to 100eth in `beforeAll` instead?

ga-reth (approved):

sdk/integration-tests/suites/flows.tsx
PaulLeCam: Each funding call is for a different account so I don't think that would change anything?
I don't think we can fund multiple accounts in a single call, at least looking at viem's APIs?

Issue #3889, opened 2025-05-01
Migrate Middleman logic into Executor


Zodomo:
#### Problem to Solve
The `SolverNetMiddleman` contract functions realistically do not need to exist in their own contract and can be performed by the `SolverNetExecutor` itself. This should lead to simpler DX as there is only one Executor to worry about.

#### Proposed Solution
Migrate `SolverNetMiddleman` functions into `SolverNetExecutor`.

PR #3891, opened 2025-05-01
feat(contracts/solve): migrate middleman into executor

Zodomo:
Migrated `SolverNetMiddleman` functions into the `SolverNetExecutor` so we can have one "execution" contract rather than two.

issue: #3889


kevinhalliday (approved):
This looks good.

Thoughts on executor calling itself if target is address zero?

PR #3892, opened 2025-05-01
feat(e2e): use hlNetworkFromDef in fund action

Zodomo:
Can't use `app.networkFromDef` in `fund.go` due to `Portals` context not being present here. Decided to supplement the portal values in `hlNetworkFromDef`.

issue: none

### DIFF
--- a/e2e/app/chains.go
+++ b/e2e/app/chains.go
@@ -100,7 +100,7 @@ func hlNetworkFromDef(ctx context.Context, def Definition) (netconf.Network, err
 			Shards:          chain.Shards,
 			AttestInterval:  chain.AttestInterval(def.Testnet.Network),
 			PortalAddress:   portalAddress,
-			DeployHeight:    0, // TODO(zodomo): Can we retrieve the deploy height from the portal registry?
+			DeployHeight:    0, // Portal height isn't needed here as we aren't deploying OmniPortal contracts to Hyperlane networks
 			HasEmitPortal:   true,
 			HasSubmitPortal: true,
 		}
--- a/e2e/app/chains.go
+++ b/e2e/app/chains.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/omni-network/omni/e2e/app/eoa"
 	"github.com/omni-network/omni/e2e/types"
+	"github.com/omni-network/omni/lib/contracts"
 	"github.com/omni-network/omni/lib/contracts/solvernet"
 	"github.com/omni-network/omni/lib/errors"
 	"github.com/omni-network/omni/lib/ethclient/ethbackend"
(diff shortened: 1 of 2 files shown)


e2e/app/chains.go
```diff
@@ -73,3 +76,55 @@ func AddSolverNetworkAndBackends(ctx context.Context, network netconf.Network, e
...
+		return netconf.Chain{
+			ID:              chain.ChainID,
+			Name:            chain.Name,
+			BlockPeriod:     chain.BlockPeriod,
+			Shards:          chain.Shards,
+			AttestInterval:  chain.AttestInterval(def.Testnet.Network),
+			PortalAddress:   portalAddress,
+			DeployHeight:    0, // Portal height isn't needed here as we aren't deploying OmniPortal contracts to Hyperlane networks
+			HasEmitPortal:   true,
+			HasSubmitPortal: true,
```
kevinhalliday, line 105: These should be false for hl chains

kevinhalliday (approved):

e2e/app/chains.go
Zodomo: ah good catch

e2e/app/chains.go
```diff
@@ -73,3 +76,66 @@ func AddSolverNetworkAndBackends(ctx context.Context, network netconf.Network, e
...
+			Name:            chain.Name,
+			BlockPeriod:     chain.BlockPeriod,
+			Shards:          shards,
+			AttestInterval:  attestInterval,
+			PortalAddress:   portalAddress,
+			DeployHeight:    0, // Portal height isn't needed here as we aren't deploying OmniPortal contracts to Hyperlane networks
+			HasEmitPortal:   hasEmitPortal,
+			HasSubmitPortal: hasSubmitPortal,
+		}
+	}
```
kevinhalliday, line 118: just return, less vars. something like

```golang
newChain := func(chain types.EVMChain) netconf.Chain {
	if solvernet.IsHLChain(chain.ChainID) {
		return netconf.Chain{
			ID:              chain.ChainID,
			Name:            chain.Name,
			BlockPeriod:     chain.BlockPeriod,
	}


	return netconf.Chain{
		ID:              chain.ChainID,
		Name:            chain.Name,
		BlockPeriod:     chain.BlockPeriod,
		Shards:          shards,
		AttestInterval:  chain.AttestInterval(def.Testnet.Network),
		PortalAddress:   addrs.Portal,
		HasEmitPortal:   true
		HasSubmitPortal: true,
	}
}

```

kevinhalliday (approved):

e2e/app/chains.go
Zodomo: ah thats nifty, will do that instead

PR #3896, opened 2025-05-02
chore(sdk): bump pnpm version

ga-reth:
- need to bump to 9.12.1 as it introduced a fix to allow scoped packages to be used in workspace catalog

issue: none


PaulLeCam (approved):
LGTM but it seems it changed the formatting of the files?

ga-reth:
@PaulLeCam good catch, have fixed

PR #3899, opened 2025-05-02
docs(sdk): add docs for core

PaulLeCam:
Also includes a test for the `generateOrder` function to ensure it works as documented

issue: none

### DIFF
--- a/sdk/integration-tests/suites/flows.tsx
+++ b/sdk/integration-tests/suites/flows.tsx
@@ -179,7 +179,7 @@ describe.concurrent('ETH transfer orders', () => {
           address: account.address,
         })
 
-        const quote = await getQuote(quoteParams, 'devnet')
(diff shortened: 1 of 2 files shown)


ga-reth:
do you mind matching the format of the hooks docs? otherwise the two are inconsistent

- Usage
- Parameters
- Return
- Example (s)

maybe we can drop the types section

ga-reth:
did you want to mention and link to generateOrder in the modular apis? and mention it can be used as a more comprehensive approach

PaulLeCam:
> do you mind matching the format of the hooks docs? otherwise the two are inconsistent
> 
>     * Usage
> 
>     * Parameters
> 
>     * Return
> 
>     * Example (s)
> 
> 
> maybe we can drop the types section

I tried following this pattern but, unlike the React SDK, some functions take more than one argument, so `parameters` can be confusing I think?

ga-reth (approved):

Issue #3907, opened 2025-05-02
Set all routes on all SolverNet Inboxes and Outboxes


Zodomo:
#### Problem to Solve
As we expand our network coverage to support new chains via Hyperlane, we need to configure these new routes in all SolverNet Inbox and Outbox contracts.

#### Proposed Solution
Implement a command that can be triggered via a GitHub action that automatically configures routes on all Inbox and Outbox contracts in the extended SolverNet network.

PR #3908, opened 2025-05-02
feat(e2e): configure solvernet routes

Zodomo:
Implemented a `set-solvernet-routes` command that configures all routes in the Inbox/Outbox contracts for all chains in the extended solvernet network.

issue: #3907


e2e/app/solvernetrouting.go
```diff
@@ -0,0 +1,148 @@
...
+
+	tx, err := solverNetInbox.SetOutboxes(txOpts, remoteChainIDs, remoteOutboxes)
+	if err != nil {
+		return errors.Wrap(err, "set outboxes on inbox", "chain", backend.Name())
+	}
+
+	_, err = backend.WaitMined(ctx, tx)
+	if err != nil {
+		return errors.Wrap(err, "wait mined set outboxes on inbox", "chain", backend.Name())
+	}
```
kevinhalliday, line 125: try to have shorter error messages. You'll note this is a pattern we follow elsewhere. We want messages to be precise enough where we can follow wrapped errors to their source, but not more verbose thent hat

```golang
tx, err := solverNetInbox.SetOutboxes(txOpts, remoteChainIDs, remoteOutboxes)
	if err != nil {
		return errors.Wrap(err, "set outboxes", "chain", backend.Name())
	}

	_, err = backend.WaitMined(ctx, tx)
	if err != nil {
		return errors.Wrap(err, "wait mined", "chain", backend.Name())
	}
```

e2e/app/solvernetrouting.go
```diff
@@ -0,0 +1,148 @@
...
+		return false, errors.Wrap(err, "solvernetinbox not deployed", "chain", backend.Name())
+	}
+
+	isDeployed, err = contracts.IsDeployed(ctx, backend, outbox)
+	if !isDeployed || err != nil {
+		return false, errors.Wrap(err, "solvernetoutbox not deployed", "chain", backend.Name())
+	}
+
+	return true, nil
+}
```
kevinhalliday, line 108: `err != nil` and `isDeployed` are different here 

`err != nil` means something went wrong making the call, we don't now if it's deployed or not, and we need to return that error

e2e/app/solvernetrouting.go
```diff
@@ -0,0 +1,148 @@
...
+	tx, err := solverNetOutbox.SetInboxes(txOpts, remoteChainIDs, remoteInboxConfigs)
+	if err != nil {
+		return errors.Wrap(err, "set inboxes on outbox", "chain", backend.Name())
+	}
+
+	_, err = backend.WaitMined(ctx, tx)
+	if err != nil {
+		return errors.Wrap(err, "wait mined set inboxes on outbox", "chain", backend.Name())
+	}
+
```
kevinhalliday, line 146: let's log tx hashes and block numbers

e2e/app/solvernetrouting.go
```diff
@@ -0,0 +1,148 @@
...
+		err = configureInbox(ctx, backend, txOpts, addrs.SolverNetInbox, remoteChainIDs, remoteOutboxes)
+		if err != nil {
+			return errors.Wrap(err, "configure inbox", "chain", chain.Name)
+		}
+
+		err = configureOutbox(ctx, backend, txOpts, addrs.SolverNetOutbox, remoteChainIDs, remoteInboxConfigs)
+		if err != nil {
+			return errors.Wrap(err, "configure outbox", "chain", chain.Name)
+		}
+	}
```
kevinhalliday, line 56: This will take a while to run on each chain. Let's use an `errgroup` (though we can do this after first merge)

kevinhalliday (commented):
Let's put this into `e2e/solve`. It should probably go into `lib/contracts/solvernet` - we already have some route logic there - but don't want to trouble ourselves with splitting up inbox / outbox stuff, cause we're doing both at same time. So I think `e2e/solve` is fine place for it

kevinhalliday (approved):

PR #3915, opened 2025-05-05
fix(e2e): skip invalid routes

Zodomo:
Omni EVM can only have routes for Core-connected chains. Hyperlane chains can only use Hyperlane routes. Core chains can have both. Routes should be properly configured for all now.

issue: #3907


e2e/solve/routing.go
```diff
@@ -32,41 +32,41 @@ func SetSolverNetRoutes(ctx context.Context, network netconf.Network, backends e
 		return errors.Wrap(err, "get addresses", "network", network.ID)
 	}
 
-	chainIDs := network.EVMChains()
 	eg, childCtx := errgroup.WithContext(ctx)
 
-	for _, chain := range chainIDs {
+	for _, chain := range network.EVMChains() {
 		// Capture loop variables for the goroutine closure to avoid race conditions
+		_chain := chain
```
kevinhalliday, line 39: you really only need this right before `eg.Go` if you want to move it

e2e/solve/routing.go
```diff
@@ -81,10 +81,16 @@ func SetSolverNetRoutes(ctx context.Context, network netconf.Network, backends e
...
-	for _, dest := range allChains {
-		if src.ID == dest.ID {
+	for _, dest := range network.EVMChains() {
+		// Skip Hyperlane routes on Omni EVM.
+		if netconf.IsOmniExecution(network.ID, src.ID) && solvernet.IsHLChain(dest.ID) {
+			continue
+		}
+
+		// Skip non-Hyperlane routes on Hyperlane chains.
+		if solvernet.IsHLChain(src.ID) && !solvernet.IsHLChain(dest.ID) {
```
kevinhalliday, line 93: `IsHlChain` will return false to a chain like eth l1, that is both core and hl no

kevinhalliday (commented):
I think we should move logic to a `SupportsHL(chainID)`  in lib/contracts/solvernet

and have that util handle chains that support both core and hl, which right now IsHLChain does not

e2e/solve/routing.go
Zodomo: Yes that is correct. Ethereum is not an HL chain, so it will skip both of these conditions and assign all chains.

Zodomo: Removing the condition skipping non-hyperlane routes on hyperlane chains. We will set all routes on hyperlane chains and just have them use hyperlane instead.

kevinhalliday (approved):

PR #3924, opened 2025-05-06
chore(e2e): fix and refactor solvernet routing

kevinhalliday:
Refactor solvernet routing to use a single Provider(src, dest) util,
that returns the preferred interpo provider between a src and dest chain.

Use things in e2e/solve/routing and lib/contracts/solvernet/deploy. Fix routing tests.

issue: none

### DIFF
--- a/lib/contracts/solvernet/config.go
+++ b/lib/contracts/solvernet/config.go
@@ -1,19 +1,19 @@
 package solvernet
 
 import (
-	"github.com/omni-network/omni/lib/errors"
 	"github.com/omni-network/omni/lib/evmchain"
 
 	"github.com/ethereum/go-ethereum/common"
 )
 
 const (
-	None      uint8 = iota // 0
-	OmniCore  uint8 = 1    // 1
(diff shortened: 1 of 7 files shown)


lib/contracts/solvernet/config.go
```diff
@@ -1,19 +1,19 @@
...
 	"github.com/ethereum/go-ethereum/common"
 )
 
 const (
-	None      uint8 = iota // 0
-	OmniCore  uint8 = 1    // 1
-	Hyperlane uint8 = 2    // 2
+	ProviderNone uint8 = iota // No provider
+	ProviderCore uint8 = 1    // Omni Core
+	ProviderHL   uint8 = 2    // Hyperlane
```
kevinhalliday, line 12: Generally good to prefix exported constants with their "type" / "namespace"

examples: 
- `evmchain.IDEthereum`
- `cctp.MsgStatusSubmitted`
- `solvertypes.RejectNone`

Zodomo (approved):

Issue #3925, opened 2025-05-06
Upgrade SolverNet contracts on Hyperlane chains


Zodomo:
#### Problem to Solve
`e2e/app/admin/upgrade.go` is what executes contract upgrades on our supported chains. Currently, this only supports Core-connected chains. We will need the capability to upgrade these contracts on Hyperlane chains too.

#### Proposed Solution
Refactor the upgrade commands such that Hyperlane chains are properly added to the network and upgrades on them are triggered.

PR #3926, opened 2025-05-06
feat(e2e): upgrade contracts on hyperlane chains

Zodomo:
Refactored the upgrade commands for SolverNet contracts such that the upgrade routine includes Hyperlane chains in the network.

issue: #3925

### DIFF
--- a/e2e/app/admin/common.go
+++ b/e2e/app/admin/common.go
@@ -11,6 +11,7 @@ import (
 	fbproxy "github.com/omni-network/omni/e2e/fbproxy/app"
 	"github.com/omni-network/omni/e2e/types"
 	"github.com/omni-network/omni/lib/contracts"
+	"github.com/omni-network/omni/lib/contracts/solvernet"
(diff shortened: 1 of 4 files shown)


e2e/app/admin/common.go
```diff
@@ -145,6 +185,35 @@ func (s shared) run(
 	return nil
 }
 
+// runHL runs a function for all configured chains (including Hyperlane chains).
+// NOTE: currently does not include options.
+func (s *shared) runHL(ctx context.Context, def app.Definition, fn func(context.Context, shared, chain) error) error {
```
kevinhalliday, line 190: this should not need to be a pointer. always prefer value receiver over pointer receiver when you don't need pointer receiver

kevinhalliday (approved):

e2e/app/admin/common.go
Zodomo: revised to use a new `addHLEndpoints(shared)` function that returns a new copy of the `shared` struct with the updated endpoints rather than using a pointer.

PR #3927, opened 2025-05-06
chore(sdk): add useQuote to example app

ga-reth:
- add `useQuote` to example react app

issue: none

### DIFF
--- a/sdk/examples/react/src/App.tsx
+++ b/sdk/examples/react/src/App.tsx
@@ -1,5 +1,5 @@
-import { useOrder } from '@omni-network/react'
-import { parseEther, zeroAddress } from 'viem'
+import { useOrder, useQuote } from '@omni-network/react'
+import { formatEther, parseEther, zeroAddress } from 'viem'
 import { baseSepolia, holesky } from 'viem/chains'
 import { useAccount, useConnect, useDisconnect, useSwitchChain } from 'wagmi'
 
@@ -8,6 +8,7 @@ function App() {
     <>
       <Account />
       <Order />
+      <Quote />
     </>
   )
 }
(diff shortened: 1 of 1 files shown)


sdk/examples/react/src/App.tsx
```diff
@@ -52,6 +53,42 @@ function Account() {
...
 
+function Quote() {
+  const account = useAccount()
+  const quote = useQuote({
+    srcChainId: baseSepolia.id,
+    destChainId: holesky.id,
+    deposit: { amount: parseEther('0.1'), isNative: true },
+    expense: { isNative: true },
+    mode: 'expense',
+    enabled: !!account?.address,
```
kevinhalliday, line 64: we don't need account for quote tho

sdk/examples/react/src/App.tsx
```diff
@@ -52,6 +53,42 @@ function Account() {
...
+          <div>isSuccess: {quote.isSuccess}</div>
+          <div>isPending: {quote.isPending}</div>
+          <div>isError: {quote.isError}</div>
+          <div>
+            quote.deposit.amount:{' '}
+            {quote.isSuccess ? formatEther(quote?.deposit.amount) : ''}
+          </div>
+          <div>
+            quote.expense.amount:{' '}
+            {quote.isSuccess ? formatEther(quote?.expense.amount) : ''}
```
kevinhalliday, line 82: if `quote.isSuccess`, to we need to optional chaining on `quote?.deposit.amount`

kevinhalliday (approved):

sdk/examples/react/src/App.tsx
ga-reth: good catch!

Issue #3744, opened 2025-04-17
Monitoring / e2e testing for 3rd party settlement


ttarsi:

ttarsi:
migrated

PR #3929, opened 2025-05-06
feat(monitor): expand monitoring to hyperlane chains

Zodomo:
Expanded the existing monitoring, where appropriate, to Hyperlane chains. New monitoring for messages routed through Hyperlane has not yet been implemented.

issue: #3744


kevinhalliday (changes requested):
Let's start with addressing one sub monitor routine at a time. And look into code there to  see what will happen when we expand network past core

Zodomo:
Going to purge this and revisit once we get RPC overrides into the Monitor first.

PR #3930, opened 2025-05-06
chore(docs): watchDidFill docs entry

ga-reth:
- docs for watchDidFill

issue: none


docs/docs/pages/sdk/core/watchDidFill.md
```diff
@@ -0,0 +1,72 @@
...
+`Hex` | `0x${string}`
+```
+
+### `unwatch`
+
+Terminates watching of contract events.
+
+```ts
+() => void
+```
```
PaulLeCam, line 54: Maybe when returning an object like this, we can use a table with key / type / description as we do for parameters?

docs/docs/pages/sdk/core/watchDidFill.md
```diff
@@ -0,0 +1,72 @@
...
+```ts
+'pending' | 'filled' | 'idle'
+```
+
+### `destTxHash`
+
+The transaction hash on the destination chain that fills the order.
+
+```ts
+`Hex` | `0x${string}`
```
PaulLeCam, line 45: Seems redundant considering that `Hex` already translates to `0x${string}`, no?

PaulLeCam (approved):

docs/docs/pages/sdk/core/watchDidFill.md
ga-reth: yah but just being explicit in case folks aren't familiar viem has a hex type

PR #3933, opened 2025-05-06
feat(solver/rebalance): send by deficits

kevinhalliday:
Send surplus to chains in deficit.

- don't send what's needed on current chain
- prioritize send by chains most in deficit

Most noteable change is change to rebalance loop. Instead of 
doing each loop (sending, swapping, filling) independently, on each chain. 
We're now doing each step sequentially, and each chain sequentially.

Reasoning:
- makes sure we're only sending one send at a time, so no race conditions on chain deficit calcs
- each step still functions indendently, but doing them in order makes logic & logs easier to follow,
  and we're not worried about speed

issue: none

### DIFF
--- a/solver/rebalance/rebalance.go
+++ b/solver/rebalance/rebalance.go
@@ -10,7 +10,6 @@ import (
 	"github.com/omni-network/omni/lib/errors"
 	"github.com/omni-network/omni/lib/ethclient/ethbackend"
 	"github.com/omni-network/omni/lib/evmchain"
-	"github.com/omni-network/omni/lib/expbackoff"
 	"github.com/omni-network/omni/lib/log"
 	"github.com/omni-network/omni/lib/netconf"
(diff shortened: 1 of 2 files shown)


solver/rebalance/rebalance.go
```diff
@@ -38,109 +37,48 @@ func rebalanceForever(
...
-				log.Error(ctx, "Swap surplus failed", err)
+	go func() {
+		for {
+			start := time.Now()
+			rebalanceOnce(ctx, db, network, pricer, backends, solver)
+			elapsed := time.Since(start)
+
+			// Sleep for the remaining time in the interval, if any.
+			if elapsed < interval {
+				time.Sleep(interval - elapsed)
```
kevinhalliday, line 48: rebalance each chain every thirty minutes. if it takes longer than thirty minutes, just do it again right away

Zodomo (approved):

PR #3941, opened 2025-05-07
fix(e2e): skip omni consensus chain

Zodomo:
The Omni consensus chain needs to be skipped within the expanded network when using the HL-specific upgrade functions.

issue: #3925

### DIFF
--- a/e2e/app/admin/common.go
+++ b/e2e/app/admin/common.go
@@ -221,11 +221,7 @@ func (s shared) runHL(ctx context.Context, def app.Definition, fn func(context.C
 	}
 	network = solvernet.AddHLNetwork(ctx, network, solvernet.FilterByEndpoints(_s.endpoints))
 
-	for _, _chain := range network.Chains {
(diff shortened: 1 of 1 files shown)


e2e/app/admin/common.go
```diff
@@ -222,6 +222,10 @@ func (s shared) runHL(ctx context.Context, def app.Definition, fn func(context.C
 	network = solvernet.AddHLNetwork(ctx, network, solvernet.FilterByEndpoints(_s.endpoints))
 
 	for _, _chain := range network.Chains {
+		if netconf.IsOmniConsensus(_s.testnet.Network, _chain.ID) {
+			continue
+		}
+
```
kevinhalliday, line 228: network.EVMChains() is used to this purpose

kevinhalliday (approved):

e2e/app/admin/common.go
Zodomo: Removed the condition and used `network.EVMChains()` instead

PR #3949, opened 2025-05-08
chore(ci): skip rebalance test

kevinhalliday:
Skip rebalance test until flaps fixed. 

issue: none


kevinhalliday:
Flaps fixed, won't need

PR #3954, opened 2025-05-08
chore(sdk): bump versions

ga-reth:
- bump versions since publish didn't strip `catalog:` from package.json
-  remove alpha comments 
- update scripts

issue: none

### DIFF
--- a/sdk/packages/core/package.json
+++ b/sdk/packages/core/package.json
@@ -22,7 +22,7 @@
     "clean": "rm -rf dist tsconfig.tsbuildinfo",
     "coverage": "vitest run --coverage",
     "prepublishOnly": "pnpm build",
-    "publish": "pnpm publish --access public",
+    "publish": "pnpm publish",
     "test": "vitest run"
   },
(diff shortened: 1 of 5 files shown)


sdk/packages/core/package.json
```diff
@@ -19,10 +19,11 @@
   "scripts": {
     "build": "pnpm clean && pnpm build:ts",
     "build:ts": "tsc -p tsconfig.build.json",
-    "prepublishOnly": "pnpm build",
     "clean": "rm -rf dist tsconfig.tsbuildinfo",
-    "test": "vitest run",
-    "coverage": "vitest run --coverage"
+    "coverage": "vitest run --coverage",
+    "prepublishOnly": "pnpm build",
+    "publish": "pnpm publish --access public",
```
PaulLeCam, line 25: I don't think we need the `--access public` here as the packages are already public.
How does it behave when publishing all packages from the workspace root please?

ga-reth: removed

PaulLeCam (approved):

PR #3957, opened 2025-05-08
feat(sdk): add example script using core

PaulLeCam:
Similar to the React example

issue: none


ga-reth:
we can create a wallet client from some browser wallets via the below, avoiding the need to export / setup private keys:

```
import "viem/window"
import { createWalletClient, custom } from "viem"

const walletClient = createWalletClient({
  chain,
  transport: custom(window.ethereum!)
})

const [address] = await walletClient.requestAddresses()
```

might I suggest we use something like this to make it super easy for devs to run the examples? and avoid the friction of managing/exporting pks?

of course you'll need client files, but we could over time create examples using diff client fwks, but for now can just do a tiny vite vanilla/react

ga-reth (approved):

PR #3966, opened 2025-05-09
chore(*): init changeset and workflow

ga-reth:
- init changesets in sdk monorepo
- setup changesets workflow
- trigger in ci-main
- ref OMNI-162

issue: none


sdk/packages/react/package.json
```diff
@@ -42,7 +42,6 @@
     "wagmi": "2.x"
   },
   "devDependencies": {
-    "@omni-network/test-utils": "workspace:^",
```
ga-reth, line 45: need to move to workspace root as changeset complains about this being here

PaulLeCam: wouldn't this cause issues with pnpm? seems strange to me to have a workspace package hoisted to the root like this

ga-reth: it works fine and have seen some other monorepos do the same thing

PaulLeCam (approved):

PR #3971, opened 2025-05-12
test(sdk): add integration test for using ethers

PaulLeCam:
Adds utility function to create a viem client from an ethers provider or signer

issue: none


ga-reth (approved):
huge! nice work man 🚀

PR #3982, opened 2025-05-12
feat(e2e/solve): filter out applied routes

Zodomo:
All valid routes are currently being written when routes are set. This change filters out routes that are already applied, saving gas.

issue: none

### DIFF
--- a/e2e/solve/routing.go
+++ b/e2e/solve/routing.go
@@ -35,11 +35,15 @@ func SetSolverNetRoutes(ctx context.Context, network netconf.Network, backends e
 	eg, childCtx := errgroup.WithContext(ctx)
 
 	for _, chain := range network.EVMChains() {
-		routes := getRoutes(chain, network, addrs.SolverNetInbox, addrs.SolverNetOutbox)
-
 		backend, err := backends.Backend(chain.ID)
 		if err != nil {
 			return errors.Wrap(err, "get backend", "chain", chain.Name)
+		}
+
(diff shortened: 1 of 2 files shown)


e2e/solve/routing.go
```diff
@@ -93,14 +97,60 @@ func getRoutes(src netconf.Chain, network netconf.Network, inbox common.Address,
...
+
+		outboxConfig, err := outbox.GetInboxConfig(callOpts, src.ID)
+		if err != nil {
+			return nil, errors.Wrap(err, "get outbox inbox config", "chain", backend.Name())
+		}
+
+		currentRoutes = append(currentRoutes, Route{
+			ChainID:     dest.ID,
+			Outbox:      inboxConfig,
+			InboxConfig: outboxConfig,
```
kevinhalliday, line 137: this looks flipped?

kevinhalliday (approved):
maybe a comment on that this does not work for route removals
we can add that when / if we need

e2e/solve/routing.go
Zodomo: It's not flipped but the naming is weird, I'll clarify it.

Zodomo:
> maybe a comment on that this does not work for route removals we can add that when / if we need

`getRoutes` will insert zeroed out configs for chains in `solvernet.IsDisabled`, which should cause `filterRoutes` to read configs for those chains and ensure they're disabled.

PR #3988, opened 2025-05-12
fix(e2e): fix inbox portal address assignment

Zodomo:
Portal address needs to be the zero address on Hyperlane-only chains.

issue: none


kevinhalliday (approved):
Can also do just portal

```golang
portal := addrs.Portal
if solvernet.IsHLOnly(c.ChainID) {
    portal := common.Address{} // HL only must have portal == address(0)
}
```

PR #4008, opened 2025-05-13
chore(solver): reconcile rebalance and e2e eoa thresholds

kevinhalliday:
Reconcile rebalance fund thresholds with the thresholds in e2e/app/eoa.

issue: none


solver/app/tokens_internal_test.go
```diff
@@ -127,25 +126,3 @@ func TestTokenResponse(t *testing.T) {
...
-		}
-
-		thresh, ok := eoa.GetSolverNetThreshold(eoa.RoleSolver, netconf.Mainnet, token.ChainID, token.Asset)
-		if !ok {
-			continue
-		}
-
-		tutil.RequireGTE(t, thresh.MinBalance(), bounds.MaxSpend, "solver min balance must be greater than max spend: token=%s, min_bal=%s, max_spend=%s", token.Asset, thresh.MinBalance(), bounds.MaxSpend)
-	}
-}
```
kevinhalliday, line 151: Removing this test. Min balances are now lower. Max spend should be around target balances. But we'll need to start differentiating max spend definitions between chains, which we don't right now.

Plus we have InsufficientInventory rejections, which is good.

Zodomo (approved):

PR #4012, opened 2025-05-13
fix(e2e): fix route filtering

Zodomo:
Route filtering had a typo that caused all routes to be applied each time.

### DIFF
--- a/e2e/solve/routing.go
+++ b/e2e/solve/routing.go
@@ -41,7 +41,7 @@ func SetSolverNetRoutes(ctx context.Context, network netconf.Network, backends e
 		}
 
 		routes := getRoutes(chain, network, addrs.SolverNetInbox, addrs.SolverNetOutbox)
-		routes, err = filterRoutes(ctx, chain, network, backend, addrs.SolverNetInbox, addrs.SolverNetOutbox, routes)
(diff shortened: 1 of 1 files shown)


e2e/solve/routing.go
```diff
@@ -126,7 +126,7 @@ func filterRoutes(ctx context.Context, src netconf.Chain, network netconf.Networ
 			return nil, errors.Wrap(err, "bind outbox", "chain", backend.Name())
 		}
 
-		outboxConfig, err := outbox.GetInboxConfig(callOpts, src.ID)
+		outboxConfig, err := outbox.GetInboxConfig(callOpts, dest.ID)
```
kevinhalliday, line 129: is this outbox or inbox config
or is it the outbox inbox config
lol

kevinhalliday (approved):

e2e/solve/routing.go
Zodomo: Well we assign outbox addresses on the inbox and we assign InboxConfigs on the outbox lol... If you have a better naming convention I'll apply it, but I couldn't come up with one, so within the go `Route` struct, the inboxConfig is applied to the inbox and the outboxConfig is applied to the outbox.

PR #4016, opened 2025-05-13
chore(e2e): move routing log

Zodomo:
Routing log message was in the wrong place.


kevinhalliday (approved):
maybe worth logging filtered routes too, after

Zodomo:
> maybe worth logging filtered routes too, after

They are logged, it just doesn't show up in the preview here.

PR #4017, opened 2025-05-13
fix(contracts/solve): adjust outbox initializer

Zodomo:
When upgrading existing contracts, the Outbox's `initializeV2` function is needed to overwrite `inboxConfig` state because the storage layout had changed. However, on new deployments this is unnecessary, and leaves `initializeV2` exposed (as it has no access restrictions as it is meant to be called by `ProxyAdmin` during an upgrade). To fix this for new deployments, the `initializer` modifier on `initialize` was replaced with `reinitializer(2)`, which will lock out `initializeV2` on fresh deployments.


kevinhalliday (approved):
lovely comment

PR #4032, opened 2025-05-14
chore(e2e): apply routes to inbox only

Zodomo:
We need to temporarily restrict route configuration to the inbox only to unblock the outbox upgrade on Omega.

ref https://linear.app/omni-network/issue/OMNI-244


Zodomo:
Closing as we will handle this in inbox initializer

PR #4036, opened 2025-05-14
feat(contracts/solve): new SolverNetAdmin script

Zodomo:
Migrated all SolverNet functions in `contracts/core/script/admin/Admin.s.sol` to `contracts/solve/script/admin/SolverNetAdmin.s.sol`, and updated `e2e` to use that instead for SolverNet upgrades and pauses.


kevinhalliday (approved):
nice

PR #4046, opened 2025-05-15
feat(contracts/solve): support simpler fill hash

Zodomo:
We had an unnecessary `abi.encode` in our fill hash, this is complicating fill hash derivation on Solana. We will remove it as it is unnecessary, and will do so in stages so no orders fail to settle.


contracts/solve/src/SolverNetInbox.sol
```diff
@@ -692,7 +696,21 @@ contract SolverNetInbox is
...
+     * @param orderId ID of the order.
+     * @param fillOriginData Fill origin data to hash. (prevents having to derive it again)
+     */
+    function _simpleFillHash(bytes32 orderId, SolverNet.FillOriginData memory fillOriginData)
+        internal
+        pure
+        returns (bytes32)
+    {
+        return keccak256(abi.encode(orderId, fillOriginData));
     }
```
kevinhalliday, line 714: I'd rather name new one `_fillHash` and old one `_fillHashDepracated` or something like that

Zodomo: Will apply change

kevinhalliday (approved):

PR #4048, opened 2025-05-15
chore(sdk): remove is native flag

ga-reth:
- remove isNative flag from quote
- it creates an awkward experience as a consumer due to the never type, instead we can just make token optional, and infer native when that's the case

issue: none

### DIFF
--- a/sdk/examples/core/src/run.ts
+++ b/sdk/examples/core/src/run.ts
@@ -81,8 +81,8 @@ console.log(
 const quote = await getQuote({
   srcChainId: baseSepolia.id,
   destChainId: holesky.id,
-  deposit: { amount: depositAmount, isNative: true },
-  expense: { isNative: true },
+  deposit: { amount: depositAmount },
+  expense: {},
   mode: 'expense',
   environment,
 })
--- a/sdk/examples/core/src/run.ts
+++ b/sdk/examples/core/src/run.ts
@@ -82,7 +82,6 @@ const quote = await getQuote({
   srcChainId: baseSepolia.id,
   destChainId: holesky.id,
   deposit: { amount: depositAmount },
-  expense: {},
   mode: 'expense',
   environment,
 })
--- a/sdk/integration-tests/suites/flows.tsx
+++ b/sdk/integration-tests/suites/flows.tsx
@@ -151,11 +151,6 @@ describe.concurrent('ETH transfer orders', () => {
         destChainId: mockL2Id,
         deposit: {
           amount: parseEther('2'),
-          token: zeroAddress,
-        },
-        expense: {
-          amount: parseEther('1'),
-          token: zeroAddress,
         },
       } as const
 
--- a/sdk/integration-tests/suites/flows.tsx
+++ b/sdk/integration-tests/suites/flows.tsx
@@ -151,11 +151,11 @@ describe.concurrent('ETH transfer orders', () => {
         destChainId: mockL2Id,
         deposit: {
           amount: parseEther('2'),
-          isNative: true,
+          token: zeroAddress,
         },
         expense: {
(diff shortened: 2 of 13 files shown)


sdk/examples/core/src/run.ts
```diff
@@ -81,8 +81,8 @@ console.log(
 const quote = await getQuote({
   srcChainId: baseSepolia.id,
   destChainId: holesky.id,
-  deposit: { amount: depositAmount, isNative: true },
-  expense: { isNative: true },
+  deposit: { amount: depositAmount },
+  expense: {},
```
PaulLeCam, line 85: Do we need to provide the `expense` object if it's empty please?

sdk/integration-tests/suites/flows.tsx
```diff
@@ -151,11 +151,11 @@ describe.concurrent('ETH transfer orders', () => {
         destChainId: mockL2Id,
         deposit: {
           amount: parseEther('2'),
-          isNative: true,
+          token: zeroAddress,
```
PaulLeCam, line 154: Can we omit the `token` here when we transfer ETH?

sdk/packages/core/src/api/getQuote.ts
```diff
@@ -2,17 +2,28 @@ import { type Address, type Hex, fromHex, zeroAddress } from 'viem'
...
-}
+export type GetQuoteParameters = Prettify<
+  {
+    srcChainId?: number
+    destChainId: number
+    environment?: Environment | string
+  } & (
+    | {
+        mode: 'deposit'
+        deposit: Prettify<Omit<Quoteable, 'amount'>>
```
PaulLeCam, line 17: We could make the `deposit` optional here rather than having to provide an empty object
```suggestion
        deposit?: Prettify<Omit<Quoteable, 'amount'>>
```

sdk/packages/core/src/api/getQuote.ts
```diff
@@ -2,17 +2,28 @@ import { type Address, type Hex, fromHex, zeroAddress } from 'viem'
...
+        expense: Prettify<Omit<Quoteable, 'amount'> & { amount: bigint }>
+      }
+    | {
+        mode: 'expense'
+        deposit: Prettify<Omit<Quoteable, 'amount'> & { amount: bigint }>
+        expense: Prettify<Omit<Quoteable, 'amount'>>
```
PaulLeCam, line 23: We could make the `expense` optional here rather than having to provide an empty object
```suggestion
        expense?: Prettify<Omit<Quoteable, 'amount'>>
```

sdk/packages/react/src/hooks/useQuote.test.ts
```diff
@@ -1,22 +1,20 @@
...
 import { useQuote } from './useQuote.js'
 
-const token = '0x123'
-
 beforeEach(() => {
   vi.spyOn(core, 'getQuote').mockResolvedValue(quote)
 })
 
-const params = {
+const params: Parameters<typeof useQuote>[0] = {
```
PaulLeCam, line 12: May the `useQuote` module could export `UseQuoteParams` to avoid having to use these type utils?

PaulLeCam (approved):
LGTM to merge as-is, I think there are some DX improvements that can be made but these could go in a follow-up PR if needed.

sdk/integration-tests/suites/flows.tsx
ga-reth: we can yeah - updated

sdk/packages/react/src/hooks/useQuote.test.ts
ga-reth: fixed!

PR #4049, opened 2025-05-15
feat(contracts/solve): addressed solvernet audit findings

Zodomo:
- OMNI-01: Refund To Non-Payable `SolverNetOutbox` Causes Revert In Dispatch Flow
  - Addressed in prior [commit](https://github.com/omni-network/omni/commit/88d501f0b419b9f3ddeffb5ce44c1314c1279a2e) `88d501f` by sending messaging refund to solver calling `fill`.

- OMNI-02: Retry Unlock Mechanism Missing For Filled Intents
  - Retry mechanism added with `retryMarkFilled` on `SolverNetOutbox`

- OMNI-03: Empty Fallback Function Introduces Unintended Call Acceptance
  - Replaced `SolverNetExecutor` `fallback()` with custom fallback from Solady's `Receiver` contract, which only supports ERC721 and ERC1155 fallbacks.

- OMNI-04: Omni Pausing Prevents Order Closure Even When Using Hyperlane
  - Addressed in prior [commit](https://github.com/omni-network/omni/commit/2c385b677b79633927ffaa93de521b8b3864bd65#diff-88bad93134d1c1f6eeaf3e65a67b002ab5443d7b52f92fdd5f21ae8816da3487) `2c385b6` by removing `OmniPortal` integration checking pause status entirely, as it is no longer a comprehensive control existing across all chains.

- OMNI-05: Missing Fallback Prevents ERC721 And Other Token Interactions
  - Addressed during OMNI-04 correction by implementing Solady's `Receiver` contract, which handles ERC721 and ERC1155 fallbacks.

- OMNI-06: Incorrect `maxSpent.receiver` Assignment In ERC-7683 Implementation
  - Removed `SolverNetOutbox` address assignment and replaced it with `bytes32(0)`. This value is not used anywhere in SolverNet, and we simply populated this value as a placeholder. Because it isn't utilized, in favor of gas savings, we now leave it empty. ERC-7683 also directs solvers to the destination outbox contract within the `FillInstruction` struct, not this field.

- OMNI-07: Denial-of-Service Condition Via Intent Frontrunning On Destination Chain
  - No changes applied for this finding. We can mitigate this entirely at the solver if it ever becomes a problem.

- OMNI-08: Unpausable `fill()` Function Risks Cross-Chain Desynchronization & Solver Penalization
  - No changes applied for this finding. Our intent in leaving `fill` unpaused was that if we needed to halt the system for specific upgrades, it would be best to halt opening and closing, so all in-flight orders could be settled prior to applying them. This consideration is held because Omni currently operates the only solver, but once opened up, this is a move we would likely take. Until then, it is considered unnecessary.

- OMNI-09: Abuse Of `reject()` Function Can Lead To Gas Grief On Solvers
  - No changes applied for this finding. Omni operates the only solver, and this function simply exists to improve user QoL by giving them instant refunds if we aren't going to fill their order. Once solving is opened up to external parties, this function will either be removed entirely, or reworked so that this form of griefing isn't possible.

- OMNI-10: Tokens That Transfer All Balance When Amount is uint256.max Are Drainnable
  - Addressed by rejecting all orders where we receive less tokens than specified. It is important to note that specifying `type(uint256).max` is not possible, as the max possible deposit amount is `type(uint96).max`. However, in the event someone still deposited a token where this would trigger with our max deposit value, this change prevents such deposits.

- OMNI-11: Loss Of Yield With Yield-Bearing Tokens During Intent Lifecycle
  - No changes applied for this finding. We will educate users and devs about this, and discourage deposits of such tokens via our API and SDK. The only yield token we've supported up until now is `stETH`, but support for it has already been deprecated, and may only be supported for fills in the future.

- OMNI-12: Fee-On-Transfer Tokens Distort Users Balances And Revert Claims
  - We solved this in OMNI-10 by preventing any deposits where less is received than what was specified.

- OMNI-13: Gas Consumption Is Not Standard On All Chains
  - No changes applied for this finding. We are aware that chains have different gas accounting models, and will adjust this as needed if we deploy to a chain where changes are necessary. Currently, this gas estimation method works everywhere we are deployed.

- OMNI-14: Lack Of Events For State-Changing Functions Reduces Traceability
  - We are not going to add additional events to state-changing functions, as we feel the `Open`, `Rejected`, `Closed`, `Filled`, and `Claimed` events already provide enough information to infer the state updates. However, we have added events for pause functions.

- OMNI-15: Lack Of Gas Control In `execute()` May Allow Gas Griefing
  - No changes applied for this finding.We are intentionally allowing solvers to handle gas sensitivity. Our solver simulates transactions prior to executing them and should reject orders that cost too much gas. We want solvers to determine whether an order is worth it, and whether they should hold off and try to execute later for lower gas costs. We do not want the negative DX from forcing integrators to specify gas requirements either. We will leave this up to our solver, and later, the solver marketplace to manage.

- OMNI-16: Approval Reset Fallback May Leave Arbitrary Allowance
  - We addressed this by falling back to attempting to set the allowance to 1 if 0 fails, and if 1 fails, we leave it in place.

- OMNI-17: Miscellaneous Code Improvement Recommendations
  - Overly restrictive execution pattern limits flexibility and adds complexity.
    - No changes applied for this finding. Users who need the `SolverNetExecutor` to hold tokens in between actions can already do so by chaining calls together. The `executeAndTransfer` and `executeAndTransfer721` functions are primarily helpers that make opening orders cheaper (as two call array slots aren't necessary), and primarily because the former allows for dynamic token receipts to be forwarded to the user. In circumstances where the deposit:received ratio is not determinstic, this allows everything to be forwarded regardless of the amount. Multi-step actions are still supported, and users have fine-tuned control via the `execute` method.
  - Single Provider Enum Limits Multi-Bridge Support; Use Bitmask or Multi-Flag Representation
    - No changes applied for this finding. We do not currently intend to send messages through multiple messaging providers, and fully intend to only use one provider per route for now. If we find the need to support multiple or add fallbacks, we will switch to using bitmasks instead.


contracts/solve/src/SolverNetExecutor.sol
```diff
@@ -126,14 +132,4 @@ contract SolverNetExecutor is ISolverNetExecutor {
     function transferNative(address to, uint256 amount) external onlyOutbox {
         to.safeTransferETH(amount);
     }
-
-    /**
-     * @dev Allows target contracts to arbitrarily return native tokens to the executor.
-     */
-    receive() external payable { }
```
kevinhalliday, line 133: Can we leave `receive` for refunds with no calldata? maybe with a check that msg.sender != address(this), to protect against misformed calls

contracts/solve/src/SolverNetInbox.sol
```diff
@@ -558,7 +561,9 @@ contract SolverNetInbox is
         if (deposit.token == address(0)) {
             if (msg.value != deposit.amount) revert InvalidNativeDeposit();
         } else {
+            uint256 balance = deposit.token.balanceOf(address(this));
             deposit.token.safeTransferFrom(msg.sender, address(this), deposit.amount);
+            if (deposit.token.balanceOf(address(this)) < balance + deposit.amount) revert InvalidERC20Deposit();
```
kevinhalliday, line 566: Little comment explaining?

contracts/solve/src/SolverNetInbox.sol
```diff
@@ -450,21 +452,22 @@ contract SolverNetInbox is
...
+                recipient: bytes32(0),
                 chainId: header.destChainId
             });
         }
         if (totalNativeValue > 0) {
             maxSpent[expenses.length] = IERC7683.Output({
                 token: bytes32(0),
                 amount: totalNativeValue,
-                recipient: _outboxes[header.destChainId].toBytes32(),
+                recipient: bytes32(0),
```
kevinhalliday, line 470: Little comment explaining?

contracts/solve/src/SolverNetExecutor.sol
kevinhalliday: Or does solady `Receiver` add that?

kevinhalliday (approved):

contracts/solve/src/SolverNetExecutor.sol
Zodomo: Yes Solady `Receiver` has a normal `receive` function.

contracts/solve/src/SolverNetInbox.sol
Zodomo: There's a comment slightly above that explains this

L455: `// maxSpent recipient field is unused in SolverNet, was previously set to outbox as a placeholder`

kevinhalliday (changes requested):

kevinhalliday (approved):

PR #4052, opened 2025-05-15
fix(e2e): fix solvernet admin script cfg

kevinhalliday:
Fix solvernet admin script config.

Requires different root dir and script name than.

issue: none


kevinhalliday:
closing in favor of  #4051

PR #4088, opened 2025-05-20
feat(contracts/core): distribution changes

Zodomo:
Distribution doesn't need to be ownable or initializable, so I simplified it.

ref https://linear.app/omni-network/issue/OMNI-306/deploy-distribution-contract-to-omega


contracts/core/src/octane/Distribution.sol
```diff
@@ -54,6 +45,13 @@ contract Distribution is OwnableUpgradeable {
      */
     function _burnFee() internal {
         require(msg.value >= Fee, "Distribution: insufficient fee");
-        payable(BurnAddr).transfer(msg.value);
+
+        BurnAddr.safeTransferETH(Fee);
+
+        // Send any overpayment back to the caller
```
fabtreb, line 51: sg

fabtreb (approved):

PR #4107, opened 2025-05-21
feat(solver): support debug trace in check

kevinhalliday:
Support debug_traceCall in /check requests

issue: none


solver/app/handler.go
```diff
@@ -75,19 +75,37 @@ func newCheckHandler(checkFunc checkFunc) Handler {
...
 
+			// Returns trace result if debug == true, else nil.
+			maybeTrace := func() map[string]any {
+				if !req.Debug {
+					return nil
+				}
+
+				trace, err := traceFunc(ctx, *req)
+				if err != nil {
+					return map[string]any{"error": err.Error()}
```
corverroos, line 86: can do `errors.Format(err)` instead to get structured args as well

PR #4125, opened 2025-05-23
feat(sdk): support debug flag in validation endpoint

PaulLeCam:
Will add integration tests in another PR

issue: none

### DIFF
--- a/sdk/packages/react/src/hooks/useOrder.ts
+++ b/sdk/packages/react/src/hooks/useOrder.ts
@@ -45,6 +45,7 @@ import {
 
 type UseOrderParams<abis extends OptionalAbis> = Order<abis> & {
   validateEnabled: boolean
+  validateDebug?: boolean
   omniContractsQueryOpts?: QueryOpts<OmniContracts>
   getOrderQueryOpts?: QueryOpts<GetOrderReturn>
   didFillQueryOpts?: QueryOpts<boolean>
@@ -103,6 +104,7 @@ export function useOrder<abis extends OptionalAbis>(
 ): UseOrderReturnType {
   const {
     validateEnabled,
+    validateDebug,
     omniContractsQueryOpts,
     getOrderQueryOpts,
     didFillQueryOpts,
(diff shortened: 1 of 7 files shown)


sdk/packages/react/src/hooks/useOrder.ts
```diff
@@ -45,6 +45,7 @@ import {
 
 type UseOrderParams<abis extends OptionalAbis> = Order<abis> & {
   validateEnabled: boolean
+  validateDebug?: boolean
```
ga-reth, line 48: how about `debugTrace` / `debugTraceValidate` or smth similar ? solver returns `debug_traceCall` result so gives more indication of what behaviour this param will enable

sdk/packages/react/src/hooks/useValidateOrder.ts
```diff
@@ -14,6 +14,7 @@ import type { QueryOpts } from './types.js'
 type UseValidateOrderParams<abis extends OptionalAbis> = {
   order: Order<abis>
   enabled: boolean
+  debug?: boolean
```
ga-reth, line 17: same as above point I guess, how about something like debugTrace / enableDebugTrace ?

ga-reth:
need a changeset, and we should add an integration test for this

ga-reth:
need a changeset and to update the docs, and I guess we should add an integration test for this also

ga-reth (commented):
need a changeset and to update the docs, and I guess we should add an integration test for this also

sdk/packages/react/src/hooks/useOrder.ts
ga-reth, line 48: replacing previous comments - I think kevin's suggestion debugValidation might be better here - wdyt?

PaulLeCam:
Yeah I prefer to have a dedicated PR for changeset + docs as merging the docs changes will also publish them, better having them when ready to publish

sdk/packages/react/src/hooks/useOrder.ts
PaulLeCam: I used `validateDebug` here as we use `validateEnabled` already so I thought that would make the props more consistent, but yes no problem using `debugValidation` if we prefer

ga-reth (approved):

PR #4129, opened 2025-05-23
feat(contracts/solve): add trusted route to outbox

Zodomo:
Adding a trusted no-op route to the `SolverNetOutbox` to service newer chains like Solana.


contracts/solve/src/SolverNetOutbox.sol
```diff
@@ -358,6 +361,9 @@ contract SolverNetOutbox is
                 uint32(fillData.srcChainId), inboxConfig.inbox, message, _fillGasLimit(fillData), msg.sender
             );
             _dispatch(uint32(fillData.srcChainId), inboxConfig.inbox, fee, message, _fillGasLimit(fillData), msg.sender);
+        } else if (inboxConfig.provider == Provider.Trusted) {
+            // Trusted routes simply emit an event as they do not rely on messaging, so we return 0 here.
+            return 0;
```
kevinhalliday, line 366: Maybe a little comment on why

"This is a temporary measure for fills from Solana"

or something like that

kevinhalliday (approved):

contracts/solve/src/SolverNetOutbox.sol
Zodomo: Adding that comment

PR #4135, opened 2025-05-28
chore(svmutil): improve event streaming

corverroos:
Improve streaming and parsing of solana events.

issue: none


lib/svmutil/helpers.go
```diff
@@ -101,3 +103,55 @@ func TokenBalanceAt(ctx context.Context, cl *rpc.Client, mint, wallet solana.Pub
 
 	return bal, nil
 }
+
+var stackRegex = regexp.MustCompile(`Program (\w+) (invoke|success|failed)(.*)?`)
+
+// FilterDataLogs filters the logs for a specific program, returning only the data logs
+// and true if logs were not truncated.
+func FilterDataLogs(logs []string, program solana.PublicKey) ([]string, bool, error) {
```
corverroos, line 111: only decode logs from a specific program, otherwise, it simply tries to decode all events, which could result in malicious program injecting logs...

kevinhalliday (approved):

PR #4137, opened 2025-05-28
chore(contracts/solve): restart gasless order impl

Zodomo:
Reimplemented gasless order logic in updated `SolverNetInbox`. This includes all changes except for the type hashing for the permit2 witness transfer, which will be done separately.

ref https://linear.app/omni-network/issue/OMNI-330


contracts/solve/src/SolverNetInboxV2.sol
```diff
@@ -817,7 +866,21 @@ contract SolverNetInboxV2 is OwnableRoles, ReentrancyGuard, Initializable, Deplo
             expenses: expenses
         });
 
-        return keccak256(abi.encode(orderId, abi.encode(fillOriginData)));
+        return (keccak256(abi.encode(orderId, abi.encode(fillOriginData))), fillOriginData);
```
kevinhalliday, line 869: why adding extra return?

PR #4165, opened 2025-06-02
chore(contracts/solve): condense implementations

Zodomo:
Condensing the separate `SolverNetInboxV2` implementation into the core `SolverNetInbox` contract.

ref none


contracts/solve/src/SolverNetInbox.sol
```diff
@@ -294,14 +330,32 @@ contract SolverNetInbox is
...
+     * @dev Token deposits are transferred from order.user to this inbox, native deposits are not supported.
+     * @param order GaslessCrossChainOrder to open.
+     * @param signature Signature from order.user.
+     */
+    function openFor(GaslessCrossChainOrder calldata order, bytes calldata signature, bytes calldata)
+        external
+        whenNotPaused(OPEN)
+        nonReentrant
+    {
+        (SolverNet.OmniOrderData memory orderData, SolverNet.Order memory _order) = _validateFor(order);
```
kevinhalliday, line 353: We don't need to call this `OmniOrderData` (if it's the same as `OrderData`). Can just have the one. And just that in the "Omni" prefiex in typehash name.

kevinhalliday (approved):

contracts/solve/src/SolverNetInbox.sol
Zodomo: I'll remove the duplicate type but leave it in the typehash.

PR #4181, opened 2025-06-03
fix(solver/app): fix same chain checks

kevinhalliday:
Use debug_traceCall to check same chain orders,
because quicknode does not return error code in eth_call output.

issue: none


solver/types/calltrace.go
```diff
@@ -0,0 +1,50 @@
...
+	Gas     string      `json:"gas"`
+	GasUsed string      `json:"gasUsed"`
+	Input   string      `json:"input"`
+	Output  string      `json:"output"`
+	Error   string      `json:"error"`
+	Calls   []CallTrace `json:"calls,omitempty"`
+}
+
+// Map converts CallTrace to a map[string]any.
+func (t CallTrace) Map() map[string]any {
```
kevinhalliday, line 20: Using this to not change return type of CheckRequest and check handlers. Was a bit easier, because we insert an {"error": "some error"} if something goes wrong with trace

Zodomo (approved):

PR #4187, opened 2025-06-04
fix(solver/rebalance): fix chain mutexes

kevinhalliday:
Fix chain mutexes, avoid deadlocks.

issue: none

### DIFF
--- a/solver/rebalance/chainmu.go
+++ b/solver/rebalance/chainmu.go
@@ -14,27 +14,27 @@ var (
 
 // lock acquires the mutexes for the given chain IDs.
 func lock(chainIDs ...uint64) {
-	mu.Lock()
-	defer mu.Unlock()
-
 	for _, chainID := range chainIDs {
-		mu, ok := chainMutexes[chainID]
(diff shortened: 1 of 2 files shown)


Zodomo (approved):

solver/rebalance/chainmu.go
```diff
@@ -14,27 +14,27 @@ var (
...
+
+func getMutex(chainID uint64) *sync.Mutex {
 	mu.Lock()
 	defer mu.Unlock()
 
-	for _, chainID := range chainIDs {
-		if mu, ok := chainMutexes[chainID]; ok {
-			mu.Unlock()
-		}
+	m, ok := chainMutexes[chainID]
```
corverroos, line 33: can replace this with slighly more readable:
```
var mutexes sync.Map

func getMutex(chainID uint64) *sync.Mutex {
  return m.LoadOrStore(chainID, new(sync.Mutex)).(*sync.Mutex)
}
```

corverroos (approved):

PR #4205, opened 2025-06-04
chore(sdk): bump versions

ga-reth:
- bump versions for rejection changes

issue: none

### DIFF
--- a/sdk/packages/core/CHANGELOG.md
+++ b/sdk/packages/core/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @omni-network/core
 
+## 0.3.1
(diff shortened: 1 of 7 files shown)


sdk/packages/core/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @omni-network/core
 
+## 0.4.0
```
PaulLeCam, line 3: Does it need to be a v0.4.0? If there's no breaking change, that could be added as v0.3.x, no?

PaulLeCam (approved):

PR #4231, opened 2025-06-10
feat(sdk): add support for SVM assets in API responses

PaulLeCam:
Starts refining types in the SDK to differentiate between EVM and SVM assets and orders

issue: https://linear.app/omni-network/issue/OMNI-384/sdk-support-for-svm-types-in-api-interactions


sdk/packages/core/src/schema/types.ts
```diff
@@ -8,9 +10,20 @@ export const hex = () =>
...
+  ) as z.ZodMiniType<EVMAddress>
+
+export const svmAddress = () =>
+  z.string().check(
+    z.regex(/^[1-9A-HJ-NP-Za-km-z]{32,44}$/, {
+      message:
+        'Invalid SVM address: must be base58 encoded and 32 to 44 characters long',
     }),
-  ) as z.ZodMiniType<Address>
+  ) as unknown as z.ZodMiniType<SVMAddress>
```
ga-reth, line 26: why do we need this type cast to unknown for SVMAddresses specifically?

sdk/packages/core/src/types/addresses.ts
```diff
@@ -0,0 +1,6 @@
+import type { Address as SVMAddress } from '@solana/addresses'
+import type { Address as EVMAddress } from 'viem'
+
+export type { EVMAddress, SVMAddress }
+
+export type AnyAddress = EVMAddress | SVMAddress
```
ga-reth, line 6: what do you think about dropping the `Any`?

```
import type { Address as SVMAddress } from '@solana/addresses'
import type { Address as EVMAddress } from 'viem'

export type { EVMAddress, SVMAddress }

export type Address = EVMAddress | SVMAddress
```

same for order also? but keep the EVM/SVM subtypes as is?

sdk/packages/core/src/schema/types.ts
PaulLeCam: `SVMAddress` is an opaque type so we should use the `address()` function from `@solana/addresses` to cast it, but that would add a runtime dependency on this package, which I wanted to avoid.
Here the address validation is made by Zod with the regexp so I think it's safe to cast it to the opaque type.

sdk/packages/core/src/types/addresses.ts
PaulLeCam: I wanted to use `AnyAddress` to make the change explicit and avoid confusion with the `Address` types exported by `@solana/addresses` and `viem`. Maybe we can use another prefix if you have suggestions?

sdk/packages/core/src/schema/types.ts
ga-reth: okay nice!

ga-reth (approved):

PR #4301, opened 2025-07-28
feat: initial nomina token

Zodomo:
Introduce new token contract. Minting function is intentionally left undefined.

ref https://linear.app/omni-network/issue/OMNI-527

issue: none


contracts/nomina/.github/workflows/test.yml
```diff
@@ -0,0 +1,40 @@
+name: CI
```
kevinhalliday, line 1: Does this run here? Doesn't need to be in root `.github`?

that would be nice

contracts/nomina/src/Nomina.sol
```diff
@@ -0,0 +1,55 @@
...
+        if (amount == 0) revert ZeroAmount();
+        if (_omni == address(0)) revert ConversionDisabled();
+
+        _omni.safeTransferFrom(msg.sender, _DEAD_ADDRESS, amount);
+        _mint(to, amount * _CONVERSION_RATE);
+    }
+
+    function _constantNameHash() internal pure override returns (bytes32) {
+        return _NAME_HASH;
+    }
```
kevinhalliday, line 54: what's this for?

kevinhalliday (commented):
Q on GH actions. And looks like you have a git submodule forge-std, as well as the package.json install.

Prob best to do package .json install?

contracts/nomina/.github/workflows/test.yml
Zodomo: That's just the default workflow created by `forge init`. I can remove it.

contracts/nomina/src/Nomina.sol
Zodomo: Improving the gas efficiency of permit2 operations in Solady ERC20

contracts/nomina/.github/workflows/test.yml
Zodomo: Removed this default workflow

PR #4303, opened 2025-07-29
chore(contracts): temp disable staking and distribution

Zodomo:
This PR temporarily disables staking, delegation, unstaking, and undelegation in the `Staking` and `Distribution` contracts. This PR **should not be merged** until we are ready to upgrade the contracts and proceed with the migration.

issue: none


Zodomo:
closing as it was determined that this is unnecessary

PR #4327, opened 2025-08-07
feat(e2e): integrate CreateX

Zodomo:
Integrated CreateX into `e2e` so that it would be present on devnet chains, and added helpers to compute `CREATE2` addresses using its guarded salt mechanics. CreateX is necessary as it is what we will use to deploy Nomina contracts, as we want to mine salts for desirable addresses.

issue: none


Zodomo:
Merged in PR #4331

PR #4330, opened 2025-08-07
feat(contracts): token improvements

Zodomo:
Addressed informational audit findings such as capitalizing the immutable `OMNI` variable and adding a two step mint authority transfer process. Also removed the `minter` assignment in the constructor, required all addresses in constructor to be initialized, and removed the disabled conversion state.

issue: none


Zodomo:
Merged in PR #4331

PR #4352, opened 2025-08-20
feat(solver): temp reject omni orders

Zodomo:
Implemented simple rejections for all orders to and from Omni EVM. These changes should be removed after the EVM upgrade is successful.

issue: none

### DIFF
--- a/solver/app/check_internal_test.go
+++ b/solver/app/check_internal_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/omni-network/omni/e2e/app/eoa"
 	"github.com/omni-network/omni/lib/contracts"
 	"github.com/omni-network/omni/lib/errors"
+	"github.com/omni-network/omni/lib/evmchain"
 	"github.com/omni-network/omni/lib/netconf"
 	"github.com/omni-network/omni/lib/tutil"
 	"github.com/omni-network/omni/solver/client"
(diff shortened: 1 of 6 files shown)


kevinhalliday (approved):

solver/app/check_internal_test.go
```diff
@@ -38,6 +39,16 @@ func TestCheck(t *testing.T) {
...
+		// TODO(zodomo): Remove this once network upgrade is complete
+		if tt.req.SourceChainID == evmchain.IDOmniMainnet || tt.req.SourceChainID == evmchain.IDOmniOmega || tt.req.SourceChainID == evmchain.IDOmniStaging {
+			continue
+		}
+
+		// TODO(zodomo): Remove this once network upgrade is complete
+		if tt.req.DestinationChainID == evmchain.IDOmniMainnet || tt.req.DestinationChainID == evmchain.IDOmniOmega || tt.req.DestinationChainID == evmchain.IDOmniStaging {
+			continue
+		}
+
```
kevinhalliday, line 51: if you wanna gut check you could over ride expected rejection here, rather than continuing

Zodomo: Had to overwrite more than the reason lol, had to overwrite accepted status, set rejected status, and set the rejection reason, then regenerate the golden file for these tests, but then it worked. It's now strict instead of just a bypass.

PR #4468, opened 2025-12-24
refactor: replace sort.Slice with slices.Sort for natural ordering

findfluctuate:
There is a [new function](https://pkg.go.dev/slices#Sort) added in the go1.21 standard library, which can make the code more concise and easy to read.

issue: none


criciss:
I made the same observation, but I saw that it has already been submitted.
Good fixes. @findfluctuate 

Hi, Could you please review this PR at your convenience?  CC @kevinhalliday

